PR #5, opened 2021-07-26
fix for windows

Andrey-Russe:
fix for windows build.
added manuals for environment setup: Windows, MacOS, Linux
Linux - TODO

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -35,5 +35,8 @@ cc_library(
     srcs = [
         "stout/scheduler.cc",
     ],
+    deps = [
+        "@com_github_google_glog//:glog",
+    ]
     visibility = ["//visibility:public"],
 )
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,4 +1,5 @@
 load("@rules_cc//cc:defs.bzl", "cc_library")
+load("//:variables.bzl", "COPTS")
 
 cc_library(
     name = "eventuals",
@@ -38,5 +39,6 @@ cc_library(
     deps = [
         "@com_github_google_glog//:glog",
     ],
+	copts = COPTS,
     visibility = ["//visibility:public"],
 )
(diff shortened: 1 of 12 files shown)


stout/semaphore.h
```diff
@@ -2,13 +2,14 @@
 
 #ifdef __MACH__
 #include <mach/mach.h>
-#elif __WINDOWS__
+#elif _WIN32
 #else
```
benh, line 6: `#include <windows.h> // For Windows semaphore.`

And update line 54 to be `_WIN32` not `_WINDOWS_`.

test/scheduler.cc
```diff
@@ -1,3 +1,5 @@
+#ifndef _WIN32 //temp solution for WINDOWS.
```
benh, line 1: Hopefully don't need this now that we include correct file above.

stout/callback.h
```diff
@@ -1,5 +1,9 @@
 #pragma once
 
+# if defined(_WIN32)
```
benh, line 3: Can we use `#ifdef` to be consistent with all the other ones?

BUILD.bazel
```diff
@@ -35,8 +35,5 @@ cc_library(
     srcs = [
         "stout/scheduler.cc",
     ],
-    deps = [
```
benh, line 38: This will break Linux and Mac OS builds now?

benh (changes requested):

benh:
Hey @Andrey-Russe, are you able to update this pull request with the changes requested?

stout/semaphore.h
Andrey-Russe: It is done.

BUILD.bazel
Andrey-Russe: It is cancelled.

benh:
This was merged after some tweaks in 5eb7153b88893eca2b9f53f6e3987dd65a2792c1.

PR #6, opened 2021-07-26
Adding pre-commit & .clang-format

ArthurBandaryk:
Added git pre-commit hook and .clang-format file.


.clang-format
```diff
@@ -0,0 +1,200 @@
+---
```
benh, line 1: Let's leave a comment with how we created this file so people in the future can try and recreate it if necessary.

benh (changes requested):
Let's move `pre-commit` into a `tools` directory and also add a `bootstrap` helper script which will "install" the `pre-commit` hook for people via `ln -s` into the `.git\hooks` directory (we want to use `ln -s` rather than just making a copy so that we can actually update it in the repository and everyone will benefit from the update).

.clang-format
ArthurBandaryk: Added comments about how .clang-format was created!

PR #10, opened 2021-08-02
.clang-format + tools dir(bootstrap, pre-commit, README.md)

ArthurBandaryk:
Added check_format_file() function in order to just check if current file was formatted correctly. If no, I error out about all places in the file which should be formatted correctly and then exit with error "1".


ArthurBandaryk:
check ,please,the commit "49427bb" (just check if specific file was formatted correctly [check_format_file() - function in pre-commit file] )

PR #11, opened 2021-08-02
Linux: fix build error, add clang config. Remove typos in README.md file. Add .gitignore for Bazel dirs

FolMing:


.bazelrc
```diff
@@ -6,4 +6,8 @@ build:windows --compiler="clang-cl" --cxxopt="/std:c++17"
 
 build:macos --cxxopt=-std=c++17
 
-build:linux --cxxopt=-std=c++17
\ No newline at end of file
+build:linux --config=clang_config --cxxopt=-std=c++17
```
benh, line 9: Let's make this specific and explicit to the Linux install for now:

`build:linux --config=linux_usr_lib_llvm_10_config --cxxopt=-std=c++17`

.bazelrc
```diff
@@ -6,4 +6,8 @@ build:windows --compiler="clang-cl" --cxxopt="/std:c++17"
...
+
+build:clang_config --crosstool_top=//toolchain:clang_suite
```
benh, line 11: And let's name the toolchain appropriately too:

`build:linux_usr_lib_llvm_10_config --crosstool_top=//toolchain:linux_usr_lib_llvm_10_toolchain`

.bazelrc
```diff
@@ -6,4 +6,8 @@ build:windows --compiler="clang-cl" --cxxopt="/std:c++17"
...
+build:clang_config --cpu=k8
```
benh, line 12: If I'm reading [this](https://docs.bazel.build/versions/main/cc-toolchain-config-reference.html) correctly we shouldn't need to specify the CPU because Bazel will pick the automatically select the cpu based on where we're currently running. It says "If no flags are specified, Bazel inspects the host system and selects a --cpu value based on its findings." So I agree we need `k8` down in the `cc_toolchain_suite` but doesn't seem like we need to specify it here and we might want to specify `piii` down in `cc_toolchain_suite` too (or leave a comment that since it hasn't been tested we didn't add it but may add it in the future.

Also, for posterity, [here's](https://cs.opensource.google/bazel/bazel/+/1b73bc37e184e71651eb631223dcce321ba16211:src/main/java/com/google/devtools/build/lib/analysis/config/AutoCpuConverter.java) the code that detects and selects the cpu.

toolchain/cc_toolchain_config.bzl
```diff
@@ -0,0 +1,59 @@
...
+            path = "/bin/false",
+        ),
+        tool_path(
+            name = "strip",
+            path = "/bin/false",
+        ),
+    ]
+    return cc_common.create_cc_toolchain_config_info(
+        ctx = ctx,
+	cxx_builtin_include_directories = [
```
benh, line 40: Indent + 2 please.

toolchain/cc_toolchain_config.bzl
```diff
@@ -0,0 +1,59 @@
...
+            "/usr/lib/llvm-10/lib/clang/10.0.0/include", # TODO: avoid hardcoded paths
```
benh, line 41: Happy to have you remove this now that we have a "toolchain" that is named specifically and explicitly for this hard coded path.

toolchain/cc_toolchain_config.bzl
```diff
@@ -0,0 +1,59 @@
...
+        ],
+        toolchain_identifier = "k8-toolchain",
+        host_system_name = "local",
+        target_system_name = "local",
+        target_cpu = "k8",
+        target_libc = "unknown",
+        compiler = "clang",
+        abi_version = "unknown",
+        abi_libc_version = "unknown",
+	tool_paths = tool_paths
```
benh, line 52: Indent + 2 please.

benh (changes requested):
I like the general approach although it is a bit weird to create a "cross compiler" configuration when we're not cross compiling, but I think that's just a deficiency in how Bazel currently works. However, I'd like this configuration to be very specific to what we're doing here, which is building on linux with an LLVM installation at /usr/lib/llvm-10. So how about naming this something like `linux_usr_lib_llvm_10_suite`. Then we can update the documentation for building on Linux to say where we expect LLVM to be installed otherwise it will fail.

toolchain/cc_toolchain_config.bzl
FolMing: Will never use WinSCP integrated editor as code editor again :)

FolMing:
The only thing I'm worried about is whether there is an option in bazel so that we can restrict user from building the library using clang config under a different OS.

benh (commented):
LGTM!

PR #13, opened 2021-08-04
Fix tests build on Windows and Linux

FolMing:
We have to link to libdl for tests to build correctly. It doesn't affect Windows and MacOS builds.

### DIFF
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -19,4 +19,9 @@ cc_test(
         "//:eventuals",
         "@gtest//:gtest_main",
     ],
+    linkopts = select({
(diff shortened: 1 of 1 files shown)


Andrey-Russe:
For WIndows we need the following:

"@bazel_tools//src/conditions:windows": ["/DYNAMICBASE WS2_32.Lib Advapi32.lib Iphlpapi.lib Userenv.lib User32.lib"],  # Windows

FolMing:
Windows build will only work after this PR gets merged: https://github.com/3rdparty/bazel-rules-libuv/pull/2

PR #14, opened 2021-08-04
Windows needs extra Windows system  libraries for starting the program "test/eventuals"

Andrey-Russe:


Andrey-Russe:
This fix will be done via N. by another way

PR #22, opened 2021-08-23
Introduction to eventuals::filesystem

FolMing:
Provides asynchronous operations with the filesystem. Uses libuv as the base.
Those operations are not interruptible due to the way of their implementation in libuv.


benh:
Committed in 5102f9bc8ad1aff1a69899b9507eb80f49cee235.

PR #27, opened 2021-09-03
Add Iterate wrapper

onelxj:


test/iterate.cc
```diff
@@ -0,0 +1,728 @@
...
+#include <set>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+#include "gtest/gtest.h"
+#include "stout/terminal.h"
+
+// std transform + arrays ????
```
benh, line 16: Is this supposed to be a TODO?

stout/iterate.h
```diff
@@ -0,0 +1,136 @@
...
+      .next([](auto& containerData, auto& k) {
+        if (containerData.it_ != containerData.end_) {
+          k.Emit(*(containerData.it_++));
+        } else {
+          k.Ended();
+        }
+      })
+      .done([](auto&, auto& k) {
+        k.Ended();
+      });
```
benh, line 35: Whenever we only need scope to be captured in a single lambda let's just use lambda captures directly. So in this case:

```suggestion
  return Stream<T>()
      .next([begin, end](auto& k) mutable {
        if (begin != end) {
          k.Emit(*(begin++));
        } else {
          k.Ended();
        }
      })
      .done([](auto& k) {
        k.Ended();
      });
```

Please update here and the rest of the review.

stout/iterate.h
```diff
@@ -0,0 +1,136 @@
...
+#include "stout/stream.h"
+
+////////////////////////////////////////////////////////////////////////
+
+namespace stout {
+namespace eventuals {
+
+////////////////////////////////////////////////////////////////////////
+
+template <typename Iter>
```
benh, line 15: s/Iter/Iterator/

Prefer full words please. Please update here and the rest of the review!

stout/iterate.h
```diff
@@ -0,0 +1,136 @@
...
+
+  struct ContainerData {
+    CIter it_;
+    Container data_;
+  };
+
+  using T = typename Container::value_type;
+
+  return Stream<T>()
+      .context(ContainerData{container.cbegin(), std::move(container)})
```
benh, line 55: This seems like it could be a bug, you're getting the beginning of a container but then moving that container? Are iterators valid even after the container has been moved? Seems like we want to store an `std::optional` of the type returned from `container.cbegin()` and then initialize it in the `.start()` callback. In that case we will need a `struct Data`, but let's call the fields `begin` and `container` to match the names used in the function scope.

Also your code will have to change slightly because you'll need to dereference `begin` since it'll be in a `std::optional`.

test/iterate.cc
```diff
@@ -0,0 +1,728 @@
...
+                k.Start(sum);
+              });
+  };
+
+  EXPECT_EQ(17, *s());
+  EXPECT_EQ(0, v.size());
+}
+
+TEST(Iterate, SetElemSum) {
+  std::set<int> set_ = {5, 12};
```
benh, line 101: Only use `_` suffix for private member variables of a struct/class.

test/iterate.cc
```diff
@@ -0,0 +1,728 @@
...
+              .ended([](auto& sum, auto& k) {
+                k.Start(sum);
+              });
+  };
+
+  EXPECT_EQ(5, *s());
+}
+
+TEST(Iterate, ArrayElemSumRvalue) {
+  std::array<int, 2> array_ = {5, 12};
```
benh, line 589: Unused?

test/iterate.cc
```diff
@@ -0,0 +1,728 @@
...
+                sum += value;
+                stream.Next();
+              })
+              .ended([](auto& sum, auto& k) {
+                k.Start(sum);
+              });
+  };
+
+  EXPECT_EQ(17, *s());
+  EXPECT_EQ(2, array_.size());
```
benh, line 625: What are you testing here? You moved the array so even though it has that size the data should be garbage. You've done this a couple of other places too ...

test/iterate.cc
```diff
@@ -0,0 +1,728 @@
...
+
+TEST(Iterate, VectorStringConcat) {
+  std::vector<std::string> v = {"Hello", "World", "!"};
+
+  auto s = [&]() {
+    return Iterate(v)
+        | Loop<std::string>()
+              .context(std::string(""))
+              .body([](auto& str, auto& stream, auto&& value) {
+                if (!str.empty())
```
benh, line 696: Always use '{' ... did clang-format not detect this?

test/iterate.cc
```diff
@@ -0,0 +1,728 @@
...
+}
+
+TEST(Iterate, VectorStringConcat) {
+  std::vector<std::string> v = {"Hello", "World", "!"};
+
+  auto s = [&]() {
+    return Iterate(v)
+        | Loop<std::string>()
+              .context(std::string(""))
+              .body([](auto& str, auto& stream, auto&& value) {
```
benh, line 695: How about s/str/result/?

onelxj: What do you mean? Need some more clarifications

PR #28, opened 2021-09-09
Move Buffer inside EventLoop + Hide filesystem::Request under filesystem::detail namespace.

FolMing:


stout/buffer.h
```diff
@@ -0,0 +1,114 @@
...
+
+////////////////////////////////////////////////////////////////////////
+
+namespace stout {
+namespace eventuals {
+
+////////////////////////////////////////////////////////////////////////
+
+// Moveable and Copyable.
+class Buffer {
```
benh, line 16: Let's actually put this into `EventLoop`, similar to how `Clock` is in there. It does make `EventLoop` a "big" class but we can pull out function implementations later if we need/want. I don't want to pollute the eventuals namespace with `Buffer` and this `Buffer` is really an `EventLoop::Buffer`.

stout/filesystem.h
```diff
@@ -15,6 +16,10 @@ namespace filesystem {
 
 ////////////////////////////////////////////////////////////////////////
 
+namespace detail {
+
+////////////////////////////////////////////////////////////////////////
+
 // Moveable, not Copyable.
 class Request {
```
benh, line 24: How about moving this into `EventLoop::Request` too?

FolMing: There are different types of requests in libuv. I can name at least two right now: uv_fs_t and uv_write_t. They are meant for different things and it's definitely not a good idea to move them into the EventLoop

benh: Ok, how about we create a `filesystem` namespace instead of `defailt` and move `Request` to `filesystem::Request` instead of `detail::Request`. `Request` is very generic and will pollute the `detail` namespace unless it's the only `Request` we ever imagine having in `stout::eventuals`, which is dubious. I also think we'll move more and more away from a generic `detail` namespace overtime. Thanks!

FolMing: It was previously `filesystem::Request` and this PR moves it into `filesystem::detail::Request`.

benh:
@FolMing please squash/fixup dead commits in the future, in this PR we don't want the 'Move buffer to single file' commit.

benh:
Committed in 689b79efb43c725cbe25fb5bbef96279831fb35b and 17446851ced4fa998629030c61a9cfa6f03ea9ae.

@FolMing also cleaned up some naming in: f787be2beefa7ac60d460502e5799ed5ab0003d7

PR #29, opened 2021-09-10
Add Collect wrapper

onelxj:


test/collect.cc
```diff
@@ -0,0 +1,45 @@
...
+  auto s = [&]() {
+    return Iterate(v)
+        | Collect<std::set<int>>([](int x) {
+             return x != 5;
+           });
+  };
+
+  std::set<int> res = *s();
+
+  EXPECT_EQ(12, *res.begin());
```
FolMing, line 43: `res.at(0)`

test/collect.cc
```diff
@@ -0,0 +1,45 @@
...
+  auto s = [&]() {
+    return Iterate(v)
+        | Collect<std::vector<int>>([](int x) {
+             return x != 5;
+           });
+  };
+
+  std::vector<int> res = *s();
+
+  EXPECT_EQ(12, res[0]);
```
FolMing, line 27: `res.at(0)` to also check if you're not OOB.

FolMing (changes requested):

stout/collect.h
```diff
@@ -0,0 +1,75 @@
...
+        stream.Next();
+      })
+      .ended([](auto& data, auto& k) {
+        k.Start(std::move(data.content_));
+      });
+}
+
+template <typename Container>
+auto TakeFirstAsContainer() {
+}
```
benh, line 69: Dead code?

stout/collect.h
```diff
@@ -0,0 +1,75 @@
...
+
+////////////////////////////////////////////////////////////////////////
+
+template <typename Container, typename F>
+auto Collect(F func) {
+  using T = typename Container::value_type;
+
+  struct DataContainer {
+    std::function<bool(T)> func_;
+    Container content_;
```
benh, line 47: Underscores only for private members of a class/struct.

benh (commented):
This PR is implementing both 'Collect' and 'Filter'. Let's factor these into two separate things. Ultimately someone should be able to do:

```c++
auto e = []() {
  return Iterate(std::vector<int>{1, 2, 3, 4, 5})
      | Filter([](int& i) {
        return i % 2;
      })
      | Collect<std::vector<int>>();
};

std::vector<int> evens = *e();
```

Make sense? Let's just implement 'Collect' in this PR and then we can do 'Filter' in another PR.

FolMing:
This PR also adds libcurl rules. Pls remove

PR #30, opened 2021-09-13
Add Collect wrapper

onelxj:


stout/collect.h
```diff
@@ -0,0 +1,65 @@
...
+
+  struct DataContainer {
+    std::function<bool(T)> func;
+    Container content;
+  };
+
+  return Loop<Container>()
+      .context(DataContainer{std::move(f), Container()})
+      .body([](auto& data, auto& stream, auto&& value) {
+        if (data.func(value)) {
```
rjhuijsman, line 46: I'll echo what Ben said in [his comment on your previous PR](https://github.com/3rdparty/stout-eventuals/pull/29#pullrequestreview-751912487): let's not implement `Filter` functionality in `Collect`.

`Collect` should collect _all_ the values, it shouldn't ask the user whether a value should be collected or not. If the user wants only some of the values collected, they should first do a `Filter`, then a `Collect`.

rjhuijsman (changes requested):

test/collect.cc
```diff
@@ -0,0 +1,43 @@
...
+  std::set<int> v = {5, 12};
+
+  auto s = [&]() {
+    return Iterate(v)
+        | Collect<std::set<int>>();
+  };
+
+  std::set<int> res = *s();
+
+  ASSERT_TRUE(res.size() == 2);
```
rjhuijsman, line 40: I believe you can use:
```c++
ASSERT_EQ(res.size(), 2);
```

rjhuijsman (approved):

PR #33, opened 2021-09-13
Add operator += to Buffer + Rename std::string args

FolMing:


rjhuijsman:
The PR title describes the "what" of this PR; can you also give me some context explaining the "why"? I suggest doing that by giving the PR a description. 

Specifically:
1. Why is it a good idea to add `+=` to `Buffer`, and...
2. Why are we renaming these parameters?

rjhuijsman (commented):
(Oops, had to leave a comment through the review page to put the ball back in your court)

FolMing:
1. Operator += is needed for future use in curl write callback. It will help us store everything inside one Buffer object and after operation is completed we will return its contents.
2. `std::string data` seems a bit awkward to me especially if you look at the code inside those methods. It's a bit painful trying to figure out where it's `data` and where it's `data_`.

rjhuijsman (approved):

PR #36, opened 2021-09-14
Remove Lambda decsription from the README.md

onelxj:


FolMing:
`Lambda` functionality was moved into `Then`, pls change the description of `Then` to show how to use the new API.

PR #37, opened 2021-09-14
Move 'Lambda' Description to the 'Then' description

onelxj:


README.md
```diff
@@ -339,15 +339,13 @@ auto e = Eventual<T>()
...
     // Return some value ***synchronously**.
     return stringify(result);
-  };
+  });
 ```
 
-### `Then`
-
-When your continuation is ***asynchronous*** (i.e., you need to create another eventual based on the result of an eventual) but you *don't* need the explicit control that you have with an `Eventual` you can use `Then`:
+In many cases you can be even more implicit and just use a C++ lambda directly too:
```
FolMing, line 348: Is it still possible?

rjhuijsman: As Nikita points out, let's remove this section, since it's not currently correct.

rjhuijsman (approved):
Good to go as soon as the two comments are addressed!

rjhuijsman (approved):

PR #39, opened 2021-09-14
Use bash instead of shell for tools/bootstrap

FolMing:
On some platforms shell is symlinked to dash and it doesn't support double
brackets. Bash should be enough for bootstrap to work fine on most systems.


while-false:
LGTM!

while-false (approved):
LGTM!

PR #40, opened 2021-09-16
Add Filter wrapper Interface

onelxj:
New Filter eventual, that takes some lambda function F that split values from the previous stream.
Example of usage:

`Iterate(v)
    |    Filter([](int x){ return x % 2 == 0; })
    |    Collect<std::vector<int>>()
...`

All odd numbers will appear at vector after Collect

### DIFF
--- a/stout/filter.h
+++ b/stout/filter.h
@@ -1,0 +1,85 @@
+#pragma once
+
+#include "stout/eventual.h"
+#include "stout/stream.h"
+
+////////////////////////////////////////////////////////////////////////
+
+namespace stout {
+namespace eventuals {
+
+////////////////////////////////////////////////////////////////////////
+
+namespace detail {
+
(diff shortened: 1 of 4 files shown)


stout/filter.h
```diff
@@ -0,0 +1,83 @@
...
+    template <typename Arg>
+    using ValueFrom = Arg;
+
+    template <typename Arg, typename K>
+    auto k(K k) && {
+      return Continuation<K, F_, Arg>{std::move(k), std::move(f_)};
+    }
+
+    F_ f_;
+  };
```
rjhuijsman, line 64: I could use some comments here about why this is needed.

rjhuijsman (changes requested):

rjhuijsman (approved):
🎉

PR #42, opened 2021-09-21
Make default EventLoop be allocated on stack

FolMing:


stout/event-loop.cc
```diff
@@ -79,20 +79,34 @@ void EventLoop::Clock::Advance(const std::chrono::nanoseconds& nanoseconds) {
...
-EventLoop& EventLoop::Default(EventLoop* replacement) {
-  delete CHECK_NOTNULL(loop);
-  loop = CHECK_NOTNULL(replacement);
-  return Default();
+void EventLoop::DefaultReset() {
+  if (loop_created) {
+    loop->~EventLoop();
+    for (int i = 0; i < sizeof(EventLoop); i++) {
+      loop_memory[i] = 0;
+    }
```
rjhuijsman, line 104: As mentioned by Gorm on the call, [`memset`](https://www.cplusplus.com/reference/cstring/memset/) is probably the right tool to zero out memory.

However, do we need to zero out this memory? Won't the constructor for the new `EventLoop` overwrite it anyway?

stout/event-loop.h
```diff
@@ -202,7 +202,7 @@ class EventLoop : public Scheduler {
   //
   // NOTE: takes ownership of event loop pointer.
```
rjhuijsman, line 203: This comment is out of date now that `DefaultReset` doesn't take a pointer anymore.

rjhuijsman (approved):
Having thought about this, this is safe because the memory we're allocating it in is _static_ (it's not actually on the stack where it will get destroyed at some point, but alive for the lifetime of the program).

Approved! Please consider the two comments before merging.

stout/event-loop.cc
FolMing: `new EventLoop` - will not overwrite it.
`new EventLoop()` - will.

Changed the code to use the second one and removed that `for` loop.

stout/event-loop.h
FolMing: Fixed!

PR #43, opened 2021-09-21
Add TakeFirstN, TakeRange, TakeLastN API for the streams

onelxj:
Adding a new API to get values from streams.

usage example:
`Iterate(container) 
| TakeFirstN(5)
| Collect<std::set<int>()
...`

`Stream<int>()... 
| TakeRange(3, 10) // 3 - start position, 10 - amount of elements to get from the stream
| Loop<int>()
...`

`Iterate(container) 
| TakeLastN(5)
| Collect<std::set<int>()
...`

### DIFF
--- a/stout/take.h
+++ b/stout/take.h
@@ -1,0 +1,195 @@
+#pragma once
+
+#include <algorithm>
+#include <deque>
+
+#include "stout/eventual.h"
+#include "stout/filter.h"
+#include "stout/stream.h"
+
+namespace stout {
+namespace eventuals {
+namespace detail {
+
+////////////////////////////////////////////////////////////////////////
+
+struct _TakeLastN {
+  template <typename K_, typename Arg_>
+  struct Continuation : public TypeErasedStream {
+    // NOTE: explicit constructor because inheriting 'TypeErasedStream'.
+    Continuation(K_ k, size_t n)
+      : k_(std::move(k)),
+        n_(n) {}
+
+    void Start(TypeErasedStream& stream) {
+      stream_ = &stream;
+      k_.Start(*this);
+    }
+
+    template <typename... Args>
+    void Fail(Args&&... args) {
+      k_.Fail(std::forward<Args>(args)...);
+    }
+
+    void Stop() {
+      k_.Stop();
+    }
+
+    template <typename... Args>
+    void Body(Args&&... args) {
+      if (data_.size() == n_) {
+        data_.pop_front();
+      }
+      data_.push_back(std::forward<Args>(args)...);
+      stream_->Next();
+    }
+
+    // Should be called from the previous eventual (Stream)
+    // when the Stream is done, so we are ready to start
+    // streaming last values.
+    void Ended() {
+      // Handle the case where the stream contained fewer than `n_` values.
+      n_ = std::min(n_, data_.size());
+      ended_ = true;
+
+      auto value = std::move(data_.front());
+      data_.pop_front();
+      k_.Body(std::move(value));
+    }
+
+    void Next() override {
+      // When Next is called from the next eventual,
+      // the element should be taken from the stored stream.
+
+      // If the stream_ has not produced its values yet,
+      // make it do so by calling stream_.Next().
+      // When it has produced all its values we'll receive an Ended() call.
+      if (!ended_) {
+        stream_->Next();
+        return;
+      }
+      if (data_.empty()) {
+        // There are no more stored values, our stream has ended.
+        k_.Ended();
+        return;
+      }
+      auto value = std::move(data_.front());
+      data_.pop_front();
+      k_.Body(std::move(value));
+    }
+
+    void Done() override {
+      k_.Ended();
+    }
+
+    K_ k_;
+    size_t n_;
+    std::deque<Arg_> data_;
+    bool ended_ = false;
+
+    TypeErasedStream* stream_ = nullptr;
+  };
+
+  struct Composable {
+    template <typename Arg>
+    using ValueFrom = Arg;
+
+    template <typename Arg, typename K>
+    auto k(K k) && {
+      return Continuation<K, Arg>{std::move(k), std::move(n_)};
+    }
+
+    size_t n_;
+  };
+};
+
+////////////////////////////////////////////////////////////////////////
+
+struct _TakeRange {
+  template <typename K_, typename Arg_>
+  struct Continuation {
+    void Start(TypeErasedStream& stream) {
+      stream_ = &stream;
+      k_.Start(stream);
+    }
+
+    template <typename... Args>
+    void Fail(Args&&... args) {
+      k_.Fail(std::forward<Args>(args)...);
+    }
+
+    void Stop() {
+      k_.Stop();
+    }
+
+    template <typename... Args>
+    // 'in_range_' needs to prevent calling Next
+    // when stream has already passed the set amount_ of elements.
+    void Body(Args&&... args) {
+      if (CheckRange()) {
+        in_range_ = true;
+        k_.Body(std::forward<Args>(args)...);
+      } else if (!in_range_) {
+        stream_->Next();
+      } else {
+        stream_->Done();
+      }
+    }
+
+    void Ended() {
+      k_.Ended();
+    }
+
+    bool CheckRange() {
+      bool result = i_ >= begin_ && i_ < begin_ + amount_;
+      ++i_;
+      return result;
+    }
+
+    K_ k_;
+    size_t begin_;
+    size_t amount_;
+    size_t i_ = 0;
+    bool in_range_ = false;
+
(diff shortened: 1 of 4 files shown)


stout/take.h
```diff
@@ -0,0 +1,191 @@
...
+    template <typename... Args>
+    void Body(Args&&... args) {
+      if (i_ == data_.size()) {
+        i_ = 0;
+        filled_ = true;
+      }
+      if (filled_) {
+        for (size_t j = 0; j < data_.size() - 1; ++j) {
+          data_[j] = data_[j + 1];
+        }
```
rjhuijsman, line 45: This process of moving everything up a step is `O(N)` copies ([big-O notation](https://www.freecodecamp.org/news/big-o-notation-why-it-matters-and-why-it-doesnt-1674cfa8a23c/)), and since this code will be called `N` times we have an overall `O(N^2)` performance here. We can do better.

The fix is to use a data structure that makes popping-from-front and pushing-onto-back an `O(1)` operation. For example [std::deque](https://en.cppreference.com/w/cpp/container/deque).

stout/take.h
```diff
@@ -0,0 +1,191 @@
...
+      }
+    }
+
+    void Done() override {
+      k_.Ended();
+    }
+
+    void SetValue(size_t pos, int v) {
+      data_[pos] = v;
+    }
```
rjhuijsman, line 81: Do we need a separate method for this, or would it be more readable to just write `data_[pos] = v` everywhere?

stout/take.h
```diff
@@ -0,0 +1,191 @@
...
+      k_.Stop();
+    }
+
+    template <typename... Args>
+    void Body(Args&&... args) {
+      if (i_ == data_.size()) {
+        i_ = 0;
+        filled_ = true;
+      }
+      if (filled_) {
```
rjhuijsman, line 42: Do we need to store a variable for this, or can we also just do `data_.size() == n_`?

If you think it's better for readability you can also consider a function named `Filled`:
```c++
bool Filled() {
  return data_.size() == n_;
}
```

Keeping track of state in multiple different variables is error-prone, hence my desire to avoid it.

stout/take.h
```diff
@@ -0,0 +1,191 @@
...
+        SetValue(data_.size() - 1, std::forward<Args>(args)...);
+      } else {
+        SetValue(i_++, std::forward<Args>(args)...);
+      }
+      stream_->Next();
+    }
+
+    void Ended() {
+      if (endedTriggered_) {
+        k_.Ended();
```
rjhuijsman, line 55: Under what circumstances is `Ended()` called twice on `_TakeLastN`? If never, can we remove this `if`-statement?

stout/take.h
```diff
@@ -0,0 +1,191 @@
...
+      } else {
+        i_ = 0;
+        endedTriggered_ = true;
+        k_.Body(data_[i_++]);
+      }
+    }
+
+    void Next() override {
+      if (!endedTriggered_) {
+        stream_->Next();
```
rjhuijsman, line 65: We could help me out with a comment here: the only time this `if`-statement is relevant is the very first time `Next()` is called, correct? That is: `Next()` is called from below only once before `Ended()` is called from above, right? That took me a while to figure out again, can we document that in a comment?

stout/take.h
```diff
@@ -0,0 +1,191 @@
...
+      } else {
+        if (i_ == data_.size()) {
+          k_.Ended();
+        } else {
+          k_.Body(data_[i_++]);
```
rjhuijsman, line 70: When using a `std::deque` you'll hopefully find that we don't need `i_` anymore, but instead just look how many items are still left in the queue.

stout/take.h
```diff
@@ -0,0 +1,191 @@
...
+      } else {
+        stream_->Done();
+      }
+    }
+
+    void Ended() {
+      k_.Ended();
+    }
+
+    bool filter_(Arg_ x) {
```
rjhuijsman, line 141: Internal functions are usually still named e.g. `Filter()`, but you can make them `private` if you want to make it extra clear that they're internal. The naming of e.g. `filter_` is reserved for class variables.

rjhuijsman: Why does this take `x`?

stout/take.h
```diff
@@ -0,0 +1,191 @@
...
+      return i_++ >= begin_ && i_ - 1 < begin_ + amount_;
```
rjhuijsman, line 142: In this line, it's pretty unclear when `i_` gets incremented. Is that before the second use of `i_` on that line, or after?

To make this code more readable let's explicitly increment `i_` on a separate line, either before or after.

stout/take.h
```diff
@@ -0,0 +1,191 @@
...
+  return detail::_TakeLastN::Composable{N};
+}
+
+auto TakeRange(size_t begin, size_t amount) {
+  return detail::_TakeRange::Composable{begin, amount};
+}
+
+auto TakeFirstN(size_t amount) {
+  return detail::_TakeRange::Composable{0, amount};
+}
```
rjhuijsman, line 184: I like how these share the same implementation, very elegant.

test/take.cc
```diff
@@ -0,0 +1,117 @@
...
+    return Iterate(v)
+        | TakeLastN(2)
+        | Collect<std::vector<int>>();
+  };
+
+  auto res = *s();
+
+  EXPECT_EQ(17, res.at(0));
+  EXPECT_EQ(3, res.at(1));
+  EXPECT_EQ(2, res.size());
```
rjhuijsman, line 31: Can you investigate how hard it would be to use [`googlemock`](https://github.com/google/googletest/tree/master/googlemock) in our tests? If that works, you can write this whole thing as:
```suggestion
  EXPECT_THAT(res, ElementsAre(17, 3));
```
... And it'll give you very nice errors if the test fails.

See [this StackOverflow answer](https://stackoverflow.com/questions/1460703/comparison-of-arrays-in-google-test) for some more examples of what you can do.

test/take.cc
```diff
@@ -0,0 +1,117 @@
...
+        | Filter([](int x) { return x % 2 == 1; })
+        | Collect<std::vector<int>>();
+  };
+
+  auto res = *s();
+
+  EXPECT_EQ(2, res.size());
+  EXPECT_EQ(5, res.at(0));
+  EXPECT_EQ(17, res.at(1));
+}
```
rjhuijsman, line 117: I'd also like to see tests for what happens if the stream runs out of elements before `N` is reached, for both `TakeRange` and `TakeLastN`.

rjhuijsman (changes requested):

stout/take.h
onelxj: since we have a pack of values, we can't do data_[pos] = v emplicit

stout/take.h
```diff
@@ -0,0 +1,191 @@
...
+  template <typename K_, typename Arg_>
+  struct Continuation : public TypeErasedStream {
+    // NOTE: explicit constructor because inheriting 'TypeErasedStream'.
+    Continuation(K_ k, size_t n)
+      : k_(std::move(k)),
+        N_(n) {}
+
+    void Start(TypeErasedStream& stream) {
+      stream_ = &stream;
+      data_ = std::vector<int>(N_);
```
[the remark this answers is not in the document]
onelxj: done

stout/take.h
onelxj: done

stout/take.h
onelxj: there && statement, in the case when the first part is false - the second part of statement couldn't be done, so we cant increment at the second part

onelxj: `bool result = i_ >= begin_ && i_ < begin_ + amount_;
      ++i_;
      return result;`

that looks weird, isn't it?

stout/take.h
```diff
@@ -0,0 +1,194 @@
...
+      bool result = i_ >= begin_ && i_ < begin_ + amount_;
+      ++i_;
+      return result;
+    }
+
+    K_ k_;
+    size_t begin_;
+    size_t amount_;
+    size_t i_ = 0;
+    bool inRange_ = false;
```
rjhuijsman, line 152: I believe our naming convention for variables is to use underscores between words, so:
```suggestion
    bool in_range_ = false;
```

[Here's an example](https://github.com/3rdparty/stout-grpc/blob/fef528dea479b09dec7bc49aa7535c0e87bca310/stout/grpc/server.h#L105) of where we already do that. Maybe Ben is following the [Google Style Guide](https://google.github.io/styleguide/cppguide.html#Variable_Names). 😃

rjhuijsman (changes requested):
This is looking great! Just some stylistic requests. 👍🏻

rjhuijsman (approved):
Good stuff 🎉

PR #45, opened 2021-09-22
Implement http::Get and http::Post methods

FolMing:
This PR introduces Get and Post methods to allow us make http requests to some specific URL. 

There are two crucial structures:

1. `PostFields` is simply a vector of pairs of string and string - the first one being the key and the second one being the value.
Here's an example of how to instantiate it:
```
PostFields fields = {
      {"title", "test"},
      {"body", "message"}};
```

2. `Response` is a return result of Get and Post methods. It has 2 public fields: `code` and `body`. They can be used later just like this:
```
Get("http://example.com") 
| Then([](auto&& response){
    do_smth_with_code(response.code);
    do_smth_with_body(response.body);
});
```

### DIFF
--- a/stout/http.h
+++ b/stout/http.h
@@ -1,0 +1,695 @@
+#pragma once
+
+#include <string>
+#include <vector>
+
+#include "curl/curl.h"
+#include "stout/event-loop.h"
+#include "stout/scheduler.h"
+
+////////////////////////////////////////////////////////////////////////
+
+namespace stout {
+namespace eventuals {
+namespace http {
+
+////////////////////////////////////////////////////////////////////////
+
+// Used for application/x-www-form-urlencoded case.
+// First string is key, second is value.
+// TODO(folming): switch to RapidJSON.
+using PostFields = std::vector<std::pair<std::string, std::string>>;
+
+////////////////////////////////////////////////////////////////////////
+
+struct Response {
+  long code;
+  std::string body;
+};
+
+////////////////////////////////////////////////////////////////////////
+
+enum class Method {
+  GET,
+  POST,
+};
+
+////////////////////////////////////////////////////////////////////////
+
+// Our own eventual for using libcurl with the EventLoop.
+//
+// The general algorithm:
+// 1. Create easy and multi handles. Set options for them.
+//    Add easy handle to multi handle with curl_multi_add_handle.
+//    TIMERFUNCTION is called to set a timer which will tell us when
+//    to perform checks on libcurl handles.
+//    SOCKETFUNCTION is called by using curl_multi_socket_action.
+//    We can pass specific socket descriptor to work with that particular
+//    socket or we can pass CURL_SOCKET_TIMEOUT to let libcurl call a function
+//    for each socket that is currently in use.
+// 2. Whenever SOCKETFUNCTION is called we check for events and set a poll
+//    handle for the particular socket. This poll handle is created on heap
+//    and that's why we put its pointer inside vector so that we can stop it
+//    when we have to interrupt the transfer.
+// 3. Whenever curl_multi_socket_action is called we can get an amount of
+//    remaining running easy handles. If this value is 0 then we read info
+//    from multi handle using check_multi_info lambda and clean everything up.
+struct _HTTP {
+  template <typename K_>
+  struct Continuation {
+    Continuation(
+        K_&& k,
+        EventLoop& loop,
+        std::string&& url,
+        Method& method,
+        std::chrono::nanoseconds& timeout,
+        PostFields&& fields_)
+      : k_(std::move(k)),
+        loop_(loop),
+        url_(std::move(url)),
+        method_(method),
+        timeout_(timeout),
+        fields_(std::move(fields_)),
+        start_(&loop, "HTTP (start)"),
+        interrupt_(&loop_, "HTTP (interrupt)") {}
+
+    Continuation(Continuation&& that)
+      : k_(std::move(that.k_)),
+        loop_(that.loop_),
+        url_(std::move(that.url_)),
+        method_(that.method_),
+        timeout_(that.timeout_),
+        fields_(std::move(that.fields_)),
+        start_(&that.loop_, "HTTP (start)"),
+        interrupt_(&that.loop_, "HTTP (interrupt)") {
+      CHECK(!that.started_ || !that.completed_) << "moving after starting";
+      CHECK(!handler_);
+    }
+
+    ~Continuation() {
+      CHECK(!loop_.InEventLoop())
+          << "attempting to destruct on the event loop "
+          << "is unsupported as it may lead to a deadlock";
+
+      if (started_) {
+        // Wait until we can destruct.
+        while (!closed_.load()) {}
+      }
+    }
+
+    void Start() {
+      CHECK(!started_ && !completed_);
+
+      loop_.Submit(
+          [this]() {
+            if (!completed_) {
+              started_ = true;
+
+              CHECK_EQ(0, uv_timer_init(loop_, &timer_));
+
+              uv_handle_set_data((uv_handle_t*) &timer_, this);
+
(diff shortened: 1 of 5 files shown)


rjhuijsman:
Whew, it's a biggie!

This all looks pretty reasonable to me, although I haven't yet dug into how `uv` and `curl` work.

When you're ready for a full review, let us know. To keep the review size manageable I suggest that we first review just the `Get()` method, become happy with that, and then continue to review the `Post()` method. Could still be in one PR (with multiple back-and-forths), or if you want to split it into two PRs that's bonus points.

stout/http.h
```diff
@@ -0,0 +1,580 @@
...
+                                        void* socketp) {
+              Data& data = *static_cast<Data*>(userp);
+              auto iterator = data.curl_context_map.find(sockfd);
+              int events = 0;
+
+              auto poll_callback = [](
+                                       uv_poll_t* handle,
+                                       int status,
+                                       int events) {
+                Data& data = *static_cast<Data*>(handle->data);
```
rjhuijsman, line 218: I find it hard to read all these lambdas inside lambdas inside lambdas inside lambdas. Could we make these all named functions elsewhere in the file?

stout/http.h
```diff
@@ -0,0 +1,592 @@
...
+
+    EasyHandle easy;
+    MultiHandle multi;
+    EasyHandle::WriteCallbackData context;
+  };
+
+  return loop.Schedule(
+      "http::Get",
+      Eventual<Result>()
+          .context(Data{loop, url, method, timeout, fields_string, loop})
```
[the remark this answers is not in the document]
FolMing: Have a look at an updated version and let me know if this makes it easier!

stout/http.h
```diff
@@ -0,0 +1,524 @@
...
+////////////////////////////////////////////////////////////////////////
+
+struct Result {
+  std::string data;
+  long response;
+};
+
+////////////////////////////////////////////////////////////////////////
+
+namespace detail {
```
rjhuijsman, line 32: I believe we've been moving `stout-eventuals` towards more class-based separation instead of using a `detail` namespace. How would you feel about having an `HTTP` class with a private `Call` method, e.g. `HTTP::Get()` (public) calling `HTTP::Call()` (private)?

stout/http.h
```diff
@@ -0,0 +1,524 @@
...
+                    k.Fail(curl_easy_strerror(CURLE_ABORTED_BY_CALLBACK));
+                    break;
+                }
+              }
+            };
+
+            static auto poll_callback = [](
+                                            uv_poll_t* handle,
+                                            int status,
+                                            int events) {
```
rjhuijsman, line 254: This looks much better with the lambda's split out like this!

stout/http.h
```diff
@@ -0,0 +1,524 @@
...
+              uv_timer_stop(data.timer_handle.handle());
+              int running_handles;
+              int flags = 0;
+
+              if (status < 0)
+                flags = CURL_CSELECT_ERR;
+              if (!status && (events & UV_READABLE))
+                flags |= CURL_CSELECT_IN;
+              if (!status && (events & UV_WRITABLE))
+                flags |= CURL_CSELECT_OUT;
```
rjhuijsman, line 266: Please use curly braces for all if-statements.

FolMing: Fixed!

stout/http.h
```diff
@@ -0,0 +1,529 @@
...
+  CURLM* multi_ = nullptr;
+};
+
+////////////////////////////////////////////////////////////////////////
+
+inline auto HttpCall(
+    EventLoop& loop,
+    const std::string& url,
+    const Method& method,
+    const std::chrono::nanoseconds& timeout = std::chrono::nanoseconds(0),
```
benh, line 151: To simplify this code let's handle timers outside. We can create an abstraction like:
```
auto e = DoUnlessTimeout(
    http::Get(...),
    std::chrono::milliseconds(500));
```
Which can then be used by multiple folks that want to be able to pass in an eventual that can be interrupted.

stout/http.h
```diff
@@ -0,0 +1,529 @@
...
+            // Start handling connection
+            CHECK_EQ(
+                curl_multi_add_handle(
+                    data.multi.handle(),
+                    data.easy.handle()),
+                CURLM_OK);
+          })
+          .interrupt([](auto& data, auto& k) {
+            data.context.should_interrupt
+                .store(true, std::memory_order_relaxed);
```
benh, line 459: How can you be sure that it hasn't already completed in which case you'd be calling `k.Start()` and `k.Stop()` here as well?

test/http.cc
```diff
@@ -0,0 +1,139 @@
...
+using eventuals::Terminate;
+
+using eventuals::http::Get;
+using eventuals::http::Post;
+using eventuals::http::PostFields;
+
+class HTTPTest : public EventLoopTest {};
+
+TEST_F(HTTPTest, GetFailTimeout) {
+  auto e = Get("http://example.com", std::chrono::milliseconds(1));
```
benh, line 23: If I understand correctly the plan is to replace all of these tests with the mock server but just wanted to point out that while unlikely the pattern you have in this tests and the ones below may lead to a flaky test and even if it's highly unlikely we still don't want to do it. But since I know the plan is to use the mock server I'm fine to start with these and then change to those in a subsequent PR.

test/http.cc
```diff
@@ -0,0 +1,139 @@
...
+
+  k.Start();
+
+  EventLoop::Waiter waiter(&EventLoop::Default(), "interrupt.Trigger()");
+
+  EventLoop::Default().Submit(
+      [&interrupt]() {
+        interrupt.Trigger();
+      },
+      &waiter);
```
benh, line 105: Can you add a comment explaining why doing this guarantees that this test won't be flaky?

test/http.cc
```diff
@@ -0,0 +1,139 @@
...
```
benh, line 134: Same request as above.

test/https.cc
```diff
@@ -0,0 +1,138 @@
...
+using eventuals::Interrupt;
+using eventuals::Terminate;
+
+using eventuals::http::Get;
+using eventuals::http::Post;
+using eventuals::http::PostFields;
+
+class HTTPSTest : public EventLoopTest {};
+
+TEST_F(HTTPSTest, GetFailTimeout) {
```
benh, line 21: If I'm reading this correctly these are all the same tests as in `http.cc`? Rather than having to copy the code everywhere (which is a burden when you make changes in one and forget to do it in another) you can use the [Value Parameterized Tests](https://github.com/google/googletest/blob/master/docs/advanced.md#value-parameterized-tests) feature. Then you can do something like:
```
// On Windows we can't test with the 'https' scheme because we don't have SSL support.
#ifdef _WIN32
const char* schemes[] = {"http://"};
#else
const char* schemes[] = {"http://", "https://"};
#endif

INSTANTIATE_TEST_SUITE_P(Schemes, HttpTest, testing::ValuesIn(schemes));
```
And then inside each test you can do:
```
std::string scheme = GetParam();
auto e = Get(scheme + "example.com", ...
```

benh (commented):
Looking really good @FolMing!

Biggest overall suggestion is if you can add some simple comments above some of the curl functions that will definitely help in understanding why we need to make so many of them!

Other than that I think once you've addressed these changes we should be good to go!

stout/http.h
```diff
@@ -0,0 +1,624 @@
...
+                        easy_),
+                    CURLM_OK);
+
+                curl_easy_cleanup(easy_);
+                curl_multi_cleanup(multi_);
+
+                k_.Stop();
+              } else {
+                k_.Stop();
+              }
```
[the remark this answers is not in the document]
FolMing: Yes! Copy-pasted code from signal and timer and completely overlooked this one!

benh (approved):
SHIP IT!

PR #52, opened 2021-09-23
Fix Filter.OddMapCollectFlow test on Windows

### LINKED ISSUE #48: Filter.OddMapCollectFlow test fails on Windows
### Description:
Filter.OddMapCollectFlow test fails on Windows with the following error:
```
test/filter.cc(98): error: Expected equality of these values:
  *res.begin()
    Which is: 6
  18
test/filter.cc(99): error: Expected equality of these values:
  *++res.begin()
    Which is: 18
  6
```

### Steps to reproduce:
Run the test on Windows.

onelxj:
Fix #48

### DIFF
--- a/test/filter.cc
+++ b/test/filter.cc
@@ -4,6 +4,7 @@
 #include <unordered_set>
 #include <vector>
 
+#include "gmock/gmock.h"
 #include "gtest/gtest.h"
 #include "stout/collect.h"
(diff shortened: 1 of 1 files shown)


rjhuijsman (approved):
Nice 🎉 

CC @while-false - one fewer flaky/broken test!

PR #54, opened 2021-09-24
Fix Signal implementation and re-enable tests

FolMing:
1. This PR adds a RAII implementation for signals inside the EventLoop to ensure the correct behavior.
Due to the design choices in libuv the handle is now allocated on heap and there's an added EventLoop::CloseCallback method which is called on the next iteration of uv_run to ensure de-allocation.

2. Added a check in EventLoop::~EventLoop to ensure that uv_close function returns 0 and not UV_EBUSY value or any other incorrect value.

### DIFF
--- a/stout/event-loop.h
+++ b/stout/event-loop.h
@@ -132,6 +132,53 @@ class EventLoop : public Scheduler {
     uv_buf_t buffer_ = {};
   };
 
+  class Signal {
+   public:
+    Signal(EventLoop& loop) {
+      handle_ = new uv_signal_t();
+      CHECK_EQ(
+          uv_signal_init(loop, handle_),
+          0);
+    }
+
+    Signal(const Signal& that) = delete;
+
+    Signal(Signal&& that) {
+      handle_ = that.handle_;
+      that.handle_ = nullptr;
+    }
+
+    ~Signal() {
+      if (handle_ == nullptr) {
+        return;
+      }
+
+      if (uv_is_active(reinterpret_cast<uv_handle_t*>(handle_))) {
+        uv_signal_stop(handle_);
+      }
+
+      uv_close(
+          reinterpret_cast<uv_handle_t*>(handle_),
+          EventLoop::CloseCallback);
+    }
+
+    // Adaptors to libuv functions.
+    uv_signal_t* handle() {
+      CHECK_NOTNULL(handle_);
+      return handle_;
+    }
+
+    uv_handle_t* base_handle() {
+      CHECK_NOTNULL(handle_);
+      return reinterpret_cast<uv_handle_t*>(handle_);
+    }
+
+   private:
+    // NOTE: Determines the ownership over the handle.
+    // If pointer is nullptr then it was transfered to another object.
+    uv_signal_t* handle_ = nullptr;
+  };
+
   class Clock {
    public:
     Clock(const Clock&) = delete;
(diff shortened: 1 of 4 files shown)


stout/event-loop.h
```diff
@@ -132,6 +132,43 @@ class EventLoop : public Scheduler {
...
+
+    Signal(const Signal& that) = delete;
+
+    Signal(Signal&& that) {
+      handle_ = that.handle_;
+      that.handle_ = nullptr;
+    }
+
+    ~Signal() {
+      if (handle_) {
```
rjhuijsman, line 152: Under what circumstance would `handle_` ever be null here?

stout/event-loop.h
```diff
@@ -132,6 +132,43 @@ class EventLoop : public Scheduler {
...
+        uv_close(
+            reinterpret_cast<uv_handle_t*>(handle_),
+            EventLoop::CloseCallback);
+      }
+    }
+
+    operator uv_signal_t*() {
+      CHECK(handle_);
+      return handle_;
+    }
```
rjhuijsman, line 166: Do we need this operator overload? If not, I prefer to avoid operator overloads generally.

We could replace it with a named `handle()` method?

stout/event-loop.h
```diff
@@ -256,6 +293,10 @@ class EventLoop : public Scheduler {
  private:
   void Prepare();
 
+  static void CloseCallback(uv_handle_t* handle) {
+    delete handle;
+  }
```
rjhuijsman, line 298: Does it make more sense to put this callback inside `Signal`?

stout/signal.h
```diff
@@ -35,25 +37,22 @@ inline auto Signal(stout::eventuals::EventLoop& loop, const int signum) {
...
-                  data.signum);
-
-              if (error) {
-                uv_close((uv_handle_t*) (&data.signal), nullptr);
-                k.Fail(uv_err_name(error));
-              }
+
+            uv_handle_set_data(
+                reinterpret_cast<uv_handle_t*>(
+                    data.signal.operator uv_signal_t*()),
```
rjhuijsman, line 43: This is... not pleasant to read. 😆 Hopefully renaming that method makes this easier to read too.

rjhuijsman (changes requested):
Nicely done. RAII for the win! 🥇 

Some readability changes requested.

stout/event-loop.h
FolMing: Definitely not. There are many more handles (that will be implemented in the near future) which use the same callback.

stout/event-loop.h
FolMing: I would also like to avoid doing this:
`uv_handle_set_data(reinterpret_cast<uv_handle_t*>(data.signal.handle()), &data)`

The problem is that you can't just pass the underlying handle ptr to certain libuv functions without converting them to uv_handle_t*.

How about `handle()` and `base_handle()`?

stout/event-loop.h
FolMing: Added a comment explaining the use of nullptr there!

stout/event-loop.h
```diff
@@ -132,6 +132,51 @@ class EventLoop : public Scheduler {
...
+
+        uv_close(
+            reinterpret_cast<uv_handle_t*>(handle_),
+            EventLoop::CloseCallback);
+      }
+    }
+
+    // Adaptors to libuv functions.
+    uv_signal_t* handle() {
+      CHECK(handle_);
```
rjhuijsman, line 165: Here too (and below) I'd like to be explicit about the type:
```suggestion
      CHECK_NOTNULL(handle_);
```

rjhuijsman (approved):
🎉

PR #55, opened 2021-09-24
Update commit hash for external use

while-false:
Bump version of stout-eventuals to a recent version containing the recent Get and Post additions.


FolMing:
If we are doing this, can you also change the README.md with a proper commit hash?

rjhuijsman (approved):

PR #56, opened 2021-09-24
Fix Timer implementation and re-enable tests

FolMing:
This PR adds a RAII implementation for timers inside the EventLoop to ensure the correct behavior.
Due to the design choices in libuv the handle is now allocated on heap and is using an EventLoop::CloseCallback method which is called on the next iteration of uv_run to ensure de-allocation.

### DIFF
--- a/stout/event-loop.h
+++ b/stout/event-loop.h
@@ -179,6 +179,53 @@ class EventLoop : public Scheduler {
     uv_signal_t* handle_ = nullptr;
   };
 
+  class Timer {
+   public:
+    Timer(EventLoop& loop) {
+      handle_ = new uv_timer_t();
+      CHECK_EQ(
+          uv_timer_init(loop, handle_),
+          0);
+    }
+
+    Timer(const Timer& that) = delete;
+
+    Timer(Timer&& that) {
+      handle_ = that.handle_;
+      that.handle_ = nullptr;
+    }
+
+    ~Timer() {
+      if (handle_ == nullptr) {
+        return;
+      }
+
+      if (uv_is_active(reinterpret_cast<uv_handle_t*>(handle_))) {
+        uv_timer_stop(handle_);
+      }
+
+      uv_close(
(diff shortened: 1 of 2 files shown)


test/BUILD.bazel
```diff
@@ -14,7 +14,7 @@ cc_test(
         "task.cc",
         "closure.cc",
         # "parallel.cc",
-        # "timer.cc",
+        "timer.cc",
```
while-false, line 17: Excited to see the tests back! Thanks!

while-false (approved):
LGTM. Thanks for adding the comments!

stout/event-loop.h
```diff
@@ -177,6 +177,51 @@ class EventLoop : public Scheduler {
...
+
+    Timer(const Timer& that) = delete;
+
+    Timer(Timer&& that) {
+      handle_ = that.handle_;
+      that.handle_ = nullptr;
+    }
+
+    ~Timer() {
+      if (handle_) {
```
rjhuijsman, line 197: As on the other PR, I'd love for us to do type-safe checks:
```suggestion
      if (handle_ != nullptr) {
```

stout/event-loop.h
```diff
@@ -177,6 +177,51 @@ class EventLoop : public Scheduler {
...
+
+        uv_close(
+            reinterpret_cast<uv_handle_t*>(handle_),
+            EventLoop::CloseCallback);
+      }
+    }
+
+    // Adaptors to libuv functions.
+    uv_timer_t* handle() {
+      CHECK(handle_);
```
rjhuijsman, line 210: As on the other PR, I suggest:
```suggestion
      CHECK_NOTNULL(handle_);
```

test/BUILD.bazel
rjhuijsman: 🎉

rjhuijsman (approved):
Nice! Thank you!

PR #58, opened 2021-09-27
Added check_format_code.yml

ArthurBandaryk:
This .yml file helps to verify that a repo's formatting matches our
specifications (.clang-format file in the root directory), but doesn't
automatically fix any issues.

For now I need to figure out how to make this action automatically
check if every line of the code exceeds 80 characters and make it work
with BUILD.bazel and WORKSPACE.bazel files.

### DIFF
--- a/.github/workflows/check_format_code.yml
+++ b/.github/workflows/check_format_code.yml
@@ -1,0 +1,16 @@
+name: C/C++ Style Check
+
+on: [push, pull_request]
+jobs:
+  formatting-check:
+    name: Formatting Check
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run bootstrap
+      run: cd tools/ && ./bootstrap
+    - name: Run clang-format style check for C/C++ programs.
+      uses: jidicula/clang-format-action@v4.0.0
+      with:
+        clang-format-version: '12'
+        check-path: '.'
--- a/test/parallel.cc
+++ b/test/parallel.cc
@@ -32,11 +32,11 @@ TEST(ParallelTest, Succeed) {
                  k.Ended();
                })
         | Parallel([]() {
-            return Then([](int i) {
-              std::this_thread::sleep_for(std::chrono::milliseconds(100));
-              return i + 1;
-            });
-          })
+             return Then([](int i) {
+               std::this_thread::sleep_for(std::chrono::milliseconds(100));
+               return i + 1;
+             });
+           })
         | Reduce(
                std::set<int>{2, 3, 4, 5, 6},
                [](auto& values) {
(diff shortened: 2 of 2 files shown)


.github/workflows/check_format_code.yml
```diff
@@ -0,0 +1,32 @@
...
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+jobs:
+  formatting-check:
+    name: Formatting Check
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
while-false, line 15: I think the matrix setup is overkill. I think we generally want all our code to be checked independant on path.

.github/workflows/check_format_code.yml
```diff
@@ -0,0 +1,32 @@
+name: Style Check
+
+on:
+  push:
+    branches: [main]
```
while-false, line 5: I'd like this check on any push, not just pushes to main

while-false (changes requested):

while-false:
On [the clang-format-action page](https://github.com/jidicula/clang-format-action) it seems that all C/C++ headers are auto magically checked. We would like (in other places) to run checks of `.proto` files too. This might be worth keeping in mind when you hack around: if we can specify extensions we want to check that might get us a long way here.

.github/workflows/check_format_code.yml
rjhuijsman: Is there a way for us to specify "check _all_ code" with this GitHub Action? The [examples of this Action](https://github.com/jidicula/clang-format-action) only show us how to give an explicit list of paths (using matrix), not "all paths".

If it's not obvious how to make the extension do all paths, I prefer to just enumerate them exactly rather than find a completely different solution.

.github/workflows/check_format_code.yml
```diff
@@ -0,0 +1,32 @@
...
+  formatting-check:
+    name: Formatting Check
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        path:
+          - check: 'stout'
+            exclude: ''
+          - check: 'test'
+            exclude: 'BUILD.bazel'
```
rjhuijsman, line 20: Do we need to exclude `BUILD.bazel`? Does clang-format try to check it if we don't?

ArthurBandaryk: Actually we do not need to exclude `BUILD.bazel` file, cause this action will check the following list of extensions:
`.h , .H, .hpp, .hh, .h++, .hxx`
`.c, .C, .cpp, .cc, .c++, .cxx`
So I will remove this file from the `exclude` section.

ArthurBandaryk:
Removed `matrix` setup section and made this action do the check of our code  independent on path, so all .cc/.h files will be checked.

rjhuijsman (approved):

rjhuijsman:
Awesome, this looks great. As discussed, let's fix the formatting issues that this tool detects in this PR, and then let's submit it and continue with buildifier and prototool in follow-up PRs.

while-false (approved):
Looks good to me!

PR #62, opened 2021-09-28
Add Range interface

onelxj:
`Range` interface gives the stream that iterates through the sequence of given ints.
Examples:
```
Range(10, 20)
| Map(Then([](int x){ return x + 5; }))
| Collect<std::vector<int>>
```

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,7 +1,7 @@
 load("@rules_cc//cc:defs.bzl", "cc_library")
 
 cc_library(
-    name = "eventuals",
+    name = "eventuals-base",
     hdrs = [
         "stout/undefined.h",
         "stout/sequence.h",
@@ -30,14 +30,9 @@ cc_library(
         "stout/closure.h",
         "stout/compose.h",
         "stout/catch.h",
-        "stout/event-loop.h",
-        "stout/timer.h",
         "stout/head.h",
-        "stout/filesystem.h",
         "stout/parallel.h",
         "stout/iterate.h",
-        "stout/dns-resolver.h",
-        "stout/signal.h",
         "stout/collect.h",
         "stout/filter.h",
         "stout/take.h",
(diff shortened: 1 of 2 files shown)


stout/range.h
```diff
@@ -0,0 +1,92 @@
...
+////////////////////////////////////////////////////////////////////////
+
+} // namespace detail
+
+////////////////////////////////////////////////////////////////////////
+
+auto Range(int from, int to) {
+  return detail::_Range::Composable<int>{from, to};
+}
+
```
rjhuijsman, line 86: Shall we also add a:
```c++
auto Range(int to) {
```
... Which assumes that `from` is `0`? That matches behavior on e.g. Python's `for i in range(10)`.

onelxj: I guess we also can add a `step` as a third parameter:
`Range(0,10,2)` -> 0 2 4 6 8

rjhuijsman: That's a good idea! Let's offer all three.

benh: I like it!

stout/range.h
```diff
@@ -0,0 +1,92 @@
...
+#include "stout/stream.h"
+
+////////////////////////////////////////////////////////////////////////
+
+namespace stout {
+namespace eventuals {
+
+////////////////////////////////////////////////////////////////////////
+
+namespace detail {
```
benh, line 14: We're slowing moving away from using the `detail` namespace and just using the prefixed `_` instead, so please drop!

stout/range.h
```diff
@@ -0,0 +1,92 @@
...
+    template <typename Arg, typename K>
+    auto k(K k) && {
+      return Continuation<K, Arg>{std::move(k), from_, to_};
+    }
+
+    int from_;
+    int to_;
+  };
+};
+
```
benh, line 75: Extra newline you can drop.

stout/range.h
```diff
@@ -0,0 +1,92 @@
...
+    void Done() override {
+      k_.Ended();
+    }
+
+    K_ k_;
+    int to_;
+    int from_;
+  };
+
+  template <typename T>
```
benh, line 61: Since you're not using this functionality right now (you're always using 'int') and it's really trivial to add later let's drop the template here.

BUILD.bazel
```diff
@@ -41,6 +41,7 @@ cc_library(
         "stout/collect.h",
         "stout/filter.h",
         "stout/take.h",
+        "stout/range.h",
```
benh, line 44: Sorry but you'll need to update this because we broke everything into multiple sub-libraries.

benh (changes requested):

stout/range.h
```diff
@@ -0,0 +1,93 @@
...
+        k_.Body(temp);
+      }
+    }
+
+    void Done() override {
+      k_.Ended();
+    }
+
+    K_ k_;
+    int to_;
```
rjhuijsman, line 56: Looks like we can declare this `const`.

stout/range.h
```diff
@@ -0,0 +1,93 @@
...
+
+struct _Range {
+  template <typename K_, typename Arg_>
+  struct Continuation : public detail::TypeErasedStream {
+    // NOTE: explicit constructor because inheriting 'TypeErasedStream'.
+    Continuation(K_ k, int from, int to, int step)
+      : k_(std::move(k)),
+        to_(to),
+        from_(std::min(from, to)),
+        step_(step) {}
```
rjhuijsman, line 22: Small thing: let's keep the order of `from`, `to` and `step` consistent - same order in the constructor, parameter lists, and definition.

stout/range.h
```diff
@@ -0,0 +1,93 @@
...
+    using ValueFrom = int;
+
+    template <typename Arg, typename K>
+    auto k(K k) && {
+      return Continuation<K, Arg>{std::move(k), from_, to_, step_};
+    }
+
+    int from_;
+    int to_;
+    int step_;
```
rjhuijsman, line 72: Looks like we can declare these `const`?

test/range.cc
```diff
@@ -0,0 +1,47 @@
...
+
+  EXPECT_THAT(*s, ElementsAre(-2, -1, 0, 1));
+}
+
+TEST(Range, DefaultFrom) {
+  auto s = Range(3)
+      | Collect<std::vector<int>>();
+
+  EXPECT_THAT(*s, ElementsAre(0, 1, 2));
+}
```
rjhuijsman, line 40: Let's also have a version of this test that has a negative `to`.

test/range.cc
```diff
@@ -0,0 +1,47 @@
...
+
+TEST(Range, SpecifiedStep) {
+  auto s = Range(0, 10, 2)
+      | Collect<std::vector<int>>();
+
+  EXPECT_THAT(*s, ElementsAre(0, 2, 4, 6, 8));
+}
```
rjhuijsman, line 47: Let's also have a version of this test that has a negative `step`.

rjhuijsman (changes requested):

PR #63, opened 2021-09-28
Ensure EventLoop is not interrupted after or during destruction

### LINKED ISSUE #57: Async handle is not closed properly
https://github.com/3rdparty/stout-eventuals/pull/55/checks?check_run_id=3717847817

^ this test run shows an issue with the async handle used inside EventLoop:
```
[----------] 16 tests from FilesystemTest
[ RUN      ] FilesystemTest.MakeDirectoryFail
WARNING: Logging before InitGoogleLogging() is written to STDERR
W20210927 08:13:52.472527  2992 event-loop.cc:138] destructing EventLoop with active handles: 
[-AI] async    00007FF66A1E53C8
[R--] async    00007FF66A1E5520
F20210927 08:13:52.472527  2992 event-loop.cc:142] Check failed: uv_loop_close(&loop_) == 0 (-4082 vs. 0) 
```

benh:
Resolves #57.


benh:
Adding a little more context here for posterity: the hypothesis on the bug we saw in #57 is that the thread that gets created and detached in `EventLoopTest.InterruptTimer` did an `EventLoop::Submit()` and was context switched out before it did the `uv_async_send()` meanwhile the main thread calls `EventLoop::Run()` and executes the lambda passed to `EventLoop::Submit()` and since the thread was detached the main thread doesn't wait for it so when it gets context switched back in it tries to do the `uv_async_send()` but the `EventLoop` is already getting destructed. This PR fixes such a bug even if #57 is not an exact incarnation of this hypothesis.

PR #64, opened 2021-09-28
Wait for event loop to no longer be alive on destruction

benh:


stout/event-loop.cc
```diff
@@ -142,13 +142,26 @@ EventLoop::~EventLoop() {
...
-  }
+  CHECK(alive) << "should still have prepare and async handles to close";
+
+  do {
+    alive = uv_run(&loop_, UV_RUN_NOWAIT);
+
+    LOG_IF_EVERY_N(WARNING, alive, 100000)
+        << "destructing EventLoop with active handles:\n"
+        << [&]() {
+             // NOTE: lambda only gets executed every 100 instances
```
FolMing, line 158: 100k?

FolMing (approved):
LGTM!

stout/event-loop.cc
benh: Thanks!

FolMing (approved):
Great stuff! Really like the way you try to simulate the output from `uv_print_all_handles`!

PR #65, opened 2021-09-29
Timer as own eventual type

benh:


stout/event-loop.h
```diff
@@ -255,46 +255,27 @@ class EventLoop : public Scheduler {
...
         ~Continuation() {
-          if (timer_ != nullptr) {
-            if (uv_is_active(handle())) {
-              uv_timer_stop(timer());
-            }
-
-            uv_close(handle(), [](uv_handle_t* handle) {
-              delete handle;
-            });
+          if (started_) {
```
benh, line 271: Make sure we're not in the event loop.

FolMing (approved):
LGTM, but without context this is very hard to read.

PR #67, opened 2021-09-29
Make `Build` and `Test` checks triggered

ArthurBandaryk:
For now this checks will be triggered on any push on any branch.


while-false (approved):

while-false (approved):
Very nice Arthur, thanks for adding in the comments!

PR #69, opened 2021-09-29
Add Range streaming Interface

onelxj:
`Range` interface gives the stream that iterates through the sequence of given ints.
Examples:
```
Range(10, 20)
| Map(Then([](int x){ return x + 5; }))
| Collect<std::vector<int>>
```

### DIFF
--- a/stout/range.h
+++ b/stout/range.h
@@ -1,0 +1,94 @@
+#pragma once
+
+#include "stout/stream.h"
+
+////////////////////////////////////////////////////////////////////////
+
+namespace stout {
+namespace eventuals {
+
+////////////////////////////////////////////////////////////////////////
+
+struct _Range {
+  template <typename K_, typename Arg_>
+  struct Continuation : public detail::TypeErasedStream {
+    // NOTE: explicit constructor because inheriting 'TypeErasedStream'.
+    Continuation(K_ k, int from, int to, int step)
+      : k_(std::move(k)),
+        from_(from),
+        to_(to),
+        step_(step) {}
+
+    void Start() {
+      k_.Start(*this);
+    }
+
+    template <typename... Args>
+    void Fail(Args&&... args) {
+      k_.Fail(std::forward<Args>(args)...);
+    }
+
+    void Stop() {
+      k_.Stop();
+    }
+
+    void Ended() {
+      k_.Ended();
+    }
+
+    void Next() override {
+      if (from_ == to_
+          || step_ == 0
+          || (from_ > to_ && step_ > 0)
+          || (from_ < to_ && step_ < 0)) {
+        k_.Ended();
+      } else {
+        int temp = from_;
+        from_ += step_;
+        k_.Body(temp);
+      }
+    }
+
+    void Done() override {
+      k_.Ended();
+    }
+
+    K_ k_;
+    int from_;
+    const int to_;
+    const int step_;
+  };
+
+  struct Composable {
+    template <typename>
+    using ValueFrom = int;
+
+    template <typename Arg, typename K>
+    auto k(K k) && {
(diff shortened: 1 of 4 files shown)


stout/range.h
```diff
@@ -0,0 +1,102 @@
...
+    const int step_;
+  };
+};
+
+inline auto Range(int to) {
+  if (to >= 0) {
+    return _Range::Composable{0, to, 1};
+  } else {
+    return _Range::Composable{0, 0, 0};
+  }
```
rjhuijsman, line 81: Right now, we essentially check for (and deal with) this scenario in three different ways in our three different `Range(...)` methods. For readability and make-sure-it-always-works-ness, it seems better to deal with invalid ranges only once: in `Next()`. If a range is found to be invalid, simply call `Ended()`.

Can you give that a try and see how it looks?

rjhuijsman (changes requested):

stout/range.h
```diff
@@ -0,0 +1,94 @@
...
+    void Stop() {
+      k_.Stop();
+    }
+
+    void Ended() {
+      k_.Ended();
+    }
+
+    void Next() override {
+      if ((from_ == to_) || (step_ == 0) || (from_ > to_ && step_ > 0) || (from_ < to_ && step_ < 0)) {
```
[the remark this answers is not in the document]
rjhuijsman: The `(from_ == to_)` and `(step == 0)` don't need to be in parentheses.

stout/range.h
```diff
@@ -0,0 +1,91 @@
...
+        k_.Ended();
+      } else {
+        int temp = from_;
+        from_ += step_;
+        k_.Body(temp);
+      }
```
rjhuijsman, line 46: Instead of having to indent this all in an `else`, consider adding a `return` to the `if` block.

rjhuijsman (approved):
[🎵 Let's check in that `(from, to, step)`🎵](https://www.youtube.com/watch?v=iBHNgV6_znU)

rjhuijsman:
(But actually, please give @benh a chance to take a look before you check it in 🙂 )

stout/range.h
```diff
@@ -0,0 +1,96 @@
...
+          || step_ == 0
+          || (from_ > to_ && step_ > 0)
+          || (from_ < to_ && step_ < 0)) {
+        k_.Ended();
+        return;
+      }
+
+      int temp = from_;
+      from_ += step_;
+      k_.Body(temp);
```
benh, line 50:
```suggestion
      if (from_ == to_
          || step_ == 0
          || (from_ > to_ && step_ > 0)
          || (from_ < to_ && step_ < 0)) {
        k_.Ended();
      } else {
        int temp = from_;
        from_ += step_;
        k_.Body(temp);
      }
```

benh: In general let's avoid using `return;` because having complete `if () {} else {}` blocks makes it harder to introduce control flow bugs when we refactor!

benh (approved):
LGTM but please address the comments and suggestions!

stout/range.h
rjhuijsman: Hahaha, I'd asked Zakhar to do the exact opposite here previously. 😆 

I like using `return` because it avoids indentation. In general, I like structuring code as:
```
if (something is not normal) {
  handle it;
  return;
}
if (something else is off) {
  handle it;
  return;
}
handle normal case;
with more code;
than the abortive cases;
```
As opposed to:
```
if (something is normal) {
  if (something else is not off) {
    handle normal case;
    with more code;
    than the abortive cases;
  } else {
    handle it;
  }
} else {
  handle it;
}
```
I find the upper version to be much more readable.

Thoughts?

onelxj: 1 more vote for using `if else` from me

PR #71, opened 2021-09-30
Add ComposedLoop interface

onelxj:
`ComposedLoop` is an instance that provides to create streams with inner loops

Let's imagine that a single common stream represents a one-level loop like

`Range(n)`  equals to `for (int i =0; i < n; ++i){...}`

When there is a composition of that Range with `ComposedLoop([](int x) { return Range(k); })` it equals to 

```
for (int i = 0; i < n; ++i)
{
    for (int j = 0; k < 2; ++k){...}
}
```

In other words - it adds the inner loop inside the loop above each time you compose it with `ComposedLoop`

The idea - pass a `Lamda` to the `ComposedLoop` that produces a `Stream` instance (for now it instance that inherits from `TypeErasedStream` structure) and this `Stream` will be the inner loop for the stream above.


stout/composed-loop.h
```diff
@@ -0,0 +1,154 @@
...
+
+    void Stop() {
+      k_.Stop();
+    }
+
+    template <typename... Args>
+    void Body(Args&&... args) {
+      // Executes when Body is called from the above-level stream.
+      if (innerStream_ == nullptr) {
+        inner_.emplace(Build(operator|(f_(std::forward<Args>(args)...), ComposedLoopRouter(this))));
```
rjhuijsman, line 83: Our formatting check doesn't enforce the 80 character rule yet, so please manually check that all the lines are <80 characters.

stout/composed-loop.h
```diff
@@ -0,0 +1,154 @@
...
+    auto k(K k) && {
+      return Continuation<K, F, Arg>{std::move(k), std::move(f)};
+    }
+
+    F f;
+  };
+};
+
+template <typename F>
+auto ComposedLoop(F f) {
```
rjhuijsman, line 145: I tried to put myself in a "first time seeing this" brain-mode. I found `ComposedLoop` to not be immediately obvious as a name, given what this thing does. There's a stream within a stream. What does that have to do with "loop"? And isn't everything in eventuals "composed"? 

How would you feel about renaming to... `InnerStream`? `NewStream`? `ForEach`?

test/composed-loop.cc
```diff
@@ -0,0 +1,66 @@
...
+    return Range(2)
+        | ComposedLoop([](int x) { 
+            std::vector<int> v = {1, 2, 3};
+          return Iterate(std::move(v)); })
+        | Map(Then([](int x) { return x + 1; }))
+        | Collect<std::vector<int>>();
+  };
+
+  EXPECT_THAT(*s(), ElementsAre(2, 3, 4, 2, 3, 4));
+}
```
rjhuijsman, line 66: Since tests are documentation too, a test that demonstrates how the `x` being passed to `ComposedLoop` can be used would be nice.

stout/composed-loop.h
```diff
@@ -0,0 +1,154 @@
...
+        return;
+      }
+      innerStream_->Next();
+    }
+
+    void Done() override {
+      Ended();
+    }
+
+    using Left_ = typename std::invoke_result<F_, int>::type;
```
rjhuijsman, line 117: Will `F_` always be invoked with an `int`?

test/composed-loop.cc
rjhuijsman: Also good to have: some test that does a ComposedLoop with something that isn't an `int`.

rjhuijsman: Also good to have: some test that shows how to use a ComposedLoop _inside_ a ComposedLoop, so both `x` and `y` are accessible at the same time, for example when you want to do the equivalent of:
```c++
for (int x = 0; x < 2; x++) {
  for (int y = 10; y < 12; y++) {
    emit(x*y);
  }
}
```

rjhuijsman (changes requested):

stout/composed-loop.h
onelxj: I like InnerStream?
What do you think, @benh ?

stout/composed-loop.h
onelxj: whoops, my fault

stout/composed-loop.h
benh: Hmm, how about `StreamForEach`?

It reads pretty nice I think:

```
Iterate(std::vector<int>{1, 2, 3})
    | StreamForEach([](int i) {
      return Range(0, i);
    })
```

test/composed-loop.cc
```diff
@@ -0,0 +1,66 @@
...
+using stout::eventuals::Collect;
+using stout::eventuals::ComposedLoop;
+using stout::eventuals::Iterate;
+using stout::eventuals::Map;
+using stout::eventuals::Range;
+using stout::eventuals::Then;
+using testing::ElementsAre;
+
+TEST(ComposedLoop, TwoLevelLoop) {
+  auto s = [&]() {
```
benh, line 24: You shouldn't need the reference capture default `&` here or below.

benh (commented):
I'll give this another look after you've made changes based on the comments!

stout/composed-loop.h
rjhuijsman: `StreamForEach` looks good to me!

PR #72, opened 2021-09-30
Introduce 'Let'

benh:
'Let()' provides syntactic sugar for using a 'Closure()' when you
effectively want to introduce a binding that will persist within the
enclosing scope.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -37,6 +37,7 @@ cc_library(
         "stout/filter.h",
         "stout/take.h",
         "stout/range.h",
+        "stout/let.h",
     ],
     srcs = [
         "stout/scheduler.cc",
--- a/stout/event-loop.cc
+++ b/stout/event-loop.cc
@@ -146,15 +146,31 @@ EventLoop::EventLoop()
   : clock_(*this) {
   uv_loop_init(&loop_);
 
-  uv_prepare_init(&loop_, &prepare_);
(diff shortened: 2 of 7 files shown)


onelxj:
LGTM but need to fix checks

Actually, I have run tests with 1000 iterations on Ubuntu and it passed successfully

benh:
Rather than tie `Let()` to `Then()` we can just leave it as a lambda and use `Closure()` so it can be used anywhere that a lambda would have been expected. So for example with the proposed `ComposedLoop()` you could do:

```
Iterate(...)
    | ComposedLoop(Let([](auto& x) {
      return Iterate(...)
          | ComposedLoop(Let([&x](auto& y) {
            return Range(x, y);
          });
    })
```

rjhuijsman:
I have two requests!
1. Could you please add PR descriptions to your PRs, for poor curious sods like yours truly who don't have context? 🙂 
2. Before you check this in could you add some comments in `let.h` that explain how it's different from `Then`, when to use it, and perhaps an example? I understand that `stout-eventuals` is still very much in flux, but leaving documentation to "some future date" is (1) setting a bad precedent for our future documentation efforts and (2) those docs would _already_ be helpful for poor curious sods like yours truly. 😃

onelxj:
The description of `Let` is awesome, thanks!

onelxj (approved):
LGTM

PR #74, opened 2021-09-30
Added code format check of `.proto` files

ArthurBandaryk:
Added `check_style_proto` in the `tools` directory. It's a simple bash script
which is run in the check_format_code.yaml action.

### DIFF
--- a/.github/workflows/check_format_code.yml
+++ b/.github/workflows/check_format_code.yml
@@ -9,6 +9,10 @@ jobs:
     - uses: actions/checkout@v2
     - name: Run bootstrap
       run: cd tools/ && ./bootstrap
+    - name: Install prototool
+      run: brew install prototool  
+    - name: Check all .proto files on correct code format
+      run: chmod +x tools/check_style_proto && tools/./check_style_proto  
     - name: Run clang-format style check for C/C++ programs.
       uses: jidicula/clang-format-action@v4.0.0
       with:
--- a/tools/check_style_proto
+++ b/tools/check_style_proto
@@ -1,0 +1,84 @@
+#!/bin/bash
+
+# Check for existence of prototool
+which prototool >/dev/null
+if [[ $? != 0 ]]; then
+  printf "Failed to find 'prototool'\n"
+  exit 1
+fi
+
+# Find all .proto files in the root directory
+current_directory=$(pwd)
+printf "Finding all .proto files in the %s\n" "${current_directory} directory..."
+IFS=:
+proto_file_paths=($(find . -type f -name "*.proto"))
+unset IFS
+
+# Check every .proto file on correct code style
+check(){
+  if [[ ${#proto_file_paths[@]} == 0 ]]; then
+    printf "There are no .proto files to check!\n"
+    exit 0
+  fi  
+
+  status_exit=0
+
+  for file in ${proto_file_paths}
+  do
+    # Check if the specific file has correct code format
+    # !!! RJ, GORM, BEN 
+    # IMPORTANT TO READ!!!
+    # In the documentation they tell us that in order to
+    # write a lint error in the form file:line:column:message
+    # if a file is unformatted we should use the next command:
+    # prototool format [file] -l (or --lint)
+    # but in fact it only prints that Format returned a diff.
+    # Actually if the file is well formatted the command above will
+    # return 0. So I used this fact in the code below!
+    # The link to the official doc:
+    # https://github.com/uber/prototool#prototool-format
+    # Read the Command Overview section!
+
+    printf "Checking ${file:2} ...\n"
+    prototool format -l ${file:2}
+    format_status=$(echo $?)
+
+    if [[ ${format_status} != 0 ]]; then
+      tput bold # Bold text in terminal
+      tput setaf 1 # Red font in terminal
+      printf "${file:2} file has incorrect code format!\n"
+      tput sgr0
+      printf "You can format proto files with prototool.\n"
+      tput bold # Bold text in terminal
+      printf "Command to format .proto files: "
+      tput setaf 2 # Green font in terminal
+      printf "prototool format -w file_name\n"
+      tput sgr0 # Reset terminal.
+      exit 1
+    fi
+
+    # Check if the specific line has more than 80 characters
+    local -i number=1
+    IFS='\n'
+    while read line; do
+      local -i length=${#line}
+      if (( length > 80 )); then
+        status_exit=1
+        tput bold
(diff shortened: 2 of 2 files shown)


ArthurBandaryk:
Also tested `check_style_proto` script locally. Was playing with correct format, then added some format errors and made some lines exceed 80 characters. The results and `test.proto` file are on the screen below.
Screen Shot 2021-09-30 at 9 40 15 PM
Screen Shot 2021-09-30 at 9 41 04 PM
Screen Shot 2021-09-30 at 9 41 37 PM

.github/workflows/check_format_code.yml
```diff
@@ -9,6 +9,10 @@ jobs:
     - uses: actions/checkout@v2
     - name: Run bootstrap
       run: cd tools/ && ./bootstrap
+    - name: Install prototool
+      run: brew install prototool
```
rjhuijsman, line 13: Huh, I didn't know Homebrew exists on Linux too! And it's preinstalled on GitHub's Ubuntu? Cool.

.github/workflows/check_format_code.yml
```diff
@@ -9,6 +9,10 @@ jobs:
     - uses: actions/checkout@v2
     - name: Run bootstrap
       run: cd tools/ && ./bootstrap
+    - name: Install prototool
+      run: brew install prototool  
+    - name: Check all .proto files on correct code format
```
rjhuijsman, line 14: English suggestion:
```suggestion
    - name: Check all .proto files for code formatting
```

tools/check_style_proto
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
```
rjhuijsman, line 1: Let's name this file `check_style_proto.sh` to indicate it's a shell script.

tools/check_style_proto
```diff
@@ -0,0 +1,84 @@
...
+
+# Check for existence of prototool
```
rjhuijsman, line 3: All comments are full sentences, so should end with a period (`.`).

tools/check_style_proto
```diff
@@ -0,0 +1,84 @@
...
+    if [[ ${format_status} != 0 ]]; then
+      tput bold # Bold text in terminal
+      tput setaf 1 # Red font in terminal
+      printf "${file:2} file has incorrect code format!\n"
+      tput sgr0
+      printf "You can format proto files with prototool.\n"
+      tput bold # Bold text in terminal
+      printf "Command to format .proto files: "
+      tput setaf 2 # Green font in terminal
+      printf "prototool format -w file_name\n"
```
rjhuijsman, line 55: Would be helpful to fill in the real filename here.

tools/check_style_proto
```diff
@@ -0,0 +1,84 @@
...
+    # In the documentation they tell us that in order to
+    # write a lint error in the form file:line:column:message
+    # if a file is unformatted we should use the next command:
+    # prototool format [file] -l (or --lint)
+    # but in fact it only prints that Format returned a diff.
+    # Actually if the file is well formatted the command above will
+    # return 0. So I used this fact in the code below!
+    # The link to the official doc:
+    # https://github.com/uber/prototool#prototool-format
+    # Read the Command Overview section!
```
rjhuijsman, line 40: Noted. Just leave your explanation here and it'll be great:

```suggestion
    # Check if the specific file has correct code format
    #
    # In the documentation they tell us that in order to
    # write a lint error in the form file:line:column:message
    # if a file is unformatted we should use the following command:
    #   prototool format [file] -l (or --lint)
    # ... but in fact that only prints an uninformative message.
    #
    # If the file is well formatted the command above will return 0. 
    # We use this fact to print a more helpful message.
```

tools/check_style_proto
```diff
@@ -0,0 +1,84 @@
...
+      printf "You can format proto files with prototool.\n"
+      tput bold # Bold text in terminal
+      printf "Command to format .proto files: "
+      tput setaf 2 # Green font in terminal
+      printf "prototool format -w file_name\n"
+      tput sgr0 # Reset terminal.
+      exit 1
+    fi
+
+    # Check if the specific line has more than 80 characters
```
rjhuijsman, line 60:
```suggestion
    # Check if any line has more than 80 characters.
```

tools/check_style_proto
```diff
@@ -0,0 +1,84 @@
...
+        printf "error: "
+        tput sgr0 # Reset terminal.
+        tput bold
+        printf "line exceeds 80 characters\n"
+        tput sgr0 # Reset terminal.
+        printf "%s\n" "${line}"
+      fi
+      (( number++ ))
+    done < ${file}
+    unset IFS
```
rjhuijsman, line 79: Shall we move all of this logic into a separate check that we apply to _both_ `.proto` files _and_ `.cc`/`.h` files?

(That would be a separate PR, we'd just omit it here for now).

rjhuijsman (changes requested):

tools/check_style_proto
while-false: +1

tools/check_style_proto
while-false: For `clang-format`, that would imply something like
`printf "clang-format --style=file -i \"%s\"\n" "${file}"`

tools/check_style_proto
```diff
@@ -0,0 +1,84 @@
...
+
+    printf "Checking ${file:2} ...\n"
+    prototool format -l ${file:2}
+    format_status=$(echo $?)
+
+    if [[ ${format_status} != 0 ]]; then
+      tput bold # Bold text in terminal
+      tput setaf 1 # Red font in terminal
+      printf "${file:2} file has incorrect code format!\n"
+      tput sgr0
```
while-false, line 50: I suggest `${file:2} -> ${file}` - the `./` is not that bad. ;D

tools/check_style_proto
while-false: I don't understand your parenthesis?

while-false (changes requested):

rjhuijsman:
To avoid any confusion:
* This PR is the sibling of `stout-grpc`'s https://github.com/3rdparty/stout-grpc/pull/24
* I reviewed this PR before I realized that Gorm had done a review on the sibling PR, where he made some really good points.
* Gorm's review on the other PR essentially makes my review on this PR obsolete - we should ignore what I said above and work on the sibling PR, then re-make this PR to match the sibling.

To avoid further confusion, I'm going to close this PR for now. @ArthurBandaryk please feel free to send another PR once the PR in `stout-grpc` has been checked in.

PR #80, opened 2021-10-02
Apply buildifier to bazel rules and update curl rules

FolMing:


BUILD.bazel
```diff
@@ -72,6 +72,10 @@ cc_library(
     visibility = ["//visibility:private"],
     deps = [
         ":eventuals-base",
+        ":eventuals-events",
+        # NOTE: Links against BoringSSL for HTTPS stuff.
+        # Compiles without any SSL support on Windows.
+        # See bazel-rules-curl for more details.
         "@com_github_curl_curl//:libcurl",
```
benh, line 79: This sounds good but what's the plan for actually https versions of the functions? Are you planning to use preprocessor ifdefs?

benh (approved):

BUILD.bazel
FolMing: The API for http and https functions is the same, the difference is the URL that user wants to interact with. I'm planning to use ifdefs for tests.

PR #86, opened 2021-10-06
Add 'StreamForEach' interface

onelxj:
`StreamForEach` is an instance that provides to create streams with inner loops

Let's imagine that a single common stream represents a one-level loop like

`Range(n)`  equals to `for (int i =0; i < n; ++i){...}`

When there is a composition of that Range with `StreamForEach([](int x) { return Range(k); })` it equals to 

```
for (int i = 0; i < n; ++i)
{
    for (int j = 0; k < 2; ++k){...}
}
```

In other words - it adds the inner loop inside the loop above each time you compose it with `StreamForEach`

The idea - pass a `Lamda` to the `StreamForEach` that produces a `Stream` instance (for now it instance that inherits from `TypeErasedStream` structure) and this `Stream` will be the inner loop for the stream above.

### DIFF
--- a/test/stream-for-each.cc
+++ b/test/stream-for-each.cc
@@ -1,0 +1,270 @@
+#include "stout/stream-for-each.h"
+
+#include <atomic>
+#include <chrono>
+#include <iostream>
+#include <string>
+#include <thread>
+#include <vector>
+
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+#include "stout/collect.h"
+#include "stout/event-loop.h"
+#include "stout/iterate.h"
+#include "stout/let.h"
+#include "stout/map.h"
+#include "stout/range.h"
+#include "stout/stream.h"
+#include "stout/terminal.h"
+#include "stout/then.h"
+#include "stout/timer.h"
+#include "test/event-loop-test.h"
+
+using stout::eventuals::Collect;
+using stout::eventuals::EventLoop;
+using stout::eventuals::Interrupt;
+using stout::eventuals::Iterate;
+using stout::eventuals::Let;
+using stout::eventuals::Map;
+using stout::eventuals::Range;
+using stout::eventuals::Stream;
+using stout::eventuals::StreamForEach;
+using stout::eventuals::Then;
+using stout::eventuals::Timer;
+using testing::ElementsAre;
+
+TEST(StreamForEach, TwoLevelLoop) {
+  auto s = []() {
+    return Range(2)
+        | StreamForEach([](int x) { return Range(2); })
+        | Collect<std::vector<int>>();
+  };
+
+  EXPECT_THAT(*s(), ElementsAre(0, 1, 0, 1));
+}
+
+TEST(StreamForEach, StreamForEachMapped) {
+  auto s = []() {
+    return Range(2)
+        | StreamForEach([](int x) { return Range(2); })
+        | Map(Then([](int x) { return x + 1; }))
+        | Collect<std::vector<int>>();
+  };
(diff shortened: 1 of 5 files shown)


rjhuijsman (approved):
This looks great! Just some requests around test organization to make this look super-professional. 🙂

test/stream-for-each.cc
```diff
@@ -0,0 +1,271 @@
...
+             c.push_back(std::vector<int>());
+             c.push_back(std::vector<int>());
+             return Iterate(std::move(c));
+           })
+        | StreamForEach([](std::vector<int> x) { return Range(2); })
+        | Collect<std::vector<int>>();
+  };
+
+  EXPECT_THAT(*s(), ElementsAre(0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1));
+}
```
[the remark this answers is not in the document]
onelxj: yep, template structures, that are constructed inside. You will see the comment in the next patch:D

test/stream-for-each.cc
```diff
@@ -0,0 +1,271 @@
...
+  ASSERT_FALSE(interrupted.load());
+
+  interrupt.Trigger();
+
+  EXPECT_THROW(future.get(), stout::eventuals::StoppedException);
+
+  ASSERT_TRUE(interrupted.load());
+}
+
+// Iterate with init_list
```
benh, line 110: Is this a TODO that you wanted to change this test to use a `std::initializer_list` instead of a `std::vector`?

test/stream-for-each.cc
```diff
@@ -0,0 +1,271 @@
...
+        | Collect<std::vector<int>>();
+  };
+
+  EXPECT_THAT(*s(), ElementsAre(1, 2, 2, 3, 3, 4));
+}
+
+TEST(StreamForEach, TwoIndexesSumMap) {
+  auto s = []() {
+    return Range(3)
+        | StreamForEach([](int x) { return Range(1, 3)
```
benh, line 163: Let's always try and wrap functions like this on a newline, it's easier to read the return and continuation that way. Here and below, thanks!

test/stream-for-each.cc
```diff
@@ -0,0 +1,271 @@
...
+        | Collect<std::vector<int>>();
+  };
+
+  EXPECT_THAT(*s(), ElementsAre(0, 1, 0, 1, 0, 1, 0, 1));
+}
+
+TEST(StreamForEach, ThreeLevelLoop2) {
+  auto s = []() {
+    return Range(2)
+        | StreamForEach([](int x) { return Range(2) | StreamForEach([](int y) { return Range(2); }); })
```
benh, line 198: How did you get this past the column length checker? Are you using the pre-commit hook?

Again, let's wrap this at the first `return` and then at subsequent `|` so it's easier to read, thanks!

benh (approved):
Woo hoo! Please address the feedback but excited to see this land!

test/stream-for-each.cc
```diff
@@ -0,0 +1,271 @@
...
+using stout::eventuals::Then;
+using stout::eventuals::Timer;
+using testing::ElementsAre;
+
+class StreamForEachTest : public EventLoopTest {};
+
+TEST_F(StreamForEachTest, Interrupt) {
+  auto e = []() {
+    return Iterate(std::vector<int>(1000))
+        | Map(Then([](int x) { return Timer(std::chrono::milliseconds(100)) | Then([x]() { return x; }); }))
```
FolMing, line 42: Why do you need a timer here?

PR #87, opened 2021-10-06
Add support of std::initializer_list for Iterate

onelxj:
Allows use Iterate interface with `std::initializer_list` as argument:

```
...
return Iterate({1,2,3})
| Collect<std::vector<int>>();
...
```

### DIFF
--- a/test/iterate.cc
+++ b/test/iterate.cc
@@ -2,6 +2,7 @@
 
 #include <array>
 #include <deque>
+#include <initializer_list>
 #include <list>
 #include <map>
 #include <set>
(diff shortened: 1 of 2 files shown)


test/iterate.cc
```diff
@@ -754,3 +755,20 @@ TEST(Iterate, VectorStringContcatenatePartial) {
 
   EXPECT_EQ("Hello World", *s());
 }
+
+TEST(Iterate, InitializerList) {
+  auto s = [&]() {
```
benh, line 760: No need for default capture!

PR #94, opened 2021-10-07
Added `copts_` list + move `-std=c++17` in `BUILD`

ArthurBandaryk:
`copts_` list contains `-Werror` option in order to turn all warnings into the
errors. `-Wno-error=deprecated-declarations` option turns `syscall` warning
into a warning even if -Werror is specified.
Also I created an issue in the google/glog repo about `syscall` warning
[#717](https://github.com/google/glog/issues/717).

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,4 +1,5 @@
 load("@rules_cc//cc:defs.bzl", "cc_library")
+load("//bazel:copts.bzl", "copts")
 
 cc_library(
     name = "eventuals-base",
@@ -12,6 +13,7 @@ cc_library(
         "stout/closure.h",
         "stout/collect.h",
         "stout/compose.h",
+        "stout/concurrent.h",
         "stout/conditional.h",
         "stout/context.h",
         "stout/eventual.h",
@@ -44,8 +46,8 @@ cc_library(
         "stout/type-traits.h",
         "stout/undefined.h",
         "stout/until.h",
-        "stout/concurrent.h",
     ],
+    copts = copts(),
     visibility = ["//visibility:private"],
     deps = [
         "@com_github_google_glog//:glog",
@@ -64,6 +66,7 @@ cc_library(
         "stout/signal.h",
         "stout/timer.h",
     ],
+    copts = copts(),
     visibility = ["//visibility:private"],
     deps = [
         ":eventuals-base",
@@ -79,6 +82,7 @@ cc_library(
     hdrs = [
         "stout/http.h",
     ],
+    copts = copts(),
     defines = [
         # Windows build fails without this define.
         "GLOG_NO_ABBREVIATED_SEVERITIES",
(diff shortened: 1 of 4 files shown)


while-false:
> All our other compiler flags seem to live in `.bazelrc` - shall we put these there too?

No. Having these flags "inside" bazel allow us to do conditional flags depending on platform and compiler. The flags in questions are common, but I think this rightfully belongs where Arthur put it.

BUILD.bazel
```diff
@@ -1,5 +1,21 @@
 load("@rules_cc//cc:defs.bzl", "cc_library")
 
+copts_ = select({
```
while-false, line 3: I like the idea of having the `copts_` and the select here. I'd like to ask you to do something like (in pseudo code):

```
copts_common_ = ["-Werror"]
copts_ = copts_common_ + select(...)
```

This would allow us to very clearly see what is common flags and what are platform specific flags.

If also paves the way for another PR where you can add compiler specific flags (again in pseudo code):

```
copts_ = copts_common_ + select(
 # os specific opts
) + select(
 # compiler specific opts
)
```

ArthurBandaryk: Ok:) I like your idea about adding copts_common options. On my way to apply it!

bazel/copts.bzl
```diff
@@ -0,0 +1,19 @@
+copts_common = [
+    "-Werror",
+]
+
+copts_ = copts_common + select({
```
benh, line 5: Thinking through this one final time I have one last request which is that we actually make this a function, i.e., `def copts():`.

Here's the reasoning: I can imagine following this pattern other places too, i.e., when we have shared configuration, however, I can imagine that in some circumstances we'll actually want to "compute" the configuration versus just declare it. While these are almost one in the same with Bazel, the added flexibility of being able to just come and update a function seems worth starting that pattern this way from the start rather than bolting it on later, even if most of our shared configurations will be declarative like this one. Make sense?

benh (changes requested):

bazel/copts.bzl
ArthurBandaryk: Yep, great idea, as usual:)

bazel/copts.bzl
```diff
@@ -0,0 +1,24 @@
...
+        "@bazel_tools//src/conditions:darwin": [
+            "-std=c++17",
+            "-Wno-error=deprecated-declarations",
+        ],
+        "//conditions:default": [
+            "-std=c++17",
+        ],
+    })
+
+    return copts_
```
benh, line 24:
```suggestion
    return copts_common + select({
        "@bazel_tools//src/conditions:windows": [
            "\"/std:c++17\"",
            "-Wno-error=microsoft-cast",
            "-Wno-error=invalid-noreturn",
            "-Wno-error=microsoft-include",
        ],
        "@bazel_tools//src/conditions:darwin": [
            "-std=c++17",
            "-Wno-error=deprecated-declarations",
        ],
        "//conditions:default": [
            "-std=c++17",
        ],
    })
```

PR #95, opened 2021-10-07
Pin threads to specific cores

FolMing:
This PR makes so that each thread in static thread pool is mapped to a specific core which is determined by the position of a particular thread inside vector.

### DIFF
--- a/stout/os.h
+++ b/stout/os.h
@@ -1,0 +1,67 @@
+#pragma once
+
+#include <thread>
+
+#include "glog/logging.h" // NOTE: must be included before <windows.h>.
+
+#ifdef __MACH__
+#include <limits>
+#elif _WIN32
+#include <windows.h>
+#else
+#include <pthread.h>
+#endif // __MACH__
+
+////////////////////////////////////////////////////////////////////////
+
+namespace stout {
+
+////////////////////////////////////////////////////////////////////////
+
+#ifdef __MACH__
+inline size_t GetRunningCPU() {
+  // NOTE: Returning incorrect value here because
+  // we don't currently know a way to correctly recognize
+  // on which core the current thread is running.
+  return std::numeric_limits<size_t>::max();
+}
+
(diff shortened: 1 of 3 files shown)


stout/static-thread-pool.cc
```diff
@@ -69,6 +73,11 @@ StaticThreadPool::StaticThreadPool()
 
               STOUT_EVENTUALS_LOG(1) << "Resuming '" << waiter->name() << "'";
 
+              STOUT_EVENTUALS_LOG(1) << "Thread " << core << "(id="
+                                     << std::this_thread::get_id()
+                                     << ") is running on core "
+                                     << GetRunningCore();
```
benh, line 79: Let's move this up to just after we set the affinity. No need to print it on the log every time! Or if you're thinking that this will be useful for debugging if affinity isn't working then let's increase the log level to 2 or even 3 (not 1). Also see the code suggestion where I've added a space before the `(` so that the core number is easier to read.

```suggestion
              STOUT_EVENTUALS_LOG(1) << "Thread " << core << " (id="
                                     << std::this_thread::get_id()
                                     << ") is running on core "
                                     << GetRunningCore();
```

stout/os.h
```diff
@@ -0,0 +1,67 @@
...
+#endif // __MACH__
+
+////////////////////////////////////////////////////////////////////////
+
+namespace stout {
+
+////////////////////////////////////////////////////////////////////////
+
+#ifdef __MACH__
+inline size_t GetRunningCore() {
```
benh, line 22: Minor nit but we use both the term "core" and the term "cpu" here, let's just stick to one so that folks understand in our API that we're referring to the same thing! For example:

```c++
inline size_t GetRunningCpu();
```

benh (changes requested):
Please address the comments otherwise LGTM!

benh (approved):

PR #98, opened 2021-10-12
Introduce 'Concurrent()' combinator

benh:
'Concurrent()' allows you to handle streams of values concurrently
(i.e., simultaneously). See comments in the header for a deeper
explanation of it's semantics.


stout/concurrent.h
```diff
@@ -0,0 +1,748 @@
...
+// it can compose with an "upstream" stream and itself is a stream
+// (for example, it should be a 'Map()' or 'StreamForEach()').
+//
+// If one of the eventuals raises a failure or stops, then we attempt
+// to call "done" on the upstream stream, wait for all of the
+// eventuals to finish, and then propagate the failure or stop
+// downstream. There is one caveat to this case which is that we can't
+// attempt to tell the upstream we're done until it has called down
+// into us which means if we've called 'Next()' and it hasn't returned
+// then we have could wait forever. For now, the way to rectify this
```
while-false, line 44: "then we have could wait forever" -> "then we could wait forever" ?

test/concurrent.cc
```diff
@@ -0,0 +1,717 @@
...
+        | Concurrent([]() {
+             return StreamForEach([](int i) {
+               return Range(i);
+             });
+           })
+        | Collect<std::vector<int>>();
+  };
+
+  EXPECT_THAT(*e(), UnorderedElementsAre(0, 0, 1));
+}
```
while-false, line 717: There is not a single comment in here, which for my untrained eye makes most of these tests rather hard to read or reason about. You have a ton of nice comments in the headers. Could I ask you to add a line or two per test?

test/concurrent.cc
```diff
@@ -0,0 +1,717 @@
...
+
+  auto [future, k] = Terminate(e());
+
+  k.Start();
+
+  ASSERT_EQ(2, callbacks.size());
+
+  EXPECT_EQ(
+      std::future_status::timeout,
+      future.wait_for(std::chrono::seconds(0)));
```
FolMing, line 68: I really like how you check for the current eventual status here and in `ASSERT_EQ(2, callbacks.size());`! That's very smart. I would appreciate the comments though!

test/concurrent.cc
```diff
@@ -0,0 +1,717 @@
...
+
+  for (auto& callback : callbacks) {
+    callback();
+  }
+
+  EXPECT_ANY_THROW(future.get());
+}
+
+TEST(ConcurrentTest, InterruptSuccess) {
```
FolMing, line 210: I am not entirely sure what this test tests. Why there's an interrupt and why the whole eventual succeeds even after triggering the interrupt? Please add a comment!

PR #99, opened 2021-10-13
Add buildifier checks in pre-commit

NikitaNikolaenko:
Checks if there are any warnings or files that are not reformatted properly. If any of this happens pre-commit hook will forbid the commit.


benh (approved):
LGTM!

Seems like the extra time we put in to abstracting things like `check_version` paid off! I'm so glad!

PR #101, opened 2021-10-14
Refactor 'Task' and add new functionality

onelxj:
Rewrite Task eventual with `Composable` and `Continuation`.

Add support of using `Task<void>...` that returns nothing.

Add propagation for `Fail` and `Stop` callbacks

### DIFF
--- a/test/task.cc
+++ b/test/task.cc
@@ -1,13 +1,20 @@
 #include "eventuals/task.h"
 
+#include "eventuals/eventual.h"
 #include "eventuals/just.h"
+#include "eventuals/map.h"
+#include "eventuals/terminal.h"
 #include "eventuals/then.h"
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
+using eventuals::Eventual;
 using eventuals::Just;
+using eventuals::Map;
 using eventuals::Task;
 using eventuals::Then;
+
+using testing::MockFunction;
 
 TEST(TaskTest, Succeed) {
   auto e1 = []() -> Task<int> {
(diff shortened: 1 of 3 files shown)


test/task.cc
```diff
@@ -1,36 +1,47 @@
 #include "stout/task.h"
 
+#include <iostream>
```
benh, line 3: Unused?

test/task.cc
```diff
@@ -1,36 +1,47 @@
...
+using stout::eventuals::Map;
 using stout::eventuals::Task;
+using stout::eventuals::Terminate;
 using stout::eventuals::Then;
+using testing::MockFunction;
 
 TEST(TaskTest, Succeed) {
   auto e1 = []() -> Task<int> {
-    return [x = 42]() {
+    return Task<int>([x = 42]() {
```
benh, line 25: Please revert, the old test behavior should still work!

test/task.cc
```diff
@@ -1,36 +1,47 @@
...
-           })
-        | e1();
+  auto e2 = [&]() -> Task<int> {
+    return Task<int>([&]() {
+      return e1()
+          | Then([](int i) {
+               return i + 1;
+             })
+          | e1();
+    });
```
benh, line 39: Please revert, the old test behavior should still work!

test/task.cc
```diff
@@ -1,36 +1,47 @@
...
   };
 
   EXPECT_EQ(42, *e2());
 
-  auto e3 = []() {
+  auto e3 = []() -> Task<int>::With<int, std::string> {
```
benh, line 44: Please revert, the old test behavior should still work!

test/task.cc
```diff
@@ -41,13 +52,149 @@ TEST(TaskTest, Succeed) {
...
-           })
-        | e3();
+  auto e4 = [&]() -> Task<int> {
+    return Task<int>([&]() {
+      return e3()
+          | Then([](int i) {
+               return i + 1;
+             })
+          | e3();
+    });
```
benh, line 62: Please revert, the old test behavior should still work!

test/task.cc
```diff
@@ -41,13 +52,149 @@ TEST(TaskTest, Succeed) {
...
 }
+
+TEST(Task, Void) {
+  int x = 0;
+  auto e1 = [&x]() -> Task<void> {
+    return Task<void>([&x]() {
+      return Then([&x]() {
+        x = 100;
+      });
+    });
```
benh, line 75:
```suggestion
    return [&x]() {
      return Then([&x]() {
        x = 100;
      });
    };
```

test/task.cc
```diff
@@ -41,13 +52,149 @@ TEST(TaskTest, Succeed) {
...
+      .Times(1);
+
+  EXPECT_CALL(functions.stop, Call())
+      .Times(0);
+
+  EXPECT_CALL(functions.start, Call())
+      .Times(0);
+
+  auto e = [&]() -> Task<int> {
+    return Task<int>([&]() {
```
benh, line 100:
```suggestion
    return [&]() {
```

test/task.cc
```diff
@@ -41,13 +52,149 @@ TEST(TaskTest, Succeed) {
...
+      return Eventual<int>()
+                 .context("error")
```
benh, line 102: Seems like you don't actually need the `.context("error")` here, can youjust do `k.Fail("error");` below please?

test/task.cc
```diff
@@ -41,13 +52,149 @@ TEST(TaskTest, Succeed) {
...
+}
+
+TEST(Task, Fail) {
+  MockFunction<void()> fail;
+
+  EXPECT_CALL(fail, Call())
+      .Times(1);
+
+  auto e = [&]() -> Task<int> {
+    return Task<int>([&]() {
```
benh, line 135:
```suggestion
    return [&]() {
```

test/task.cc
```diff
@@ -41,13 +52,149 @@ TEST(TaskTest, Succeed) {
...
+      return Eventual<int>()
+                 .context("error")
```
benh, line 137: Same comment about `context("error")` here.

test/task.cc
```diff
@@ -41,13 +52,149 @@ TEST(TaskTest, Succeed) {
...
+}
+
+TEST(Task, StopOnCallback) {
+  MockFunction<void()> stop;
+
+  EXPECT_CALL(stop, Call())
+      .Times(0);
+
+  auto e = [&]() {
+    return Task<int>([&]() {
```
benh, line 162:
```suggestion
    return [&]() {
```

stout/task.h
```diff
@@ -72,223 +84,298 @@ struct HeapTask {
...
+            break;
+          case Action::Stop:
+            e->Stop(
+                interrupt,
+                std::move(start),
+                std::move(fail),
+                std::move(stop));
+            break;
+          default:
+            break;
```
benh, line 326:
```suggestion
            LOG(FATAL) << "unreachable";
```

stout/task.h
```diff
@@ -72,223 +84,298 @@ struct HeapTask {
...
-
-      auto* e = static_cast<detail::HeapTask<E>*>(e_.get());
-
-      e->Start(interrupt, std::move(start), std::move(fail), std::move(stop));
-    };
-  }
+    template <typename... Args>
+    void Fail(Args&&... args) {
+      auto exception = std::make_exception_ptr(
+          std::forward<decltype(args)>(args)...);
```
benh, line 177: I'm realizing that we don't handle the case when `args` are empty here or in `terminal.h`. Can you please handle that case here by just using a `std::runtime_error` and then also do the same in `terminal.h`? Thanks!

stout/task.h
```diff
@@ -72,223 +84,298 @@ struct HeapTask {
...
+  Adaptor_ adaptor_;
 };
 
 ////////////////////////////////////////////////////////////////////////
 
-} // namespace detail
-
-////////////////////////////////////////////////////////////////////////
+struct _TaskWith {
+  enum class Action {
```
benh, line 143: Let's add a small comment about why we have this enum and how it works with "dispatch". Thanks!

PR #103, opened 2021-10-14
Add IsLockedBy function to Lock

FolMing:


stout/lock.h
```diff
@@ -108,6 +109,19 @@ class Lock {
     return head_.load(std::memory_order_relaxed) == nullptr;
   }
 
+  bool IsLockedBy(Scheduler::Context* context) {
+    Waiter* current_waiter = head_.load(std::memory_order_acquire);
+    if (current_waiter != nullptr) {
+      if (current_waiter->context == context) {
```
onelxj, line 115: Can `current_waiter ` be changed at this moment? Need to double-check

stout/lock.h
```diff
@@ -108,6 +109,19 @@ class Lock {
     return head_.load(std::memory_order_relaxed) == nullptr;
   }
 
+  bool IsLockedBy(Scheduler::Context* context) {
+    Waiter* current_waiter = head_.load(std::memory_order_acquire);
+    if (current_waiter != nullptr) {
```
onelxj, line 114: Can u rewrite this like `if (current_waiter == nullptr){return false;} ...`
That will improve the readability.

PR #104, opened 2021-10-17
Add 'Lock::OwnedByCurrentSchedulerContext()' for checking usage of locks

benh:
Currently there is no way to determine which scheduler context owns
the lock. When writing functions that are not obviously synchronized
due to the lexical scope it would be nice to add a 'CHECK()' that both
serves as documentation as well as to ensure that the locking contract
has not been broken. This commit enables one to do the following at
the beginning of each such function:

`CHECK(lock.OwnedByCurrentSchedulerContext());`


FolMing (approved):
LGTM!

PR #107, opened 2021-10-18
Add missing interrupt registration for streams

onelxj:


stout/filter.h
```diff
@@ -44,10 +44,18 @@ struct _Filter {
       k_.Ended();
     }
 
+    void Register(Interrupt& interrupt) {
+      assert(interrupt_ == nullptr);
```
benh, line 48: We're trying to move to `CHECK()` instead of `assert()`, please update, thanks!

stout/filter.h
```diff
@@ -44,10 +44,18 @@ struct _Filter {
...
+      interrupt_ = &interrupt;
```
benh, line 49: It doesn't look like you're ever using `interrupt_`. Did you mean to? If not, all you need here is `k_.Register(interrupt);`.

stout/range.h
```diff
@@ -34,14 +34,16 @@ struct _Range {
...
-      k_.Register(interrupt);
-    }
-
     void Ended() {
       k_.Ended();
     }
 
+    void Register(Interrupt& interrupt) {
+      assert(interrupt_ == nullptr);
+      interrupt_ = &interrupt;
```
benh, line 43: Same question as above.

PR #110, opened 2021-10-19
Rename from stout-eventuals to just eventuals

benh:
Looking for a quick review from any of @FolMing, @onelxj, @ArthurBandaryk, or @while-false. This is just a mechanical rename, no semantics changed!


onelxj (changes requested):
Also, need to rename at http.h

PR #128, opened 2021-10-29
Fix bug with 'CHECK(!InEventLoop())' that was too strict

benh:
We use 'CHECK(!InEventLoop())' to make sure that we're only
destructing objects that are using the event loop _outside_ of the
event loop which is fine except when those eventuals were never
actually started so it doesn't matter where they are destructed. This
case presents itself after we move a continuation _before_ it's
started and then destruct the continuation that is no longer valid
we'll still call the destructor which would fire that 'CHECK()'. In
this case, the continuation is "dead" which we know because it hasn't
been started.


onelxj (approved):
That looks great! Well done!
Check please my suggestion on shortwave and ship it!

PR #130, opened 2021-11-01
Refactor `DoAll` for more ergonomic use

ArthurBandaryk:
For now `Succeed` test is ready for your review. Soon I will prepare `Fail` and `Interrupt` tests too.

### DIFF
--- a/eventuals/do-all.h
+++ b/eventuals/do-all.h
@@ -5,21 +5,32 @@
 #include "eventuals/compose.h"
 #include "eventuals/terminal.h"
 
+////////////////////////////////////////////////////////////////////////
+
 namespace eventuals {
+
+////////////////////////////////////////////////////////////////////////
+
 namespace detail {
+
+////////////////////////////////////////////////////////////////////////
 
 struct _DoAll {
   template <typename K_, typename... Eventuals_>
   struct Adaptor {
-    Adaptor(K_& k)
-      : k_(k) {}
+    Adaptor(K_& k, Interrupt& interrupt)
+      : k_(k),
+        interrupt_(interrupt) {}
+
     Adaptor(Adaptor&& that)
-      : k_(that.k_) {
+      : k_(that.k_),
+        interrupt_(that.interrupt_) {
       CHECK(that.counter_.load() == sizeof...(Eventuals_))
           << "moving after starting is illegal";
     }
 
     K_& k_;
+    Interrupt& interrupt_;
 
     std::tuple<
         std::variant<
@@ -47,7 +58,20 @@ struct _DoAll {
                   }
                   if (counter_.fetch_sub(1) == 1) {
                     // You're the last eventual so call the continuation.
-                    k_.Start(std::move(values_));
+                    std::optional<std::exception_ptr> exception =
+                        GetExceptionIfExists();
+
+                    if (exception) {
+                      try {
+                        std::rethrow_exception(*exception);
+                      } catch (const StoppedException&) {
+                        k_.Stop();
+                      } catch (...) {
+                        k_.Fail(std::current_exception());
+                      }
+                    } else {
+                      k_.Start(GetTupleOfValues());
+                    }
                   }
                 })
                 .fail([this](auto&&... errors) {
@@ -57,7 +81,21 @@ struct _DoAll {
                               std::forward<decltype(errors)>(errors)...));
                   if (counter_.fetch_sub(1) == 1) {
                     // You're the last eventual so call the continuation.
-                    k_.Start(std::move(values_));
+                    std::optional<std::exception_ptr> exception =
+                        GetExceptionIfExists();
+
+                    CHECK(exception);
+                    try {
+                      std::rethrow_exception(*exception);
+                    } catch (const StoppedException&) {
+                      k_.Stop();
+                    } catch (...) {
+                      k_.Fail(std::current_exception());
+                    }
+                  } else {
+                    // Interrupt the remaining eventuals so we can
+                    // propagate the failure.
+                    interrupt_.Trigger();
                   }
                 })
                 .stop([this]() {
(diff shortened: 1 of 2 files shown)


eventuals/do-all.h
```diff
@@ -83,19 +172,26 @@ struct _DoAll {
 
   template <typename K_, typename... Eventuals_>
   struct Continuation {
+    Continuation(K_&& k, std::tuple<Eventuals_...>&& eventuals)
+      : k_(std::move(k)),
+        eventuals_(std::move(eventuals)) {}
+    Continuation(Continuation&& that)
```
benh, line 178:
```suggestion
        eventuals_(std::move(eventuals)) {}

    Continuation(Continuation&& that)
```

eventuals/do-all.h
```diff
@@ -67,11 +99,68 @@ struct _DoAll {
...
+            stopped = false;
+            exception.emplace(std::current_exception());
+          }
+        } else {
+          stopped = false;
+        }
+      };
+
+      std::apply(
+          [check_value](auto&... value) {
```
benh, line 151:
```suggestion
          [&check_value](auto&... value) {
```
Seems like you might as well take by reference?

eventuals/do-all.h
```diff
@@ -67,11 +99,68 @@ struct _DoAll {
...
 
+    std::tuple<
+        std::conditional_t<
+            std::is_void_v<typename Eventuals_::template ValueFrom<void>>,
+            std::monostate,
+            typename Eventuals_::template ValueFrom<void>>...>
+    GetTupleOfValues() {
+      return std::apply(
+          [](auto&... value) {
+            ((CHECK(value.valueless_by_exception() == false)), ...);
```
benh, line 125:
```suggestion
            (CHECK_EQ(0, value.index()), ...);
```
I love that you're checking this! But rather than checking for `valueless_by_exception()` how about we check for the index? That way we are sure the `std::get<0>` will work below and that not only is it not `valueless_by_exception()` but it also doesn't hold a `std::exception_ptr` at index 1.

eventuals/do-all.h
```diff
@@ -47,7 +50,21 @@ struct _DoAll {
...
+                        k_.Stop();
+                      } catch (...) {
+                        k_.Fail(std::current_exception());
+                      }
+                    } else {
+                      k_.Start(GetTupleOfValues());
+                    }
+                  } else {
+                    interrupt_.Trigger();
                   }
```
benh, line 68: We don't want to interrupt if an eventual was successful, just want to wait for the rest!
```suggestion
                  }
```

eventuals/do-all.h
```diff
@@ -57,7 +74,22 @@ struct _DoAll {
...
+                      try {
+                        std::rethrow_exception(*exception);
+                      } catch (const StoppedException&) {
+                        k_.Stop();
+                      } catch (...) {
+                        k_.Fail(std::current_exception());
+                      }
+                    } else {
+                      k_.Start(GetTupleOfValues());
+                    }
```
benh, line 90: At this point we should know that at least one exception exists because we just added one earlier in the lambda.
```suggestion
                    CHECK(exception);
                    try {
                      std::rethrow_exception(*exception);
                    } catch (const StoppedException&) {
                      k_.Stop();
                    } catch (...) {
                      k_.Fail(std::current_exception());
                    }
```

eventuals/do-all.h
```diff
@@ -67,11 +99,68 @@ struct _DoAll {
...
+
+                    if (exception) {
+                      try {
+                        std::rethrow_exception(*exception);
+                      } catch (const StoppedException&) {
+                        k_.Stop();
+                      } catch (...) {
+                        k_.Fail(std::current_exception());
+                      }
+                    }
```
benh, line 112:
```suggestion
                    CHECK(exception);
                    try {
                      std::rethrow_exception(*exception);
                    } catch (const StoppedException&) {
                      k_.Stop();
                    } catch (...) {
                      k_.Fail(std::current_exception());
                    }
```

test/do-all.cc
```diff
@@ -30,33 +30,21 @@ TEST(DoAllTest, Succeed) {
...
-  using T1 = decltype(std::get<0>(std::get<0>(result)));
-  using T2 = decltype(std::get<0>(std::get<1>(result)));
-  using T3 = decltype(std::get<0>(std::get<2>(result)));
+  using T1 = decltype(std::get<0>(result));
+  using T2 = decltype(std::get<1>(result));
+  using T3 = decltype(std::get<2>(result));
 
   EXPECT_EQ((std::is_same_v<T1, int&>), true);
   EXPECT_EQ((std::is_same_v<T2, std::string&>), true);
   EXPECT_EQ((std::is_same_v<T3, std::monostate&>), true);
```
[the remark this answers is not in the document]
benh: Better yet, let's turn these into `static_assert()`:
```suggestion
  static_assert(std::is_same_v<T0, int&>);
  static_assert(std::is_same_v<T1, std::string&>);
  static_assert(std::is_same_v<T2, std::monostate&>);
```

eventuals/do-all.h
ArthurBandaryk: > I love that you're checking this! But rather than checking for `valueless_by_exception()` how about we check for the index? That way we are sure the `std::get<0>` will work below and that not only is it not `valueless_by_exception()` but it also doesn't hold a `std::exception_ptr` at index 1.

Great idea, Ben, I like it. But the following suggestion - `((CHECK_EQ(0, value.index())), ...);` will not compile. On the screen below you can watch the error. So my suggestion will be to use `CHECK`. That's why the final synax would be: `((CHECK(value.index() == 0)), ...);`
Снимок экрана от 2021-11-05 16-11-16

eventuals/do-all.h
```diff
@@ -11,15 +11,18 @@ namespace detail {
 struct _DoAll {
   template <typename K_, typename... Eventuals_>
   struct Adaptor {
-    Adaptor(K_& k)
-      : k_(k) {}
+    Adaptor(K_& k, Interrupt& interrupt)
+      : k_(k),
+        interrupt_(interrupt) {}
     Adaptor(Adaptor&& that)
```
benh, line 17:
```suggestion
        interrupt_(interrupt) {}

    Adaptor(Adaptor&& that)
```

eventuals/do-all.h
```diff
@@ -83,19 +168,27 @@ struct _DoAll {
...
+      : k_(std::move(k)),
+        eventuals_(std::move(eventuals)) {}
+
+    Continuation(Continuation&& that)
+      : k_(std::move(that.k_)),
+        eventuals_(std::move(that.eventuals_)),
+        adaptor_(std::move(that.adaptor_)),
+        ks_(std::move(that.ks_)),
+        handler_(std::move(that.handler_)) {}
     template <typename... Args>
```
benh, line 181:
```suggestion
        handler_(std::move(that.handler_)) {}

    template <typename... Args>
```

eventuals/do-all.h
```diff
@@ -128,21 +226,18 @@ struct _DoAll {
...
-        std::tuple<
-            std::variant<
-                std::conditional_t<
-                    std::is_void_v<
-                        typename Eventuals_::template ValueFrom<void>>,
-                    std::monostate,
-                    typename Eventuals_::template ValueFrom<void>>,
-                std::exception_ptr>...>;
+
+    using ValueFrom = std::tuple<
```
benh, line 236:
```suggestion
    using ValueFrom = std::tuple<
```

eventuals/do-all.h
```diff
@@ -67,11 +95,68 @@ struct _DoAll {
...
 
+    std::tuple<
+        std::conditional_t<
+            std::is_void_v<typename Eventuals_::template ValueFrom<void>>,
+            std::monostate,
+            typename Eventuals_::template ValueFrom<void>>...>
+    GetTupleOfValues() {
+      return std::apply(
+          [](auto&... value) {
+            ((CHECK(value.index() == 0)), ...);
```
benh, line 121: What about using less parens: `(CHECK_EQ(0, value.index(), ...);`

If that doesn't work then: 
```suggestion
            // NOTE: not using `CHECK_EQ()` here because compiler ... please fill in Artur!
            ((CHECK(value.index() == 0)), ...);
```

eventuals/do-all.h
```diff
@@ -57,7 +73,19 @@ struct _DoAll {
...
+                    CHECK(exception);
+                    try {
+                      std::rethrow_exception(*exception);
+                    } catch (const StoppedException&) {
+                      k_.Stop();
+                    } catch (...) {
+                      k_.Fail(std::current_exception());
+                    }
+                  } else {
+                    interrupt_.Trigger();
```
benh, line 88:
```suggestion
                    // Interrupt the remaining eventuals so we can propagate the failure.
                    interrupt_.Trigger();
```

eventuals/do-all.h
```diff
@@ -67,11 +95,68 @@ struct _DoAll {
...
+
+                    CHECK(exception);
+                    try {
+                      std::rethrow_exception(*exception);
+                    } catch (const StoppedException&) {
+                      k_.Stop();
+                    } catch (...) {
+                      k_.Fail(std::current_exception());
+                    }
                   }
```
benh, line 109: Why aren't we interrupting everyone else like we are for the failure case above?

benh (changes requested):

benh (approved):
🕺🏼

PR #131, opened 2021-11-02
Introduce 'Finally()' and 'Foreach()'

benh:
The core of this PR is adding combinators 'Finally()' and 'Foreach()' as I need them for eventuals-grpc and @FolMing also needs 'Finally()' for his HTTP work.

Also included in this PR is a helpful 'CHECK()' that 'Context' is used correctly.


FolMing (approved):
Thank you! LGTM!

PR #135, opened 2021-11-04
Use Composable instead of GeneratorWith in Generator

onelxj:


eventuals/generator.h
```diff
@@ -319,9 +319,6 @@ struct _GeneratorWith {
 
   template <typename Value_, typename... Args_>
   struct Composable {
-    template <typename>
-    using ValueFrom = Value_;
```
benh, line 323: Rather than delete these I'd rather update the `ValueFrom` in `GeneratorWith` to use this. Given that every single other `Composable` struct has a `ValueFrom` it is very surprising to _not_ have one here.

So the code in `GeneratorWith` can be:

```
  template <typename Arg>
  using ValueFrom = decltype(e_)::ValueFrom<Arg>;
```

Or better yet, should we make `GeneratorWith` extend from `_GeneratorWith::Composable` or just make the public type `GeneratorWith` be the same as the `_GeneratorWith::Composable`? (And then of course do the same for `TaskWith`?

eventuals/generator.h
```diff
@@ -448,34 +448,9 @@ struct _GeneratorWith {
...
-
-    return future.get();
-  }
-
- private:
-  detail::_GeneratorWith::Composable<Value_, Args_...> e_;
-};
-
-////////////////////////////////////////////////////////////////////////
+// Using to improve the readability.
```
benh, line 451:
```suggestion
// Creating a type alias to improve the readability.
```

benh (approved):

PR #136, opened 2021-11-08
Refactor 'HTTP' to defer the continuation until after the libuv handle is closed

FolMing:


eventuals/http.h
```diff
@@ -607,12 +604,15 @@ struct _HTTP {
     uv_timer_t timer_;
     std::vector<uv_poll_t*> polls_;
 
+    // Result vars.
```
benh, line 607:
```suggestion
    // Response variables.
```

benh (approved):
LGTM! Thank you!

PR #137, opened 2021-11-08
Introduce `Transformer`

onelxj:
`Transformer` is a type-erased that basically allows transforming return types of eventual during the work.
 
Pseudo-code:
```
Stream(<type From>)
  | Transformer<<type From>, <type To>>("transformer labmda")
  | Other stuff that takes <type To>
```

### DIFF
--- a/eventuals/transformer.h
+++ b/eventuals/transformer.h
@@ -1,0 +1,379 @@
+#pragma once
+
+#include <memory>
+#include <optional>
+
+#include "eventuals/callback.h"
+#include "eventuals/stream.h"
+
+////////////////////////////////////////////////////////////////////////
+
+namespace eventuals {
+
+////////////////////////////////////////////////////////////////////////
+
+namespace detail {
+
+////////////////////////////////////////////////////////////////////////
+
+template <typename E_, typename From_, typename To_>
+struct HeapTransformer {
+  struct Adaptor {
+    Adaptor(
+        Callback<TypeErasedStream&>* begin,
+        Callback<std::exception_ptr>* fail,
+        Callback<>* stop,
+        Callback<To_>* body,
+        Callback<>* ended)
+      : begin_(begin),
+        fail_(fail),
+        stop_(stop),
+        body_(body),
+        ended_(ended) {}
+
+    // All functions are called as Continuation after produced Stream.
+    void Begin(TypeErasedStream& stream) {
+      (*begin_)(stream);
+    }
+
+    template <typename... Args>
+    void Body(Args&&... args) {
+      (*body_)(std::forward<decltype(args)>(args)...);
+    }
+
+    template <typename... Args>
+    void Fail(Args&&... args) {
+      (*fail_)(
+          std::make_exception_ptr(
+              std::forward<decltype(args)>(args)...));
+    }
+
+    // NOTE: overload so we don't create nested std::exception_ptr.
+    void Fail(std::exception_ptr exception) {
+      (*fail_)(std::move(exception));
+    }
+
+    void Stop() {
+      (*stop_)();
+    }
+
+    void Ended() {
+      (*ended_)();
+    }
+
+    // Already registered in 'adapted_'
+    void Register(Interrupt&) {}
+
+    Callback<TypeErasedStream&>* begin_;
+    Callback<std::exception_ptr>* fail_;
+    Callback<>* stop_;
+    Callback<To_>* body_;
+    Callback<>* ended_;
+  };
+
+  HeapTransformer(E_ e)
+    : adapted_(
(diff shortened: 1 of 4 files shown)


eventuals/transformer.h
```diff
@@ -0,0 +1,390 @@
...
+
+      Dispatch(Action::Body, std::forward<From>(from)...);
+    }
+
+    void Ended() {
+      k_.Ended();
+    }
+
+    void Register(Interrupt& interrupt) {
+      k_.Register(interrupt);
```
benh, line 194:
```suggestion
      interrupt_ = &interrupt;
      k_.Register(interrupt);
```

eventuals/transformer.h
```diff
@@ -0,0 +1,390 @@
...
+    void Dispatch(
+        Action action,
+        std::optional<From_>&& from = std::nullopt,
+        std::optional<std::exception_ptr>&& exception = std::nullopt) {
+      dispatch_(
+          action,
+          std::move(exception),
+          std::move(from),
+          e_,
+          *interrupt_,
```
benh, line 206: Isn't this always a `nullptr`!? Do you have a test for this?

test/transformer.cc
```diff
@@ -0,0 +1,39 @@
...
+#include "gtest/gtest.h"
+
+using eventuals::Collect;
+using eventuals::Iterate;
+using eventuals::Map;
+using eventuals::Transformer;
+
+using testing::ElementsAre;
+
+TEST(Transformer, Succeed) {
```
benh, line 19: What about failure tests, interrupt tests?

eventuals/transformer.h
```diff
@@ -0,0 +1,390 @@
...
+                std::move(from.value()),
+                interrupt,
+                std::move(begin),
+                std::move(fail),
+                std::move(stop),
+                std::move(body),
+                std::move(ended));
+            break;
+          case Action::Fail:
+            e->Fail(
```
benh, line 300:
```suggestion
          case Action::Fail:
            CHECK(exception);
            e->Fail(
```

eventuals/transformer.h
```diff
@@ -0,0 +1,390 @@
...
+              [](void* e) {
+                delete static_cast<detail::HeapTransformer<E, From_, To_>*>(e);
+              });
+        }
+
+        auto* e = static_cast<HeapTransformer<E, From_, To_>*>(e_.get());
+
+        switch (action) {
+          case Action::Body:
+            e->Body(
```
benh, line 290:
```suggestion
          case Action::Body:
            CHECK(from);
            e->Body(
```

eventuals/transformer.h
```diff
@@ -0,0 +1,390 @@
...
+    void Fail(Args&&... args) {
+      std::exception_ptr exception;
+
+      if constexpr (sizeof...(args) > 0) {
+        exception = std::make_exception_ptr(
+            std::forward<decltype(args)>(args)...);
+      } else {
+        exception = std::make_exception_ptr(
+            std::runtime_error("empty error"));
+      }
```
benh, line 171: After thinking about this more, let's just make this a compile time error.
```suggestion
      static_assert(sizeof...(args) > 0, "Transformer expects Fail() to be called with an argument");
      auto exception = std::make_exception_ptr(
          std::forward<decltype(args)>(args)...);
```

benh (changes requested):
Bumblebee! 🐝🚕

eventuals/transformer.h
```diff
@@ -0,0 +1,375 @@
...
+        Callback<TypeErasedStream&>&&,
+        Callback<To_>&&,
+        Callback<std::exception_ptr>&&,
+        Callback<>&&,
+        Callback<>&&>
+        dispatch_;
+  };
+};
+
+} // namespace detail
```
benh, line 343:
```suggestion
};

////////////////////////////////////////////////////////////////////////

} // namespace detail
```

eventuals/transformer.h
```diff
@@ -0,0 +1,375 @@
...
+    auto [future, k] = Terminate(std::move(e_));
+    k.Start();
+    return future.get();
+  }
+
+ private:
+  detail::_Transformer::Composable<From_, To_> e_;
+};
+
+} // namespace eventuals
```
benh, line 373:
```suggestion
};

////////////////////////////////////////////////////////////////////////

} // namespace eventuals
```

benh (changes requested):
Two more tests please, one test where the body of the transformer propagates a stop (you can just make the body of the `Map` be an `Eventual` that does `k.Stop()`), and one test where it propagates a failure. Thank you!

benh (approved):

PR #139, opened 2021-11-11
Introduce `Pipe` class

ArthurBandaryk:
`Pipe` class implements a synchronized queue. It is generic and contains
`std::deque<t> values_`. This class is inherited from `Synchronizable`
(`eventuals/lock.h`).

### DIFF
--- a/eventuals/pipe.h
+++ b/eventuals/pipe.h
@@ -1,0 +1,79 @@
+#include <deque>
+
+#include "eventuals/callback.h"
+#include "eventuals/lock.h"
+#include "eventuals/map.h"
+#include "eventuals/repeat.h"
+#include "eventuals/then.h"
+#include "eventuals/until.h"
+
+////////////////////////////////////////////////////////////////////////
+
+namespace eventuals {
+
+////////////////////////////////////////////////////////////////////////
+
+template <typename T>
+class Pipe : public Synchronizable {
+ public:
+  auto Write(T&& value) {
+    return Synchronized(Then([this, value = std::move(value)]() mutable {
+      if (!is_closed_) {
+        values_.emplace_back(std::move(value));
+        notify_();
+      }
+    }));
+  }
+
+  auto Read() {
+    return Repeat()
+        | Synchronized(
+               Wait([this](auto notify) {
+                 notify_ = std::move(notify);
+                 return [this]() {
+                   return values_.empty() && !is_closed_;
+                 };
+               })
+               | Map([this]() {
+                   if (!values_.empty()) {
+                     auto value = std::move(values_.front());
+                     values_.pop_front();
+                     return std::make_optional(std::move(value));
+                   } else {
+                     CHECK(is_closed_);
+                     return std::optional<T>();
+                   }
+                 }))
+        | Until([](auto& value) {
+             return !value.has_value();
+           })
+        | Map([](auto&& value) {
+             CHECK(value);
+             return std::move(*value);
+           });
+  }
+
+  auto Close() {
+    return Synchronized(Then([this]() {
+      is_closed_ = true;
+      notify_();
+    }));
+  }
+
+  auto Size() {
+    return Synchronized(Then([this]() {
+      return values_.size();
+    }));
+  }
+
+ private:
+  Callback<> notify_ = []() {};
+  std::deque<T> values_;
+  bool is_closed_ = false;
+};
+
+////////////////////////////////////////////////////////////////////////
+
+} // namespace eventuals
+
+////////////////////////////////////////////////////////////////////////
--- a/test/pipe.cc
+++ b/test/pipe.cc
@@ -1,0 +1,86 @@
+#include "eventuals/pipe.h"
+
+#include <string>
+#include <thread>
+
+#include "eventuals/collect.h"
+#include "eventuals/terminal.h"
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+using eventuals::Collect;
+using eventuals::Pipe;
+
+using testing::ElementsAre;
+
+TEST(Pipe, UniqueValue) {
+  Pipe<int> pipe;
+
+  *pipe.Write(1);
+  *pipe.Close();
+
+  auto e = [&pipe]() {
+    return pipe.Read()
+        | Collect<std::vector<int>>();
+  };
+
+  EXPECT_THAT(*e(), ElementsAre(1));
+}
+
+
+TEST(Pipe, Values) {
+  Pipe<int> pipe;
+
+  std::thread t([&pipe]() {
+    for (size_t i = 1; i <= 5; ++i) {
+      *pipe.Write(i);
+    }
+    *pipe.Close();
+  });
+
+  auto e = [&pipe]() {
+    return pipe.Read()
+        | Collect<std::vector<int>>();
+  };
+
+  EXPECT_THAT(*e(), ElementsAre(1, 2, 3, 4, 5));
+
+  t.join();
+}
+
+
+TEST(Pipe, Close) {
+  Pipe<int> pipe;
+
+  *pipe.Write(1);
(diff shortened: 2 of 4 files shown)


test/pipe.cc
```diff
@@ -0,0 +1,96 @@
...
+TEST(Pipe, Values) {
+  Pipe<int> pipe;
+
+  auto do_write = [](Pipe<int>& pipe) {
+    for (size_t i = 1; i <= 5; ++i) {
+      *pipe.Write(i);
+    }
+  };
+
+  std::thread t{do_write, std::ref(pipe)};
```
benh, line 45: It's easier to read inline lambdas and to capture by ref in the lambda vs via `std::ref`.
```suggestion
  std::thread t([&pipe]() {
    for (size_t i = 1; i <= 5; ++i) {
      *pipe.Write(i);
    }
  });
```

test/pipe.cc
```diff
@@ -0,0 +1,96 @@
...
+    return pipe.Read()
+        | Loop<int>()
+              .context(/*sum = 0*/ 0)
+              .body([](auto&& sum, auto& repeated, auto&& value) {
+                sum += value;
+                repeated.Done();
+              })
+              .ended([](auto& sum, auto& k) {
+                k.Start(std::move(sum));
+              });
```
benh, line 28: Let's simplify and use something like `Collect`!

test/pipe.cc
```diff
@@ -0,0 +1,96 @@
...
+                sum += value;
+                if (sum <= 10) {
+                  repeated.Next();
+                } else {
+                  repeated.Done();
+                }
+              })
+              .ended([](auto& sum, auto& k) {
+                k.Start(std::move(sum));
+              });
```
benh, line 61: Let's simplify and use `Collect`!

test/pipe.cc
```diff
@@ -0,0 +1,96 @@
...
+              .body([](auto&& sum, auto& repeated, auto&& value) {
+                sum += value;
+                if (sum <= 3)
+                  repeated.Next();
+                else
+                  repeated.Done();
+              })
+              .ended([](auto& sum, auto& k) {
+                k.Start(std::move(sum));
+              });
```
benh, line 91: Let's use `Collect` here too.

eventuals/pipe.h
```diff
@@ -0,0 +1,66 @@
...
+  auto Close() {
+    return Synchronized(Then([this]() {
+      is_closed_ = true;
+      notify_();
+    }));
+  }
+
+  size_t values_size() const {
+    return values_.size();
+  }
```
benh, line 58: This is not safe! Everything else is `Synchronized`.

test/pipe.cc
```diff
@@ -0,0 +1,96 @@
...
+              .body([](auto&& sum, auto& repeated, auto&& value) {
+                sum += value;
+                repeated.Done();
+              })
+              .ended([](auto& sum, auto& k) {
+                k.Start(std::move(sum));
+              });
+  };
+
+  EXPECT_EQ(1, pipe.values_size());
```
benh, line 31: Don't worry about checking this, just make sure all the values you wrote on the pipe were read on the pipe.

test/pipe.cc
```diff
@@ -0,0 +1,96 @@
...
+                  repeated.Next();
+                else
+                  repeated.Done();
+              })
+              .ended([](auto& sum, auto& k) {
+                k.Start(std::move(sum));
+              });
+  };
+
+  EXPECT_EQ(3, pipe.values_size());
```
benh, line 94:
```suggestion
```

test/pipe.cc
```diff
@@ -0,0 +1,96 @@
...
+  EXPECT_EQ(0, *e());
```
benh, line 95: I'm not sure how this is supposed to be equal to 0? Aren't you doing `sum += value;` above? Shouldn't this be strictly greater than 0?

eventuals/pipe.h
```diff
@@ -0,0 +1,66 @@
...
+
+namespace eventuals {
+
+template <typename T>
+class Pipe : public Synchronizable {
+ public:
+  auto Write(T&& value) {
+    return Synchronized(Then([this, value = std::move(value)]() mutable {
+      values_.emplace_back(std::move(value));
+      notify_();
```
benh, line 18:
```suggestion
      if (!closed_) {
        values_.emplace_back(std::move(value));
        notify_();
      }
```

eventuals/pipe.h
```diff
@@ -0,0 +1,66 @@
...
+               })
+               | Map([this]() {
+                   if (!is_closed_) {
+                     CHECK(!values_.empty());
+                     auto value = std::move(values_.front());
+                     values_.pop_front();
+                     return std::make_optional(std::move(value));
+                   } else {
+                     return std::optional<T>();
+                   }
```
benh, line 39:
```suggestion
                   if (!values_.empty()) {
                     auto value = std::move(values_.front());
                     values_.pop_front();
                     return std::make_optional(std::move(value));
                   } else {
                     CHECK(closed_);
                     return std::optional<T>();
                   }
```

benh (changes requested):

test/pipe.cc
ArthurBandaryk: I like your idea of using `Collect`. The only thing that is bugging me right now is that the users should always remember to use `*pipe.Close()` in order to make the stream done. If not, we will never get to `.ended` callback in the `Collect()` function.
Actually I've also forgot to call `Close()` :)

test/pipe.cc
ArthurBandaryk: Check my comment above:)

test/pipe.cc
ArthurBandaryk: Check my comment above:)

test/pipe.cc
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  auto e = [&pipe]() {
+    return pipe.Read()
+        | Collect<std::vector<int>>();
+  };
+
+  EXPECT_EQ(
+      (std::vector<int>{1, 2, 3, 4, 5}),
+      *e());
```
benh, line 46:
```suggestion
  EXPECT_THAT(*e(), ElementsAre(1, 2, 3, 4, 5));
```
See for example how it's done in [test/range.cc](https://github.com/3rdparty/eventuals/blob/main/test/range.cc).

test/pipe.cc
```diff
@@ -0,0 +1,66 @@
...
+  *pipe.Write(2);
+  *pipe.Close();
+  *pipe.Write(3);
+
+  auto e = [&pipe]() {
+    return pipe.Read()
+        | Collect<std::vector<int>>();
+  };
+
+  EXPECT_EQ((std::vector<int>{1, 2}), *e());
```
benh, line 65:
```suggestion
  EXPECT_THAT(*e(), ElementsAre(1, 2));
```

test/pipe.cc
benh: That's ok that they have to remember to call `Close()`! That's part of the API!

test/pipe.cc
benh: I think it's fine to have a `Size()` function, it's not adding anything extra in the tests. But you should create an separate test just for it since you added it please!

eventuals/pipe.h
```diff
@@ -0,0 +1,73 @@
...
+                 }))
+        | Until([](auto& value) {
+             return !value.has_value();
+           })
+        | Map([](auto&& value) {
+             CHECK(value);
+             return std::move(*value);
+           });
+  }
+  auto Close() {
```
benh, line 51:
```suggestion
  }

  auto Close() {
```

benh (changes requested):

benh (approved):
💪

PR #141, opened 2021-11-12
Check all files on correct code format

ArthurBandaryk:
Like in `eventuals-grpc` repo we should check all files on corect code format for every `push` or `pull request`.

### DIFF
--- a/eventuals/task.h
+++ b/eventuals/task.h
@@ -170,7 +170,8 @@ struct _TaskWith {
       } else {
         static_assert(
             sizeof...(from) > 0,
-            "Expecting \"from\" argument for 'Task<From, To>' but no argument passed");
+            "Expecting \"from\" argument for 'Task<From, To>' \
+but no argument passed");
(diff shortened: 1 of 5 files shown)


eventuals/task.h
```diff
@@ -170,7 +170,8 @@ struct _TaskWith {
       } else {
         static_assert(
             sizeof...(from) > 0,
-            "Expecting \"from\" argument for 'Task<From, To>' but no argument passed");
+            "Expecting \"from\" argument for 'Task<From, To>' \
+but no argument passed");
```
onelxj, line 174: add indent there

onelxj (changes requested):

eventuals/task.h
FolMing: Omg, that line break is atrocious :D

benh:
@ArthurBandaryk can you update this review without the `tools` file changes but just the formatting issues?

PR #146, opened 2021-12-01
Replace EventLoop::Run() with EventLoop::RunUntil(future)

FolMing:


eventuals/event-loop.h
```diff
@@ -841,9 +851,13 @@ inline auto& Clock() {
 ////////////////////////////////////////////////////////////////////////
 
 inline std::chrono::nanoseconds EventLoop::Clock::Now() {
+  CHECK(!loop_.running_);
```
benh, line 854: Why can't the loop be running?

eventuals/event-loop.h
```diff
@@ -841,9 +851,13 @@ inline auto& Clock() {
...
+
   if (Paused()) { // TODO(benh): add 'unlikely()'.
     return *paused_ + advanced_;
   } else {
+    // uv_now() doesn't update time by itself.
+    uv_update_time(loop_);
```
benh, line 860: Hmm, I'm still not sure based on this comment why we need to update the time. I believe you that you ran into an issue, I just don't know what it is and need you to elaborate this comment because it sounds tricky!

benh (changes requested):

eventuals/event-loop.h
FolMing: The time inside uv_loop is stored as uint64_t variable and is updated every iteration of uv_run. Its update is done non-atomically, hence trying to read the clock while running an EventLoop in another thread is UB and will return an incorrect result.

benh: I'm still not following. Based on your concerns it seems like we should be able to do `Now()` at the very least if we're `InEventLoop()`, which is exactly when most libuv users would likely call `uv_now()` themselves.

Given this `CHECK()` the only time you could really use `Now()` is in testing before the event loop has started, which might be what triggered you to make this change but it's definitely not how most users are going to want to use `Now()` when the event loop is _always_ running.

FolMing: Updated!

eventuals/event-loop.h
FolMing: Added a comment!

eventuals/event-loop.h
```diff
@@ -844,6 +854,17 @@ inline std::chrono::nanoseconds EventLoop::Clock::Now() {
   if (Paused()) { // TODO(benh): add 'unlikely()'.
     return *paused_ + advanced_;
   } else {
+    CHECK(loop_.in_event_loop_ || !loop_.running_)
```
benh, line 857:
```suggestion
    CHECK(InEventLoop() || !loop_.Running())
```

eventuals/event-loop.h
```diff
@@ -844,6 +854,17 @@ inline std::chrono::nanoseconds EventLoop::Clock::Now() {
...
+        << "Calling Clock::Now() from a different thread "
+        << "while EventLoop is running will create a data race";
+    // uv_now() doesn't update time inside uv_loop by itself.
+    // Updating it by ourselves will make sure that we return the most
+    // accurate current time.
+    // NOTE(folming): While adding support for asio::io_context
+    // I've noticed that adding io_context.poll() after uv_run() causes
+    // some tests to fail due to Clock::Now() returning an inaccurate value.
```
benh, line 865: What does "inaccurate" here mean? This comment makes it sound like by changing the time to something "later", all tests pass. But since not all tests were failing in the first place, it sounds like something deterministic is going on that should be figured out, and while adding `uv_update_time()` fixes it, does it actually fix the problem or just mask the problem? If you believe it really fixes the issue, then we really need to articulate what the issue is in this comment.

benh (changes requested):

eventuals/event-loop.h
benh: Seems like we're still missing a comment that explains why we ned to do this regardless of the fact that `uv_now()` doesn't update time by itself?

eventuals/event-loop.h
benh: Did you push the new code? I'm still seeing the old code it seems like.

PR #148, opened 2021-12-01
Add support for asio::io_context inside EventLoop

FolMing:


benh (changes requested):
What about if someone calls `RunForever()`? Does that also do the right thing for `asio`?

PR #149, opened 2021-12-01
Introduce TCP API

FolMing:


benh:
@FolMing do you have any tests that demonstrate how the new abstractions are meant to be used?

FolMing:
> @FolMing do you have any tests that demonstrate how the new abstractions are meant to be used?

test/tcp.cc

eventuals/tcp.h
```diff
@@ -0,0 +1,1455 @@
...
+
+      void Register(class Interrupt& interrupt) {
+        k_.Register(interrupt);
+
+        handler_.emplace(&interrupt, [this]() {
+          if (!completed_) {
+            completed_ = true;
+            k_.Stop();
+          }
+        });
```
benh, line 175: There's a race here with `Start()`. The reason why it's safe to follow this pattern with libuv is because we do `loop_.Submit()` which ensures that the interrupt handler and start will execute serially.

I also don't think that this interrupt handler really buys us very much, as in, there's nothing asynchronous about `open()`, and we don't need to support interrupting things that aren't asynchronous just haven't started yet.

I could see us saving a pointer to `interrupt` and then in `Start()` check to see if we've been interrupted and then do a `k.Stop()` instead of doing the open.

Looking at the rest of this file makes it seem like there are quite a few places where we don't have asynchronous code that we can just replace with an `Eventual()`, for example:

```cpp
class Socket {
 public:
  auto Open() {
    return Eventual<void>()
      .interruptible()
      .start([](auto& k, Interrupt::Handler& handler) {
        if (handler.interrupt().Triggered()) {
          k.Stop();
        } else {
          ...
          socket_.open(..., error);
          if (!error) {
            k.Start();
          } else {
            k.Fail(..);
          }
          ...
        }
      });
  }
};
```

The actual asynchronous functions, like the ones which do `async_read()` or `async_write()` can also follow this pattern because if I understand correctly from reading the documentation as well as from seeing the code in this PR we don't have any way of interrupting them except for stopping the `io_context`, which doesn't seem like something we can do. So, if they aren't interruptible, we might as well follow the simpler pattern.

FolMing: .cancel() cancels all asynchronous operations associated with the acceptor/socket.

benh: Ok great! Is it safe to be called from any thread?

If the answer is yes, then I still like the strategy suggested by instead of just checking if the interrupt has been triggered we can add a handler which will call `cancel()`!

FolMing: That's a great question, I do not see it being answered explicitly by documentation, I'll do some testing on it.

eventuals/event-loop.cc
```diff
@@ -133,11 +133,30 @@ void EventLoop::DestructDefault() {
...
+    // NOTE: we'll truly run forever because handles like 'async_' will
+    // keep the loop alive forever.
+    uv_run(&Default().loop_, UV_RUN_DEFAULT);
+
+    // We should never get out of run.
+    CHECK(false);
+  });
+
+  auto asio_thread = std::thread([]() {
+    auto work_guard = asio::make_work_guard(Default().io_context_);
```
benh, line 149: Let's add a comment that we create `work_guard` so that we'll run forever.

eventuals/event-loop.cc
```diff
@@ -133,11 +133,30 @@ void EventLoop::DestructDefault() {
...
+
+    Default().io_context_.restart();
+    Default().io_context_.run();
+
+    // We should never get out of run.
+    CHECK(false);
```
benh, line 155:
```suggestion
    LOG(FATAL) << "unreachable";
```

eventuals/event-loop.cc
```diff
@@ -133,11 +133,30 @@ void EventLoop::DestructDefault() {
...
+  Default().running_ = true;
+  Default().in_event_loop_ = true;
+
+  auto uv_thread = std::thread([]() {
+    // NOTE: we'll truly run forever because handles like 'async_' will
+    // keep the loop alive forever.
+    uv_run(&Default().loop_, UV_RUN_DEFAULT);
+
+    // We should never get out of run.
+    CHECK(false);
```
benh, line 145:
```suggestion
    LOG(FATAL) << "unreachable";
```

eventuals/tcp.h
```diff
@@ -0,0 +1,583 @@
...
+    struct Data {
+      Socket* socket;
+      Protocol protocol;
+    };
+    Data data = {this, protocol};
+
+    return Eventual<void>()
+        .interruptible()
+        .context(data)
+        .start([](auto& context, auto& k, Interrupt::Handler& handler) {
```
benh, line 46:
```suggestion
        .start([this, protocol](auto& context, auto& k, Interrupt::Handler& handler) {
```

eventuals/tcp.h
```diff
@@ -0,0 +1,583 @@
...
+          asio::post(
+              context.socket->io_context(),
+              [context,
+               &k,
+               &handler]() {
+                if (handler.interrupt().Triggered()) {
+                  k.Stop();
+                } else {
+                  asio::error_code ec;
```
benh, line 55:
```suggestion
                  asio::error_code error;
```

eventuals/tcp.h
```diff
@@ -0,0 +1,583 @@
...
+                          ec);
+                      break;
+                    default:
+                      k.Fail("Unknown protocol");
+                      return;
+                  }
+
+                  context.socket->protocol_ = context.protocol;
+
+                  if (!ec) {
```
benh, line 75: It reads really nice with `error`:
```suggestion
                  if (!error) {
```
😊

eventuals/tcp.h
```diff
@@ -0,0 +1,583 @@
...
+  // If there is any other operation on acceptor
+  // on a different thread, return value might be incorrect.
+  bool IsOpen() {
+    return underlying_handle().is_open();
+  }
+
+  // Maybe thread-unsafe.
+  // If there is any other operation on acceptor
+  // on a different thread, return value might be incorrect.
+  uint16_t BindedPort() {
```
benh, line 182:
```suggestion
  uint16_t BoundPort() {
```

eventuals/tcp.h
```diff
@@ -0,0 +1,583 @@
...
+    return underlying_handle().local_endpoint().port();
+  }
+
+  // Maybe thread-unsafe.
+  // If there is any other operation on acceptor
+  // on a different thread, return value might be incorrect.
+  std::string BindedIp() {
```
benh, line 189:
```suggestion
  std::string BoundAddress() {
```

eventuals/tcp.h
```diff
@@ -0,0 +1,583 @@
...
+  }
+
+ private:
+  EventLoop& loop_;
+  asio::ip::tcp::socket socket_;
+  Protocol protocol_ = Protocol::NOT_SET;
+
+  asio::ip::tcp::socket& underlying_handle() {
+    return socket_;
+  }
```
benh, line 200: Is this because you don't want to just use `socket_` everywhere? It adds unnecessary complexity.

eventuals/tcp.h
```diff
@@ -0,0 +1,583 @@
...
+                  default:
+                    completed_ = true;
+                    k_.Fail("Unknown protocol");
+                    return;
+                }
+
+                if (ec) {
+                  completed_ = true;
+                  k_.Fail(ec.message());
+                }
```
benh, line 254:
```suggestion
                  k_.Fail(ec.message());
                  return;
                }
```

eventuals/tcp.h
```diff
@@ -0,0 +1,583 @@
...
+
+  asio::ip::tcp::socket& underlying_handle() {
+    return socket_;
+  }
+
+  asio::io_context& io_context() {
+    return loop_.io_context();
+  }
+
+  struct _Connect {
```
benh, line 206: It's definitely not obvious why you can just use `Eventual` here as well? It'll remove having to implement `Fail()` and `Stop()` which are just copied code.

eventuals/tcp.h
```diff
@@ -0,0 +1,583 @@
...
+ public:
+  Socket(EventLoop& loop = EventLoop::Default())
+    : loop_(loop),
+      socket_(loop.io_context()) {}
+
+  Socket(const Socket& that) = delete;
+  Socket(Socket&& that)
+    : loop_(that.loop_),
+      socket_(std::move(that.socket_)),
+      protocol_(that.protocol_) {}
```
benh, line 31: Because you're grabbing `this` in the some of the code below you can't actually move this safely. I know we had discussed move semantics once upon a time and one of the upsides of `asio::ip::tcp::socket` is that it is moveable, but we'll have to do something else to support this. 

Ideally we could use something like `stout::borrowed_ptr` here but it doesn't have sufficient functionality yet. Instead we could add a `std::atomic<bool> borrowed_` member here which we set to true whenever we borrow `this` inside of the `Eventual` functions. Then when we're sure that we're done borrowing `this` we can set `borrowed_ = false` and then the object can be moveable again. We can `CHECK(!borrowed_.load());` in the move constructor.

Let me know if you'd like to have a synchronous discussion on this.

eventuals/tcp.h
```diff
@@ -0,0 +1,583 @@
...
+ public:
+  Acceptor(EventLoop& loop = EventLoop::Default())
+    : loop_(loop),
+      acceptor_(loop.io_context()) {}
+
+  Acceptor(const Acceptor& that) = delete;
+  Acceptor(Acceptor&& that)
+    : loop_(that.loop_),
+      acceptor_(std::move(that.acceptor_)),
+      protocol_(that.protocol_) {}
```
benh, line 378: Same comment here re: move semantics as `Socket`.

benh (changes requested):
I'm seeing some light at the end of the tunnel!

🙌

eventuals/tcp.h
FolMing: It actually makes everything easier. You will see it when we add ssl::Socket which has a stream wrapper on top of socket.

benh: Can you leave a code sample here? Otherwise we should add it later because I can't be the best reviewer I can be when pieces of code are missing!

benh:
👋 @FolMing!

Let me know if/when you'd like me to give another review!

FolMing:
> 👋 @FolMing!
> 
> Let me know if/when you'd like me to give another review!

Working on it!

PR #154, opened 2021-12-04
Introduce 'WaitForSignal()'

benh:
The current helpers for signal handling are relatively low-level as
they were designed to wrap the low-level 'uv_signal*()' family of
functions. In practice we often want to be able to specify multiple
signals and find out which signal was raised; this commit introduces
the helper 'WaitForSignal()' which does just that.

### DIFF
--- a/eventuals/event-loop.h
+++ b/eventuals/event-loop.h
@@ -472,25 +472,25 @@ class EventLoop : public Scheduler {
     return clock_;
   }
 
-  auto Signal(const int signum);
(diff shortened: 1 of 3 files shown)


rjhuijsman (approved):
I like it!

benh:
FYI, added another commit here which just renames `Signal()` to `WaitForSignal()` and `WaitForSignal()` to `WatiForOneOfSignals()` because the naming of `Signal()` seemed unfortunately ambiguous.

PR #156, opened 2021-12-10
Refactor workflow into an action

ArthurBandaryk:
We call the `composite` action (`action.yml`) from `3rdparty/dev-tools` submodule. This action helps us to check files for correct code style, no matter if they have `.cc | .h ` or `.bzl` extensions.


.github/workflows/check_code_style.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Initialize and update submodule
+        run: git submodule update --init
+
+      # Call the composite action to check files
+      # for correct code style. This action (action.yml)
+      # is in `3rdparty/dev-tools` submodule.
+      - uses: 3rdparty/dev-tools@main
```
benh, line 23: This is unfortunate naming. It looks like though that we _can_ have multiple actions per repository, see [here](https://github.community/t/possible-to-have-multiple-actions-in-a-single-repo/159823). Then we can put `action.yml` into the `check-code-style` directory of `dev-tools` and do:
```suggestion
      - uses: 3rdparty/dev-tools/check-code-style@main
```

.github/workflows/check_code_style.yml
```diff
@@ -0,0 +1,23 @@
...
+      - "main" 
+
+jobs:
+  check_code_style:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Initialize and update submodule
+        run: git submodule update --init
```
benh, line 18: Should we also be recursive?
```suggestion
        run: git submodule update --init --recursive
```

benh (changes requested):

.github/workflows/check_code_style.yml
ArthurBandaryk: Yep, I think this is a great idea!

.github/workflows/check_code_style.yml
ArthurBandaryk: This [PR](https://github.com/3rdparty/dev-tools/pull/14) is waiting for you @benh !

benh: Ok approved! Now you can update and commit!

.github/workflows/check_code_style.yml
```diff
@@ -0,0 +1,23 @@
...
+
+      # Call the composite action to check files
+      # for correct code style. This action (action.yml)
```
benh, line 21:
```suggestion
      # for correct code style. This action
```

benh (approved):

PR #157, opened 2021-12-18
Refactor 'Task' to be a non-templated type

benh:
There are a few commits in here in support of the primary commit which has the same name as the PR title.


eventuals/expected.h
```diff
@@ -25,30 +25,66 @@ Overloaded(Fs...) -> Overloaded<Fs...>;
...
+    : variant_(std::move(that).template convert<Value_>()) {}
+
+  template <typename U>
+  _Expected(const _Expected<U>& that)
+    : variant_(that.template convert<Value_>()) {}
+
+  // Providing 'k()' in order to compose with other eventuals.
+  template <typename Arg, typename K>
+  auto k(K k) && {
+    return Eventual<Value_>([variant = std::move(variant_)](auto& k) mutable {
```
onelxj, line 79: should it propagate Fail if we start an Eventual from Stop for example?

benh: This won't propagate a `Fail()` if it gets a `Stop()`, it'll just propagate the `Stop()`. It'll only propagate `Fail()` if it gets a `Start()`.

onelxj: I understand, I'm just speaking about what if we do in Stop k.Fail(std::move(std::get<1>(variant))) in case when variant.index() != 0 if it exists

benh: As stated, 'Stop()' will continue to get propagated, regardless of what `variant.index()` is. If we propagate `Fail()` instead then someone downstream might attempt to recover from the error instead of knowing that we want to actually stop, hence we propagate `Stop()`. I'll add this as a comment.

onelxj (approved):
LGTM!

PR #160, opened 2021-12-21
Allow to use custom `user.bazelrc` config

onelxj:


.bazelrc
```diff
@@ -1,3 +1,6 @@
+# Allow users to add local preferences.
+try-import %workspace%/user.bazelrc
+
 # Specific Bazel build/test options.
 
 build --enable_platform_specific_config
```
benh, line 6:
```suggestion
# Specific Bazel build/test options.

build --enable_platform_specific_config

# Allow users to add local preferences.
try-import %workspace%/user.bazelrc
```

benh (approved):
Let's swap the ordering, see the suggestion. Thanks!

PR #161, opened 2021-12-22
Remove redundant code

ArthurBandaryk:
We wanna remove this part of code. Check [this](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md) link. They've updated repo, so on Ubuntu 20.04 we already have preinstalled Clang. That's why ` KyleMayes/install-llvm-action` is already not relevant for us.


benh (approved):
Let's also remove from `eventuals-grpc` thanks!

PR #165, opened 2021-12-23
Document recommended git pre-commit hook setup

CodingCanuck:


README.md
```diff
@@ -559,3 +559,11 @@ string s = *e; // BLOCKING! Only use in tests.
 #### Scheduling and Memory
 
 *... to be completed ...*
+
+
+## Contributing
+
+The eventuals library maintains a code style that's enforced by GitHub
+pre-commit hooks. See the
```
benh, line 567:
```suggestion
The eventuals library maintains a code style that's enforced by `git`
pre-commit [hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks). See the
```

benh (approved):
You have wrapped the text (I'm guessing at 80 characters 😜) which the rest of the file does not do, so for now let's stay consistent with what's already there and not wrap please.

Happy to open a discussion on this and document for posterity. Personally I'm a fan of the wrap (what you did) because it reads better that way in `emacs`. 🤪 However, I was convinced long ago that it can make for really convoluted diffs, where as the diff on the non-wrap version tends to be very straightforward. Thus, I've just learned to get used to the less than ideal rendering in my editor. Some [folks](https://about.gitlab.com/blog/2016/10/11/wrapping-text/) argue that the editor rendering is more important than the helping a human review the diff. What do you think?

README.md
CodingCanuck: Applied manually.

But I'm confused: isn't it GitHub that runs these hooks? In other words, my local git installation doesn't run these unless I opt-in to a configuration change, meaning my local git doesn't enforce this style / prohibit commits that violate this style. I assume it's GitHub that *enforces* (does not allow me to violate) this style when I try to merge something into main. Is that understanding correct?

So: it's a git hook, that git hook is run by github to enforce style, and you can optionally make your git run the same hook as github does. Does that sound correct?

benh: The `pre-commit` hook is not actually "installed" or used by GitHub. Instead, [this](https://github.com/3rdparty/eventuals/blob/main/.github/workflows/check_code_style.yml) GitHub Workflow is run which uses [this](https://github.com/3rdparty/dev-tools/blob/main/check-code-style/action.yml) GitHub Action, which uses [this](https://github.com/3rdparty/dev-tools/blob/main/check_style.sh) shell script that ultimately calls `clang-format` (and checks the line-length. And FYI @ArthurBandaryk is currently in the process of making it so that both the action and the `pre-commit` hook calls the exact same shell script so that they do the exact same thing whether you're using the local `pre-commit` hook or GitHub is running the workflow. So perhaps what you are trying to say here is something like "GitHub runs a [workflow](https://github.com/3rdparty/eventuals/blob/main/.github/workflows/check_code_style.yml) that enforces the code style but you can also install a [`pre-commit`](https://github.com/3rdparty/dev-tools/blob/main/pre-commit) hook to check the code style locally before submitting a PR"?

CodingCanuck: Thanks, that was very clarifying!

I settled on this:

The eventuals library maintains a code style that's enforced by a GitHub [workflow](https://github.com/3rdparty/eventuals/blob/main/.github/workflows/check_code_style.yml). You can also install a `git` [`pre-commit`](https://github.com/3rdparty/dev-tools/blob/main/pre-commit) [hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) to check the code style locally before sending a pull request: see the [dev-tools README](https://github.com/3rdparty/dev-tools/tree/main#readme) for instructions.

PR #167, opened 2021-12-23
Add HTTP success tests based on ASIO sockets

FolMing:


test/http.cc
```diff
@@ -29,6 +30,132 @@ const char* schemes[] = {"http://", "https://"};
 
 INSTANTIATE_TEST_SUITE_P(Schemes, HTTPTest, testing::ValuesIn(schemes));
 
+
+TEST_P(HTTPTest, GetSuccessASIO) {
+  std::string scheme = GetParam();
+
+  // Currently we don't provide tests for https.
```
rjhuijsman, line 37: Can you expand the comment to explain why, and when we should be able to add https tests?

test/http.cc
```diff
@@ -29,6 +30,132 @@ const char* schemes[] = {"http://", "https://"};
...
+    asio::read_until(sock, request_buffer, "\r\n\r\n", ec);
+    EXPECT_FALSE(ec);
+
+    // Content
+    asio::read_until(sock, request_buffer, "\r\n\r\n", ec);
+    EXPECT_FALSE(ec);
+
+    asio::write(sock, asio::buffer(response, response.size()), ec);
+    EXPECT_FALSE(ec);
+  });
```
rjhuijsman, line 137: This all looks veeerrrry similar to what we have in the GET test. Could we factor this out into a function that the two functions share?

rjhuijsman (changes requested):
This is very helpful! Just two readability requests.

test/http.cc
benh: Agree, let's put this into the `HTTPTest` as that's where we'll do it in the future as well!

benh (changes requested):

FolMing:
Will be added in #178

PR #171, opened 2021-12-24
Use gMock matchers to simplify test assertions

CodingCanuck:
See gMock matcher documentation: http://google.github.io/googletest/gmock_cook_book.html#using-matchers


FolMing (approved):
We definitely should start using those matchers, thanks!

PR #177, opened 2022-01-05
Fix doc-only PRs by providing dummy build results.

### LINKED ISSUE #176: Skipping builds (for markdown changes) prevents submission
I merged #168 to disable Build and Test checks for changes that only affect markdown files.

That's fine, but it's running afoul of our [branch protection rules](https://github.com/3rdparty/eventuals/settings/branch_protection_rules/22183151) that require a passing result from Build and Test in order to merge a change:
Screen Shot 2022-01-05 at 11 26 49 AM
 
Since the build gets skipped it never succeeds, which prevents the submission of markdown-only changes like #170.

This specific issue is covered in this FAQ: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks where the recommended solution is to create a dummy check that only runs on skipped files and emits success: "fix this by creating a generic workflow, with the same name, that will return true in any case"

CodingCanuck:
Fixes #176.

### DIFF
--- a/.github/workflows/nobuild.yml
+++ b/.github/workflows/nobuild.yml
@@ -5,7 +5,7 @@ on:
   pull_request:
     branches:
       - "main"
-    paths-ignore:
+    paths:
       - '**.md'
 
 jobs:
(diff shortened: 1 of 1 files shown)


sp33drac3r (approved):
LGTM

cc: @benh as FYI

PR #178, opened 2022-01-08
HTTP refactor, add CA and Headers functionality

FolMing:


benh:
Really interesting ideas here @FolMing, let's do a braintrust! Would be great to share your ideas with folks and compare/contrast with other libraries out there and the lessons learned from them. Have you looked at Go's [net/http](https://pkg.go.dev/net/http) package? I think it is a relatively modern library we can build on.

FolMing:
> Really interesting ideas here @FolMing, let's do a braintrust! Would be great to share your ideas with folks and compare/contrast with other libraries out there and the lessons learned from them. Have you looked at Go's [net/http](https://pkg.go.dev/net/http) package? I think it is a relatively modern library we can build on.

Let's do it!
I did not look into other libraries, I was thinking only about how to implement what @rjhuijsman specifically wanted.

PR #179, opened 2022-01-11
Introduce http::Client + http::Request and refactor http::Get/Post to use

benh:
Including a few clean up commits as I went along.


eventuals/http.h
```diff
@@ -645,48 +767,66 @@ struct _HTTP {
...
+          .timeout(std::move(timeout))
+          .fields(std::move(fields))
+          .Build());
+}
+
+////////////////////////////////////////////////////////////////////////
+
+inline auto Get(
+    std::string&& url,
+    std::chrono::nanoseconds&& timeout = std::chrono::nanoseconds(0)) {
```
FolMing, line 819: I do see it working with 
`Get("http://someurl.com", std::chrono::seconds(10))`, 
but it will not work with
```
const std::string hostname = "http://someurl.com";
const auto timeout = std::chrono::seconds(10);
Get(hostname, timeout);
```

eventuals/http.h
```diff
@@ -645,48 +767,66 @@ struct _HTTP {
...
 ////////////////////////////////////////////////////////////////////////
 
 inline auto Post(
-    const std::string& url,
-    const PostFields& fields,
-    const std::chrono::nanoseconds& timeout = std::chrono::nanoseconds(0)) {
-  return Post(EventLoop::Default(), url, fields, timeout);
+    std::string&& url,
+    PostFields&& fields,
+    std::chrono::nanoseconds&& timeout = std::chrono::nanoseconds(0)) {
```
FolMing, line 828: Same as above.

test/BUILD.bazel
```diff
@@ -1,53 +1,85 @@
...
+        "pipe.cc",
+        "timer.cc",
+        "filesystem.cc",
+        "dns-resolver.cc",
+        "signal.cc",
+        "http.cc",
+    ],
     copts = copts(),
     deps = [
-        "//:eventuals",
```
FolMing, line 50: I agree that this is not necessary anymore as it's getting propagated from `concurrent-tests`, but don't you think that it should still be in here just to be more explicit that we are testing eventuals library and not anything else?

test/http.cc
```diff
@@ -12,9 +13,115 @@ using eventuals::Terminate;
 
 using eventuals::http::Get;
 using eventuals::http::Post;
-using eventuals::http::PostFields;
 
-class HTTPTest
+class HttpMockServer {
```
FolMing, line 17: Will not work on Windows.

test/http.cc
```diff
@@ -12,9 +13,115 @@ using eventuals::Terminate;
...
+        .WillByDefault([this](int client) {
+          // Receive data up to end of headers.
+          std::string data;
+
+          do {
+            char buffer[1024];
+            size_t bytes = recv(client, buffer, sizeof(buffer), 0);
+            if (bytes <= 0) {
+              PLOG(ERROR) << "Failed to receive headers";
+              close(client);
```
FolMing, line 48: We check return values in each and every other socket calls, but we are not doing it for `close()`, why?

test/http.cc
```diff
@@ -12,9 +13,115 @@ using eventuals::Terminate;
...
+            }
+            data += std::string(buffer, bytes);
+          } while (data.find("\r\n\r\n") == std::string::npos);
+
+          ReceivedHeaders(client, data);
+        });
+
+    ON_CALL(*this, ReceivedHeaders)
+        .WillByDefault([](int client, const std::string& data) {
+          close(client);
```
FolMing, line 59: Same as above.

test/http.cc
```diff
@@ -12,9 +13,115 @@ using eventuals::Terminate;
...
+          Accepted(client);
+        }
+      } while (client >= 0);
+    });
+  }
+
+  ~HttpMockServer() {
+    // Close the server to force 'accept()' to error out and the
+    // thread to exit and be joinable.
+    close(server_);
```
FolMing, line 78: Same as above.

test/http.cc
```diff
@@ -27,13 +134,40 @@ const char* schemes[] = {"http://"};
...
+        const std::string response =
+            "HTTP/1.1 200 OK\r\n"
+            "Content-Length: 25\r\n"
+            "\r\n"
+            "<html>Hello World!</html>\r\n"
+            "\r\n";
+
+        EXPECT_GT(send(client, response.data(), response.size(), 0), 0);
+
+        close(client);
```
FolMing, line 155: Same as above.

test/BUILD.bazel
```diff
@@ -1,53 +1,85 @@
...
+        "expected.cc",
+        "unpack.cc",
+        "transformer.cc",
+        "pipe.cc",
+        "timer.cc",
+        "filesystem.cc",
+        "dns-resolver.cc",
+        "signal.cc",
+        "http.cc",
+    ],
```
FolMing, line 79: Buildifier?

FolMing:
There's also an interesting bug with MOCK_METHOD on Windows: https://github.com/google/googletest/issues/2490

I've seen it fixed in glog v1.11.0, we probably should upgrade our glog dependency in bazel/deps.bzl.

test/http.cc
benh: Replaced with `asio` version!

eventuals/http.h
```diff
@@ -626,67 +852,79 @@ struct _HTTP {
...
-    EventLoop& loop,
-    const std::string& url,
-    const std::chrono::nanoseconds& timeout = std::chrono::nanoseconds(0)) {
+inline auto Client::Do(Request&& request) {
+  // TODO(benh): need 'Client::Default()'.
+  auto& loop = EventLoop::Default();
+
+  if (verify_peer_ && !request.verify_peer()) {
+    request.verify_peer_ = verify_peer_;
+  }
```
FolMing, line 871: This looks ugly! Is there a better way to override request settings with client settings?

eventuals/http.h
```diff
@@ -626,67 +852,79 @@ struct _HTTP {
...
+
+  if (certificate_ && !request.certificate()) {
+    request.certificate_ = certificate_;
+  }
```
FolMing, line 875: ^

eventuals/http.h
```diff
@@ -14,27 +15,230 @@ namespace http {
...
+
+  friend class Client;
+
+  std::string uri_;
+  Method method_;
+  std::vector<std::pair<std::string, std::string>> headers_;
+  std::string body_;
+  std::chrono::nanoseconds timeout_;
+  PostFields fields_;
+  std::optional<bool> verify_peer_;
```
FolMing, line 88: We need something like tribool in here: https://www.boost.org/doc/libs/1_78_0/doc/html/boost/logic/tribool.html

eventuals/http.h
```diff
@@ -14,27 +15,230 @@ namespace http {
...
+    static_assert(!HasMethod_, "Duplicate 'method(...)' for http::Request");
+    request_.method_ = method;
+    return RequestBuilder<HasUri_, true>(std::move(request_));
+  }
+
+  RequestBuilder<HasUri_, HasMethod_> timeout(
+      std::chrono::nanoseconds&& timeout) && {
+    request_.timeout_ = std::move(timeout);
+    return RequestBuilder<HasUri_, HasMethod_>(std::move(request_));
+  }
```
FolMing, line 119: We shouldn't support multiple timeout()'s, should we?

eventuals/http.h
```diff
@@ -14,27 +15,230 @@ namespace http {
...
+  RequestBuilder<HasUri_, HasMethod_> fields(PostFields&& fields) && {
+    request_.fields_ = std::move(fields);
+    return RequestBuilder<HasUri_, HasMethod_>(std::move(request_));
+  }
+
+  RequestBuilder<HasUri_, HasMethod_> verify_peer(bool value) && {
+    // TODO(benh): consider checking that the scheme is 'https'.
+    request_.verify_peer_ = value;
+    return RequestBuilder<HasUri_, HasMethod_>(std::move(request_));
+  }
```
FolMing, line 130: ^

eventuals/http.h
```diff
@@ -14,27 +15,230 @@ namespace http {
...
+
+  RequestBuilder<HasUri_, HasMethod_> certificate(
+      x509::Certificate&& certificate) && {
+    // TODO(benh): consider checking that the scheme is 'https'.
+    request_.certificate_ = std::move(certificate);
+    return RequestBuilder<HasUri_, HasMethod_>(std::move(request_));
```
FolMing, line 136: ^

FolMing (commented):
Not a full review!

eventuals/http.h
```diff
@@ -14,27 +15,230 @@ namespace http {
...
+ public:
+  // Constructs a new http::Request "builder" with the default
+  // undefined values.
+  static RequestBuilder<false, false> Builder();
+
+  const auto& uri() {
+    return uri_;
+  }
+
+  const auto& method() {
```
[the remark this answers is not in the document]
benh: The actual `http::Request` is designed to be immutable. If we think we want to be able to move out values from `http::Request` then I think we can provide `&&` overloads of these functions that return rvalue references for each field.

benh:
> There's also an interesting bug with MOCK_METHOD on Windows: [google/googletest#2490](https://github.com/google/googletest/issues/2490)
> 
> I've seen it fixed in googletest v1.11.0, we probably should upgrade our com_github_google_googletest dependency in bazel/deps.bzl:
> 
> ```
> maybe(
>         http_archive,
>         name = "com_github_google_googletest",
>         url = "https://github.com/google/googletest/archive/release-1.11.0.tar.gz",
>         sha256 = "b4870bf121ff7795ba20d20bcdd8627b8e088f2d1dab299a031c1034eddc93d5",
>         strip_prefix = "googletest-release-1.11.0",
>         repo_mapping = repo_mapping,
>     )
> ```

Added!

FolMing (approved):
LGTM!

PR #180, opened 2022-01-11
Configure clang-format before running checks for code style

ArthurBandaryk:
We should be sure that before running all checks for code style we have clang-format installed, and if yes, we recreate a symlink to the correct version of clang-format in order to start checking all files style with correct clang-format version.

### DIFF
--- a/.github/workflows/check_code_style.yml
+++ b/.github/workflows/check_code_style.yml
@@ -10,9 +10,34 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with: 
+          submodules: 'true'
 
-      - name: Initialize and update submodule
-        run: git submodule update --init --recursive
+      # If we do not have clang-format preinstalled => 
+      # our workflow will fail.
+      - name: Check clang-format version
+        run: clang-format --version
+
+      # IMPORTANT NOTE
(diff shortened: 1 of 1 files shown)


.github/workflows/check_code_style.yml
```diff
@@ -13,9 +13,40 @@ jobs:
...
+      - name: Check clang-format version
+        id: clang_format_status
+        run: clang-format --version
+        continue-on-error: true
+      
+      - name: Exit if we do not have clang-format installed
+        if: steps.clang_format_status.outcome != 'success'
+        run: |
+          echo "There is no clang-format installed."  
+          exit 1
```
FolMing, line 28: Do we need an additional step? If yes, pls add a comment why.

FolMing (approved):
LGTM, except one small thing!

.github/workflows/check_code_style.yml
ArthurBandaryk: You are right, if `clang-format --version` fails => our workflow will fail immediately. So it's redundant , I'll remove it.

ArthurBandaryk:
Do not merge this PR until [PR](https://github.com/3rdparty/dev-tools/pull/17) in dev-tools will be approved and merged!!!

FolMing:
@ArthurBandaryk the PR you've mentioned was merged, should we merge this one as well then?

benh (approved):

PR #181, opened 2022-01-11
Optimize away heap allocation from `Task` synchronous helpers

onelxj:


eventuals/task.h
```diff
@@ -544,15 +610,9 @@ struct Task {
...
+  template <typename To, typename Error>
   static auto Failure(Error error) {
-    // TODO(benh): optimize away heap allocation.
-    return [error = std::make_unique<Error>(std::move(error))]() mutable {
-      return Eventual<_TaskFailure>()
-          .start([&](auto& k) mutable {
-            k.Fail(Error(std::move(*error)));
-          });
-    };
+    return TaskFromToWith<void, To>(std::make_exception_ptr(std::move(error)));
```
benh, line 615: Let's just revert this since then we don't have to specify `To` and both implementations do dynamic memory allocation.

eventuals/task.h
```diff
@@ -513,6 +565,23 @@ struct Task {
...
+
+      To(std::exception_ptr&& error)
+        : TaskFromToWith<From_, To_>(std::move(error)) {}
+
+      // Synchronous failure helper with erased `To_` type.
+      template <typename Error>
+      static auto Failure(Error error) {
+        return TaskFromToWith<void, To_>(
+            std::make_exception_ptr(std::move(error)));
+      }
```
benh, line 584:
```suggestion
```

benh (changes requested):

eventuals/task.h
```diff
@@ -238,34 +252,54 @@ struct _TaskFromToWith {
...
-        std::optional<
-            std::conditional_t<
-                std::is_void_v<From_>,
-                std::monostate,
-                From_>>&&,
-        std::unique_ptr<void, Callback<void*>>&,
-        Interrupt&,
+    // The 'dispatch_' might be both function that creates an eventual
+    // or argument that should be passed throw 'Task' in the 'Start' method
+    // of 'Task::Continuation' used in 'Task::Success'.
```
benh, line 257:
```suggestion
    // The 'dispatch_' is a `std::variant` because its either a function that 
    // creates an eventual or the value from 'Task::Success' that should be 
    // passed on to the continuation.
```

eventuals/task.h
```diff
@@ -360,25 +394,35 @@ struct _TaskFromToWith {
...
```
benh, line 399:
```suggestion
    // See comment in `Continuation` for explanation of `dispatch_` member.
```

eventuals/task.h
```diff
@@ -553,7 +609,7 @@ struct Task {
             k.Fail(Error(std::move(*error)));
           });
     };
-  }
+  };
```
benh, line 612: Doesn't seem like this is necessary?

eventuals/task.h
```diff
@@ -360,25 +394,35 @@ struct _TaskFromToWith {
...
+    // of 'Task::Continuation' used in 'Task::Success'.
+    std::variant<
         std::conditional_t<
             std::is_void_v<To_>,
-            Callback<>&&,
-            Callback<To_>&&>,
-        Callback<std::exception_ptr>&&,
-        Callback<>&&>
+            std::monostate,
+            std::decay_t<To_>>,
```
benh, line 404: Why do you need the `std::decay_t` here and above? If you don't actually need it please remove otherwise please leave a comment!

benh (approved):
Looks good but happy to chat through the need for `std::decay` before you merge.

eventuals/task.h
onelxj: When `To_` is a reference it can't be stored at `std::variant`

variant:1338:7 error: static_assert failed due to requirement - "variant must have no reference alternative"

eventuals/task.h
```diff
@@ -166,18 +166,32 @@ struct _TaskFromToWith {
     Fail = 2,
   };
 
-  template <typename K_, typename From_, typename To_, typename... Args_>
+  template <
+      typename K_,
+      typename From_,
+      typename To_,
+      typename... Args_>
   struct Continuation {
```
CodingCanuck, line 174: Since this has non-trivial methods, the [Google C++ style guide](https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes) would say that this should be a `class` rather than struct.

Probably beyond the scope of this review, though, since it might apply to many files.

eventuals/task.h
```diff
@@ -238,34 +254,56 @@ struct _TaskFromToWith {
     K_ k_;
```
CodingCanuck, line 254: Should these member variables be private?

If they're intentionally public, they probably deserve comments.

eventuals/task.h
```diff
@@ -238,34 +254,56 @@ struct _TaskFromToWith {
...
+    // passed on to the continuation.
+    std::variant<
         std::conditional_t<
             std::is_void_v<To_>,
-            Callback<>&&,
-            Callback<To_>&&>,
-        Callback<std::exception_ptr>&&,
-        Callback<>&&>
+            std::monostate,
+            // Using std::decay_t because std::variant doesn't allow storing
```
CodingCanuck, line 264: How will this work for non-copyable types (e.g. `std::unique_ptr`) and types which are not moveable or copyable (e.g. [`std::mutex`](https://en.cppreference.com/w/cpp/thread/mutex))? Maybe expand this comment.

(I wonder if [std::reference_wrapper](https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper) would work if we want to store a reference)

eventuals/task.h
```diff
@@ -238,34 +254,56 @@ struct _TaskFromToWith {
...
-            std::conditional_t<
-                std::is_void_v<From_>,
-                std::monostate,
-                From_>>&&,
-        std::unique_ptr<void, Callback<void*>>&,
-        Interrupt&,
+    // The 'dispatch_' is a `std::variant` because its either a function that
+    // creates an eventual or the value from 'Task::Success' that should be
+    // passed on to the continuation.
+    std::variant<
```
CodingCanuck, line 260: Gosh, this is a long declaration line.

Some local (private) typedefs might help make this much more readable, e.g., one typedef for the Callback (not sure if any others make sense to pull out)

CodingCanuck (approved):

eventuals/task.h
benh: Let's please add a test for this case!

eventuals/task.h
onelxj: Will refactor in another commit

eventuals/task.h
onelxj: Will refactor in another commit

eventuals/task.h
onelxj: Now this will be applicable only for moveable types. Will add std::reference_wrapper in the future if needed.

CodingCanuck (approved):

benh (approved):

PR #187, opened 2022-01-14
EventLoop::Run() -> EventLoop::RunUntil(), Add support for asio::io_context in EventLoop

FolMing:


benh (approved):
Let's drop the `Add support for asio::io_context inside EventLoop` commit for now but the rest look good!

PR #190, opened 2022-01-17
Fix Windows build

FolMing:


.github/workflows/macos.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         uses: mxschmitt/action-tmate@v3
 
       - name: Test
-        run: bazel test --spawn_strategy=local --copt="-g" --strip="never" --test_output=errors test:eventuals --test_arg=--gtest_shuffle --test_arg=--gtest_repeat=100
+        run: bazel test --spawn_strategy=local --copt="-g" --strip="never" --test_output=errors --experimental_ui_max_stdouterr_bytes=-1 test:eventuals --test_arg=--gtest_shuffle --test_arg=--gtest_repeat=100
```
CodingCanuck, line 39: Can you add a comment explaining why this flag is here? (here and in the other yml files)

CodingCanuck (approved):
Ben's out today, so here's an approval!

test/http-mock-server.h
```diff
@@ -25,6 +25,9 @@ class HttpMockServer {
     virtual std::string Receive() = 0;
     virtual void Send(const std::string& data) = 0;
     virtual void Close() = 0;
+
+   protected:
+    constexpr static size_t buffer_size_ = 4096;
```
benh, line 30:
```suggestion
    constexpr static size_t kBufferSize = 4096;
```

test/http-mock-server.h
```diff
@@ -112,7 +115,9 @@ class HttpMockServer {
     EXPECT_FALSE(error) << error.message();
 
     acceptor_.bind(
-        asio::ip::tcp::endpoint(asio::ip::tcp::v4(), 0),
+        asio::ip::tcp::endpoint(
+            asio::ip::address_v4::from_string("127.0.0.1"),
+            0),
```
benh, line 120: I'd like to add a comment here explaining why we can't use the equivalent of  `INADDR_ANY` and must specify the loopback IP address on Windows.

Also, should we use `asio::ip::address_v1::loopback()`?

test/http-mock-server.h
```diff
@@ -143,7 +148,7 @@ class HttpMockServer {
     auto certificate = x509::Certificate::Builder()
                            .subject_key(rsa::Key(*key))
                            .sign_key(rsa::Key(*key))
-                           .ip(endpoint_.address())
+                           .hostname("localhost")
```
benh, line 151: Let's do:
```suggestion
                           .hostname(host())
```

test/http-mock-server.h
```diff
@@ -306,7 +311,7 @@ class HttpMockServer {
   }
 
   std::string host() const {
-    return endpoint_.address().to_string();
+    return "localhost";
```
benh, line 314: And again let's add a comment that refers to the comment above on why we need to use the loopback address and thus `localhost` here. Thanks!

benh (approved):
Woohoo!

test/http-mock-server.h
FolMing: Are you sure we need a comment on why we should use loopback and not open up our application for any incoming connection? :D

benh:
Merged these into #179!

test/http-mock-server.h
benh: Added!

test/http-mock-server.h
benh: Done!

test/http-mock-server.h
benh: I added a comment!

PR #192, opened 2022-01-19
Introduce clang-tidy review

ArthurBandaryk:


benh:
Is this still in draft state? Can we switch it back to draft if so? Thanks!

PR #194, opened 2022-01-20
Introduce single `build_and_test.yml`  for building and testing eventuals

ArthurBandaryk:
We can use `matrix` and have only 1 .yml file which can do building and running all tests for eventuals on macOS, Ubuntu and Windows.


while-false (approved):
Thanks Arthur!

FolMing:
Is there a way to optimize away those "Debug smth using tmate (if failure)" steps? And also, seeing all of the skipped actions is a bit distracting and confusing.

I would love it if it was more like this:
- Build
- Test
- Debug on any failure

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,56 @@
...
+        if: ${{ matrix.os == 'macos-latest' || matrix.os == 'ubuntu-latest' }} 
+        run: bazel test --spawn_strategy=local --copt="-g" --strip="never" --test_output=errors test:eventuals --test_arg=--gtest_shuffle --test_arg=--gtest_repeat=100
+
+      - name: Debug Test using tmate (if failure)
+        if: ${{ failure() }}
+        uses: mxschmitt/action-tmate@v3
+
+      - name: Test step on Windows
+        if: ${{ matrix.os == 'windows-latest' }} 
+        run: bazel test --spawn_strategy=local --strip="never" --test_output=errors test:eventuals --test_arg=--gtest_shuffle --test_arg=--gtest_repeat=100
```
benh, line 52: Does it work if we do `-c dbg` for Windows? Then we can have a single Build step and single Test step with no `if`. Also let's wrap these long lines, I've been meaning to do that for a while. See suggestion below for both!
 
```suggestion
      - name: Build
        run: bazel build \
               --spawn_strategy=local \
               -c dbg \
               --strip="never" \
               :eventuals

      - name: Test
        run: bazel test \
               --spawn_strategy=local \
               -c dbg \
               --strip="never" \
               --test_output=errors \
               test:eventuals \
               --test_arg=--gtest_shuffle \
               --test_arg=--gtest_repeat=100
```

benh (changes requested):

benh:
> Is there a way to optimize away those "Debug smth using tmate (if failure)" steps? And also, seeing all of the skipped actions is a bit distracting and confusing.
> 
> I would love it if it was more like this:
> 
> * Build
> * Test
> * Debug on any failure

I think that actually will work @FolMing! Let's give it a try @ArthurBandaryk, only include the debug step at the end!

FolMing (approved):
Now it looks super clean, thank you!

benh (approved):

PR #196, opened 2022-01-21
Introduce 'ConditionVariable'

benh:
While we previously had support for a kind of condition variable via
'Wait()' their syntax was low-level. This commit introduces a
higher-level abstraction that is very similar to condition variables
in higher-level languages.

### DIFF
--- a/eventuals/lock.h
+++ b/eventuals/lock.h
@@ -460,11 +460,8 @@ struct _Wait {
               // and we make sure we're not notifiable after the first
               // notification so we don't try and add ourselves to the
               // list of waiters again.
-              //
-              // TODO(benh): make sure *we've* acquired the lock
-              // (where 'we' is the current "eventual").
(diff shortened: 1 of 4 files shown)


eventuals/lock.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <atomic>
+#include <list>
```
FolMing, line 4: This looks unnecessary.

FolMing (commented):
LGTM, although I'm not fully familiar with this part of eventuals!
Not giving an approval since @ArthurBandaryk is the one whose review was requested.

eventuals/lock.h
benh: Oops, remnants of an earlier implementation, THANK YOU!

ArthurBandaryk (approved):

PR #198, opened 2022-01-22
Ensure event loop eventuals always execute in the event loop

benh:
Otherwise it's possible that we'll have race conditions.


FolMing (approved):
Seems reasonable!

benh:
@FolMing thanks for the commit adding the header, I went ahead squashed that in as I also added `<memory>` to all of the files in the commit that I added `std::unique_ptr` to so that we don't run into this in the future!

PR #199, opened 2022-01-23
Use std::map as a return value for Response::headers()

FolMing:
Please use [RFC 7230, section 3.2.2](https://www.rfc-editor.org/rfc/rfc7230.html#section-3.2.2) to fully understand the changes made in this PR!

### DIFF
--- a/eventuals/builder.h
+++ b/eventuals/builder.h
@@ -121,6 +121,74 @@ class FieldWithDefault<Value_, true> : public Field<Value_, true> {
 
 ////////////////////////////////////////////////////////////////////////
 
+template <typename Value_, bool has_>
+class RepeatedField;
+
+template <typename Value_>
+class RepeatedField<Value_, false> : public Field<Value_, false> {
+ public:
+  template <
+      typename Value,
+      std::enable_if_t<std::is_convertible_v<Value, Value_>, int> = 0>
+  RepeatedField(Value&& value)
+    : default_(std::forward<Value>(value)) {}
+
+  RepeatedField(RepeatedField&&) = default;
+
+  template <typename Value>
+  auto Set(Value&& value) {
+    static_assert(std::is_convertible_v<Value, Value_>);
+    return RepeatedField<Value_, true>{std::forward<Value>(value)};
+  }
+
+  auto& value() & {
+    return default_;
+  }
+
+  const auto& value() const& {
+    return default_;
+  }
+
+  auto&& value() && {
+    return std::move(default_);
+  }
+
+  const auto&& value() const&& {
+    return std::move(default_);
+  }
+
+  const auto* operator->() const {
+    return &default_;
+  }
+
+  auto* operator->() {
+    return &default_;
+  }
+
+ private:
+  Value_ default_;
+};
+
+template <typename Value_>
+class RepeatedField<Value_, true> : public Field<Value_, true> {
+ public:
+  template <
+      typename Value,
+      std::enable_if_t<std::is_convertible_v<Value, Value_>, int> = 0>
+  RepeatedField(Value&& value)
+    : Field<Value_, true>(std::forward<Value>(value)) {}
+
+  RepeatedField(RepeatedField&&) = default;
+
+  template <typename Value>
+  auto Set(Value&& value) {
+    static_assert(std::is_convertible_v<Value, Value_>);
+    return RepeatedField<Value_, true>{std::forward<Value>(value)};
+  }
+};
+
+////////////////////////////////////////////////////////////////////////
+
 class Builder {
  protected:
   // Helper that creates a "builder" by calling it's constructor with a
--- a/eventuals/http.h
+++ b/eventuals/http.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <map>
 #include <memory>
 #include <string>
 #include <vector>
@@ -30,7 +31,7 @@ enum Method {
 // TODO(folming): switch to RapidJSON.
 using PostFields = std::vector<std::pair<std::string, std::string>>;
 using Header = std::pair<std::string, std::string>;
-using Headers = std::vector<Header>;
+using Headers = std::map<std::string, std::string>;
 
 ////////////////////////////////////////////////////////////////////////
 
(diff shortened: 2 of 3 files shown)


eventuals/builder.h
```diff
@@ -117,6 +123,12 @@ class FieldWithDefault<Value_, true> : public Field<Value_, true> {
     : Field<Value_, true>(std::forward<Value>(value)) {}
 
   FieldWithDefault(FieldWithDefault&&) = default;
+
+  template <typename Value>
+  auto Set(Value&& value) {
+    static_assert(std::is_convertible_v<Value, Value_>);
+    return FieldWithDefault<Value_, true>{std::forward<Value>(value)};
+  }
```
FolMing, line 131: Apparently, we could not use multiple Request::_Builder::header() before, so I had to do this change. I did the same for `Field<Value_, true>` as well for consistency.

eventuals/http.h
```diff
@@ -547,9 +566,91 @@ struct _HTTP {
...
+                          // Cookie: cookie1=value1, cookie2=value2
+                          auto iterator = headers.find(key);
+                          if (iterator == headers.end()) {
+                            // Header doesn't exist yet.
+                            headers.emplace(std::move(key), std::move(value));
+                          } else {
+                            // Header already exists.
+                            headers[key] += ", ";
+                            headers[key] += std::move(value);
+                          }
```
FolMing, line 648: This one is large, but should be self-explanatory.

test/http.cc
```diff
@@ -52,7 +52,10 @@ TEST_P(HttpTest, Get) {
   auto response = future.get();
 
   EXPECT_EQ(200, response.code());
-  EXPECT_THAT(response.headers(), testing::ContainsRegex("Foo: Bar"));
+  EXPECT_THAT(
+      response.headers(),
+      testing::Contains(http::Header("Foo", "Bar")));
+  EXPECT_EQ(response.headers().at("Content-Length"), "25");
```
FolMing, line 58: Even if we are using std::map, we can't use operator[] as response.headers() gives back a constant reference!

eventuals/http.h
```diff
@@ -180,7 +181,25 @@ class Request::_Builder : public builder::Builder {
...
+    // A sender MUST NOT generate multiple header fields with the same field
+    // name in a message unless either the entire field value for that
+    // header field is defined as a comma-separated list [i.e., #(values)]
+    // or the header field is a well-known exception (as noted below).
+    //
+    // NOTE: If user tries to add an already existing header,
+    // append the new one to the old one using comma.
+    // Example:
+    // Cookie: cookie1=value1, cookie2=value2
+    auto iterator = headers_.value().find(key);
```
FolMing, line 194: Would have used std::map::contains(), but it's a C++20 method.

FolMing (commented):
A comment to help you guys understand the changes!

eventuals/http.h
```diff
@@ -547,9 +566,91 @@ struct _HTTP {
...
+                        // HTTP/1.1 200
+                        // SomeHeaderKey1: SomeHeaderValue1
+                        // SomeHeaderKey2: SomeHeaderValue2
+                        // --------------------------------
+                        while (!headers_buffer_stringstream.eof()) {
+                          std::string line;
+                          std::getline(headers_buffer_stringstream, line);
+
+                          // Find where ':' is.
+                          auto column_pos = std::find(
```
benh, line 586: What's a `pos`? 🤪 Keep me honest, I'm trying to be use complete words too but sometimes I forget!

eventuals/http.h
```diff
@@ -547,9 +566,91 @@ struct _HTTP {
...
+                          if (column_pos == line.cend()) {
+                            continue;
+                          }
+
+                          // Assign key and value.
+                          auto key = std::string(line.cbegin(), column_pos);
+                          auto value = std::string(column_pos + 1, line.cend());
+
+                          // Helper lambda for removing leading
+                          // and trailing spaces.
```
benh, line 601: Let's add a `TODO` here to bring in the [`absl`](https://abseil.io) library which does this all for us. We bring it in for `stout-flags` [here](https://github.com/3rdparty/stout-flags/blob/main/bazel/deps.bzl#L14).

test/http.cc
```diff
@@ -231,6 +234,56 @@ TEST_P(HttpTest, GetHeaders) {
...
```
benh, line 240: Love seeing usage of `testing::Contains()` above, let's also do that here!

test/http.cc
```diff
@@ -231,6 +234,56 @@ TEST_P(HttpTest, GetHeaders) {
...
+
+  EventLoop::Default().RunUntil(future);
+
+  auto response = future.get();
+
+  EXPECT_EQ(200, response.code());
+  EXPECT_THAT(
+      response.headers(),
+      testing::Contains(http::Header("Foo", "Bar1, Bar2")));
+  EXPECT_EQ(response.headers().at("Content-Length"), "25");
```
benh, line 287: Ditto using `testing::Contains()` here.

test/http.cc
benh: Makes sense, but why not use `testing::Contains()` like you do on the previous line? It's safer and the preferred method of doing it.

test/http.cc
```diff
@@ -231,6 +234,56 @@ TEST_P(HttpTest, GetHeaders) {
...
+  http::Client client = server.Client();
+
+  EXPECT_CALL(server, ReceivedHeaders)
+      .WillOnce([](auto socket, const std::string& data) {
+        EXPECT_THAT(data, testing::ContainsRegex("foo: bar1, bar2"));
+
+        socket->Send(
+            "HTTP/1.1 200 OK\r\n"
+            "Foo: Bar1\r\n"
+            "Foo: Bar2\r\n"
```
benh, line 260: Based on the RFC that you commented on above, doesn't this violate the rule? Clearly curl handles it given this test passes, but is that what you're trying to test? Or were you just trying to test multiple header values?

benh (changes requested):
Most of this looks great! Let me know if you'd like to discuss the `Repeated` comments synchronously, otherwise for now I'm happy to just have a builder field of type `Headers` instead of wrapping it in a `builder::FieldWithDefault`.

test/http.cc
FolMing: @benh it's a showcase for other people, who will use our library.

test/http.cc
FolMing: Again, just a showcase for others!

test/http.cc
FolMing: ^

test/http.cc
FolMing: From what I read I understand that server "can" do this and we, as a client, "must not" do this.
Take a look: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

eventuals/builder.h
```diff
@@ -29,6 +29,12 @@ class Field<Value_, true> {
   Field(Value_ value)
     : value_(std::move(value)) {}
 
+  template <typename Value>
+  auto Set(Value&& value) {
+    static_assert(std::is_convertible_v<Value, Value_>);
+    return Field<Value_, true>{std::forward<Value>(value)};
+  }
+
```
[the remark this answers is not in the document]
FolMing: `ComboField`? :D

eventuals/builder.h
FolMing: I personally think going with just `Headers` type is the way to handle such cases, dunno if we really want to wrap things up.

rjhuijsman (commented):
I like the API of having headers be an `std::map` a lot! 

I don't know this code well enough to give more than that surface-level review though, so I'll leave the real reviewing to Ben. 🙂

benh (approved):
Thanks Nik!

PR #202, opened 2022-01-25
Optimize away heap allocation from `Task` synchronous helpers

onelxj:


CodingCanuck:
Is this a redo of https://github.com/3rdparty/eventuals/pull/181 ? I take it there's no easy way to directly diff this PR against the old one.

Could you please call out the main differences / fixes from the previous PR in this PR's description? I'd find it helpful to understand how this differs from the first attempt.

eventuals/task.h
```diff
@@ -166,18 +167,59 @@ struct _TaskFromToWith {
...
+                Callback<>&&,
+                Callback<To_>&&>,
+            Callback<std::exception_ptr>&&,
+            Callback<>&&>;
+
     template <typename... From>
     void Start(From&&... from) {
-      if constexpr (std::is_void_v<From_>) {
-        Dispatch(Action::Start, std::monostate{});
+      if (dispatch_.index() == 0) {
```
CodingCanuck, line 203: Please use a switch statement on dispatch_.index().

It would help to clarify that the else case is when index == 2.

onelxj:
Yes, it is, but we have reverted #181 because it hasn't got the support of references.

There is an introduction of `std::reference_wrapper` for `Task::Success` cases and also now it compiles this:
```
auto e = []() -> Task::Of<const int&>{
   return [](){
      return Just(10);
   };
}
```

test/task.cc
```diff
@@ -392,3 +392,102 @@ TEST(Task, Failure) {
...
+  };
+
+  auto [future, k] = Terminate(e());
+
+  k.Start();
+
+  x = 42;
+
+  EXPECT_EQ(42, future.get());
+}
```
CodingCanuck, line 493: Please add a newline at the end of this file.

CodingCanuck (approved):

test/task.cc
FolMing: This should definitely be tested by our check_code_style workflow! Thanks @CodingCanuck!

eventuals/task.h
```diff
@@ -510,6 +595,13 @@ struct Task {
       template <typename... Args_>
       using With = TaskFromToWith<From_, To_, Args_...>;
 
+      To(
+          std::conditional_t<
+              std::is_void_v<To_>,
+              std::monostate,
+              To_> value)
```
benh, line 602: Ok, I'm seeing this one enough that I think we give it a name at the top-level:

```cpp
template <typename T>
using MonostateIfVoidOr = std::conditional_t< ...>;
```

Then:

```suggestion
      To(MonstateIfVoidOr<To_> value)
```

eventuals/task.h
```diff
@@ -389,12 +460,26 @@ struct _TaskFromToWith {
...
 
+  TaskFromToWith(
+      std::conditional_t<
+          std::is_void_v<To_>,
+          std::monostate,
+          std::conditional_t<
+              !std::is_reference_v<To_>,
+              To_,
+              std::reference_wrapper<
+                  std::remove_reference_t<To_>>>> value)
```
benh, line 480: Seeing this one enough (and it's big enough) so let's give it a top-level name:

```cpp
template <typename T>
using MonostateIfVoidOrReferenceWrapperOr = std::conditional_t< ...>;
```

Then:

```suggestion
  TaskFromToWith(MonstateIfVoidOrReferenceWrapperOr<To_> value)
```

eventuals/task.h
```diff
@@ -238,41 +282,71 @@ struct _TaskFromToWith {
...
     Interrupt* interrupt_ = nullptr;
   };
 
-  template <typename From_, typename To_, typename... Args_>
+  template <
+      typename From_,
+      typename To_,
+      typename... Args_>
   struct Composable {
+    using Dispatch_ =
```
benh, line 309: Let's actually move this into `struct _TaskFromToWith` and do:

```cpp
template <typename From, typename To>
using DispatchCallback = Callback<...>;
```

Then we can use `DispatchCallback` in both `struct Composable` and `struct Continuation`.

eventuals/task.h
```diff
@@ -166,18 +168,58 @@ struct _TaskFromToWith {
...
   };
 
-  template <typename K_, typename From_, typename To_, typename... Args_>
+  template <
+      typename K_,
+      typename From_,
+      typename To_,
+      typename... Args_>
   struct Continuation {
+    using Dispatch_ =
```
benh, line 177: See comment below for replace this so it's not duplicated.

eventuals/task.h
```diff
@@ -357,28 +431,25 @@ struct _TaskFromToWith {
...
-    Callback<
-        Action,
-        std::optional<std::exception_ptr>&&,
-        Args_&&...,
-        // Can't have a 'void' argument type so we are using 'std::monostate'.
-        std::optional<
+    // See comment in `Continuation` for explanation of `dispatch_` member.
+    // Using 'std::optional' because of implicitly deleted 'std::variant'
+    // constructor.
+    std::optional<
```
benh, line 440: All of this now becomes:

```cpp
    std::optional<
        std::variant<
            MonostateIfVoidOrReferenceWrapperOr<To_>,
            DispatchCallback<From_, To_>>> value_or_dispatch_;
```

eventuals/task.h
```diff
@@ -238,41 +282,71 @@ struct _TaskFromToWith {
...
+    Composable(
+        std::conditional_t<
+            std::is_void_v<To_>,
+            std::monostate,
+            std::conditional_t<
+                !std::is_reference_v<To_>,
+                To_,
+                std::reference_wrapper<
+                    std::remove_reference_t<To_>>>> value)
+      : dispatch_(std::move(value)) {}
```
benh, line 342:
```suggestion
    Composable(MonostateIfVoidOrReferenceWrapperOr<To_> value)
      : value_or_dispatch_(std::move(value)) {}
```

And let's rename `dispatch_` to `value_or_dispatch_` to make it even more clear.

benh (approved):
Alright!

The reference wrapper definitely made the types more dense so let's factor out some more `using`, see the comments. I might not have marked all the spots but please find them all and use the `using` names.

Thank you for the perseverance on this one!

eventuals/task.h
onelxj: There should be `Args` also.

PR #211, opened 2022-02-02
Make Context non-abstract class and make Concurrent use different schedulers

onelxj:


onelxj:
I'll update the commit name and all other stuff before merging, no worries.

eventuals/scheduler.h
```diff
@@ -71,21 +103,44 @@ class Scheduler {
...
+    Context Clone() {
+      Context child(scheduler_, "cloned", nullptr, this);
+
+      child.scheduler()->Clone(child, *this);
+
+      return child;
     }
 
+    // There is the most common set of variables to create contexts.
+    bool blocked_ = false;
```
benh, line 115: Let's actually change this up a bit. Let's make `blocked_` be private and have a "setter" called `block()` that sets `blocked_` to `true` and a getter called `blocked()` that returns whether or not it's blocked.

eventuals/scheduler.h
```diff
@@ -71,21 +103,44 @@ class Scheduler {
...
+    void* data_ = nullptr;
+
+    // Storing a pointer to 'Context' from what we just cloned.
+    Context* parent_ = nullptr;
+
+    // There are a lot of variants where we need to store contexts
+    // like a list.
+    Context* next_ = nullptr;
+
+    Callback<> callback_;
```
benh, line 127:
```suggestion
    // Schedulers can use 'callback' to store the function that "starts" or "unblocks"/"resumes" the context. 
    Callback<> callback;
```

eventuals/static-thread-pool.cc
```diff
@@ -167,6 +165,13 @@ bool StaticThreadPool::Continuable(Context* context) {
 
 ////////////////////////////////////////////////////////////////////////
 
+void StaticThreadPool::Clone(Context& child, Context& parent) {
+  // Storing additional requirements pinned core.
+  child.data_ = Scheduler::Context::Get()->data_;
```
benh, line 170: I understand that this is the same thing but:
```suggestion
  child->data = parent->data;
```
Thinking more about this, why don't we drop the `parent` argument to `Clone()` altogether and instead we can do:
```suggestion
  child->data = child->parent()->data;
```

And please leave a comment why it's ok to reuse the parent's data pointer without allocating a new one.

eventuals/static-thread-pool.cc
```diff
@@ -151,14 +150,13 @@ void StaticThreadPool::Submit(Callback<> callback, Context* context) {
...
-  auto* waiter = static_cast<Waiter*>(CHECK_NOTNULL(context));
-
-  CHECK(!waiter->waiting) << waiter->name();
-  CHECK(waiter->next == nullptr) << waiter->name();
+  CHECK(!context->blocked_) << context->name();
+  CHECK(context->next_ == nullptr) << context->name();
 
-  auto& pinned = waiter->requirements()->pinned;
+  auto& pinned =
+      static_cast<StaticThreadPool::Requirements*>(context->data_)->pinned;
```
benh, line 157: Let's be more explicit here and everywhere else (in other files too):
```suggestion
  auto* requirements = static_cast<StaticThreadPool::Requirements*>(context->data);
  auto& pinned = requirements->pinned;
```

test/static-thread-pool.cc
```diff
@@ -151,3 +163,27 @@ TEST(StaticThreadPoolTest, Spawn) {
...
+    return StaticThreadPool::Spawn(
+        "concurrent test",
+        Iterate(digits)
+            | ConcurrentOrdered([&]() {
+                auto* context = Scheduler::Context::Get();
+                return Map(Let([&](int& i) {
+                  EXPECT_NE(id, std::this_thread::get_id());
+
+                  // Should be executed on a cloned StaticThreadPool context.
+                  EXPECT_NE(context, Scheduler::Context::Get());
```
benh, line 181:
```suggestion
                  EXPECT_NE(context, Scheduler::Context::Get());
                  EXPECT_EQ(context, Scheduler::Context::Get()->parent());
```

test/static-thread-pool.cc
```diff
@@ -151,3 +163,27 @@ TEST(StaticThreadPoolTest, Spawn) {
...
+
+  auto id = std::this_thread::get_id();
+  auto e = [&]() {
+    return StaticThreadPool::Spawn(
+        "concurrent test",
+        Iterate(digits)
+            | ConcurrentOrdered([&]() {
+                auto* context = Scheduler::Context::Get();
+                return Map(Let([&](int& i) {
+                  EXPECT_NE(id, std::this_thread::get_id());
```
benh, line 178: This line isn't really testing anything related to what you've introduced in this PR because you haven't specified any `requirements`. A better thing to test would be to have a requirement that you are pinned to a specific thread and then make sure that both the parent and child context are on the same thread!

eventuals/scheduler.h
```diff
@@ -37,12 +37,44 @@ class Scheduler {
...
-    Context(Context&& that) = delete;
+    const std::string& name() {
+      return name_;
+    }
+
+    Context(
+        Scheduler* scheduler,
+        const std::string& name = "default",
+        void* data = nullptr,
+        Context* parent = nullptr)
```
benh, line 52: We shouldn't let people set their parent as that could lead to bugs!
```suggestion
```

eventuals/scheduler.h
```diff
@@ -37,12 +37,44 @@ class Scheduler {
...
+      : data_(data),
+        parent_(parent),
```
benh, line 54:
```suggestion
        parent_(Scheduler::Context::Get()),
```

eventuals/scheduler.h
```diff
@@ -37,12 +37,44 @@ class Scheduler {
...
+        scheduler_(scheduler),
+        name_(name) {}
+
+    Context(Context&& that)
+      : data_(that.data_),
+        parent_(that.parent_),
+        scheduler_(that.scheduler()),
+        name_(that.name()) {
+      // NOTE: should only get moved before it's "started".
+      CHECK(!that.blocked_ && !callback_ && next_ == nullptr);
```
benh, line 64: Glad that you caught this!

Let's actually not support move construction at all! In fact, let's set the move constructor to `= delete;`. Then let's remove the `Scheduler::Context::Clone()` and instead have the `Context()` constructor be sufficient for doing a clone and inside the constructor call `scheduler_->Clone(this);`.

eventuals/scheduler.h
```diff
@@ -71,21 +103,44 @@ class Scheduler {
       }
     }
 
-    auto* scheduler() {
-      return scheduler_;
+    Context Clone() {
```
benh, line 106: So let's delete this ...

eventuals/scheduler.h
```diff
@@ -37,12 +37,44 @@ class Scheduler {
...
+
+    Context(
+        Scheduler* scheduler,
+        const std::string& name = "default",
+        void* data = nullptr,
+        Context* parent = nullptr)
+      : data_(data),
+        parent_(parent),
+        scheduler_(scheduler),
+        name_(name) {}
```
benh, line 56: See other comments, but in here let's do:
```suggestion
        name_(name) {
          scheduler_->Clone(this);
        }
```

eventuals/concurrent.h
```diff
@@ -477,14 +477,16 @@ struct _Concurrent {
       static_cast<Fiber<E>*>(fiber)->k.emplace(
           Build(FiberEventual(fiber, std::move(arg))));
 
-      // NOTE: we need to "submit" using the default scheduler so that
-      // each fiber will properly be using it's own scheduler context.
-      Scheduler::Default()->Submit(
+      auto child = fiber->Clone();
+      fiber->child = std::move(child);
```
benh, line 481: See other comments, but this should become:
```suggestion
      // TODO(benh): differentiate the names of the fibers for
      // easier debugging!
      fiber->child.emplace("[concurrent fiber]");
```

eventuals/concurrent.h
```diff
@@ -98,6 +95,9 @@ struct _Concurrent {
 
       // Each fiber forms a linked list of currently created fibers.
       std::unique_ptr<TypeErasedFiber> next;
+
+      // Need to store a cloned context in which would be stored callback.
+      std::optional<Context> child;
```
benh, line 100: Since we don't have a child/parent variable here let's do:
```suggestion
      std::optional<Context> context;
```

eventuals/concurrent.h
```diff
@@ -71,14 +71,11 @@ struct _Concurrent {
...
-        // TODO(benh): differentiate the names of the fibers for
-        // easier debugging!
-        static std::string name = "[concurrent fiber]";
-        return name;
-      }
+        : Scheduler::Context(
+            Scheduler::Context::Get()->scheduler(),
+            // TODO(benh): differentiate the names of the fibers for
+            // easier debugging!
+            std::string("[concurrent fiber]")) {}
```
benh, line 78: This is dead code! You use the context member in this class instead!

benh (changes requested):
Looking good!

Good catch on the non-moveable context after being started. Because of that, let's actually delete the move constructor. Then I think we need two `Context()` constructors, one which takes a scheduler pointer and the name and one which takes just the name. The second one will be the "clone" constructor and call `Scheduler::Clone()`. Then I know we talked about removing the `parent` argument from the `Clone()` function but thinking more about it I think we actually keep the `parent` and actually remove `parent` from the `Context` altogether ... I don't think you ever need it except when you do the clone!

test/static-thread-pool.cc
benh: Since we're going to remove the parent let's instead check that the current context's scheduler is the same as the outer context's scheduler!

eventuals/static-thread-pool.cc
benh: Ok, final comment here, let's remove the `parent` argument and then we can keep the code just as you had it, but still add the comment!

eventuals/static-thread-pool.h
```diff
@@ -146,26 +118,45 @@ class StaticThreadPool : public Scheduler {
...
+                std::move(name))) {}
+
+    Scheduler::Context* context() {
+      auto* constructed = context_.get();
+      if (!constructed->data) {
+        constructed->data = requirements_;
+      }
+
+      return constructed;
+    }
```
benh, line 143: Let's chat about replacing this with a `Context` constructor that takes 'data'.

eventuals/static-thread-pool.h
```diff
@@ -500,6 +504,16 @@ struct _StaticThreadPoolSchedule {
 
     Interrupt* interrupt_ = nullptr;
 
+    StaticThreadPool::Requirements* requirements_ = nullptr;
+
+    // Need to store context instead of ineritance because
```
benh, line 509:
```suggestion
    // Need to store context instead of using inheritance because
```

eventuals/static-thread-pool.h
```diff
@@ -520,6 +534,7 @@ struct _StaticThreadPoolSchedule {
           std::move(k),
           pool_,
           requirements_,
+          std::move(requirements_->name),
```
benh, line 537: Some folks might want to reuse a requirements, so I think we should properly propagate a name all the call to create the `Composable`.

benh (changes requested):

eventuals/static-thread-pool.cc
```diff
@@ -167,6 +167,15 @@ bool StaticThreadPool::Continuable(Context* context) {
 
 ////////////////////////////////////////////////////////////////////////
 
+void StaticThreadPool::Clone(Context* child) {
+  // Storing additional requirements pinned core.
+  // No need to reallocate parent's 'data' because requirements that
+  // stored there is common for whole StaticThreadPool.
+  child->data = child->parent()->data;
```
benh, line 174:
```suggestion
  child->data = Scheduler::Context::Get()->data;
```

test/static-thread-pool.cc
```diff
@@ -151,3 +163,25 @@ TEST(StaticThreadPoolTest, Spawn) {
...
+                auto* outer_context = Scheduler::Context::Get();
+
+                return Map(Let([&](int& i) {
+                  // Should be executed on a cloned StaticThreadPool context.
+                  EXPECT_NE(outer_context, Scheduler::Context::Get());
+                  return i;
+                }));
+              })
+            | Collect<std::vector<int>>());
+  };
```
benh, line 184:
```suggestion
  StaticThreadPool::Requirements requirements("pinned to 2", Pinned(2));

  auto e = [&]() {
    return StaticThreadPool::Spawn(
        requirements,
        Iterate(digits)
            | ConcurrentOrdered([&]() {
                auto* parent = Scheduler::Context::Get();
                EXPECT_TRUE(requirements == parent.data());
                return Map(Let([&](int& i) {
                  // Should be executed on a cloned StaticThreadPool context.
                  auto* child = Scheduler::Context::Get();
                  EXPECT_NE(parent, child);
                  EXPECT_TRUE(requirements == child->data());
                  return i;
                }));
              })
            | Collect<std::vector<int>>());
  };
```

benh (changes requested):

eventuals/static-thread-pool.h
```diff
@@ -42,7 +63,8 @@ class StaticThreadPool : public Scheduler {
       : Requirements(std::string(name), std::move(pinned)) {}
 
     Requirements(std::string name)
-      : name(std::move(name)) {}
+      : name(std::move(name)),
+        pinned(Pinned::Any()) {}
 
     Requirements(std::string name, Pinned pinned)
```
benh, line 69: Let's delete the constructor that only takes `name` and make the constructor that takes `Pinned` have a default value of `Pinned::Any()`.

eventuals/static-thread-pool.h
```diff
@@ -85,7 +107,7 @@ class StaticThreadPool : public Scheduler {
   static inline thread_local bool member = false;
 
   // If 'member', for which core?
```
benh, line 109: Replace 'core' with 'cpu' please.

eventuals/static-thread-pool.h
```diff
@@ -19,15 +19,36 @@ namespace eventuals {
...
   Pinned(const Pinned& that)
-    : core(that.core) {}
+    : cpu_(that.cpu_) {}
 
-  std::optional<unsigned int> core;
+ private:
+  Pinned(unsigned int cpu)
+    : cpu_(cpu) {}
+
+  unsigned int cpu_;
```
benh, line 51: Sorry if this got missed in discussion, but let's keep this as a `std::optional` and then we can return a `std::optional` from `cpu()` as well.

eventuals/static-thread-pool.h
```diff
@@ -160,48 +181,30 @@ struct _StaticThreadPoolSchedule {
...
-        if (interrupt_ != nullptr) {
-          k_.Register(*interrupt_);
-        }
-        k_.Fail("'" + context_->name() + "' required core is > total cores");
+      if (StaticThreadPool::member && StaticThreadPool::cpu == pinned.CPU()) {
+        Adapt();
+        auto* previous = Scheduler::Context::Switch(context_.get());
+        adapted_->Start(std::forward<Args>(args)...);
+        previous = Scheduler::Context::Switch(previous);
+        CHECK_EQ(previous, context_.get());
```
benh, line 189: Let's revert this because _this_ is when we want to pick the least loaded core. NOT when we're creating a `Pinned`.

benh (changes requested):

eventuals/static-thread-pool.cc
```diff
@@ -158,19 +154,18 @@ bool StaticThreadPool::Continuable(Context* context) {
       static_cast<StaticThreadPool::Requirements*>(context->data);
   auto& pinned = requirements->pinned;
 
-  CHECK(pinned.core) << context->name();
-
-  unsigned int core = pinned.core.value();
+  unsigned int cpu = pinned.cpu().value();
```
benh, line 157: Please put the `CHECK(pinned.cpu()) << context->name();` back!

eventuals/static-thread-pool.h
```diff
@@ -19,32 +19,45 @@ namespace eventuals {
...
 
-  Pinned(unsigned int core)
-    : core(core) {}
+  static Pinned Any() {
+    return Pinned();
+  }
+
+  static Pinned ExactCPU(unsigned int cpu) {
+    CHECK(cpu < std::thread::hardware_concurrency())
+        << "specified CPU can't be reached";
```
benh, line 32: Instead of `can't be reached` let's say `is not valid`.

eventuals/static-thread-pool.h
```diff
@@ -139,12 +151,14 @@ struct _StaticThreadPoolSchedule {
...
 
-    // To avoid casting default 'Scheduler' to 'StaticTheadPool' each time.
+    // Helper to avoid casting default 'Scheduler' to 'StaticTheadPool'
+    // each time.
     auto* pool() {
       return static_cast<StaticThreadPool*>(context_->scheduler());
     }
 
-    // To avoid casting 'void*' to 'StaticThreadPool::Requirements*' each time.
+    // Helper to avoid casting default 'Scheduler' to 'StaticTheadPool'
```
benh, line 160:
```suggestion
    // Helper to avoid casting 'void*' to 'StaticThreadPool::Requirements*'
```

eventuals/static-thread-pool.h
```diff
@@ -139,12 +151,14 @@ struct _StaticThreadPoolSchedule {
                 std::move(name),
                 CHECK_NOTNULL(requirements))) {}
 
-    // To avoid casting default 'Scheduler' to 'StaticTheadPool' each time.
+    // Helper to avoid casting default 'Scheduler' to 'StaticTheadPool'
```
benh, line 154:
```suggestion
    // Helper to avoid casting 'Scheduler*' to 'StaticThreadPool*'
```

eventuals/static-thread-pool.h
```diff
@@ -160,48 +174,38 @@ struct _StaticThreadPoolSchedule {
...
 
-      assert(pinned.core);
-
-      if (!(*pinned.core < pool()->concurrency)) {
-        CHECK(!adapted_);
-        if (interrupt_ != nullptr) {
-          k_.Register(*interrupt_);
-        }
-        k_.Fail("'" + context_->name() + "' required core is > total cores");
+      if (StaticThreadPool::member && StaticThreadPool::cpu == pinned.cpu()) {
```
benh, line 185:
```suggestion
      CHECK(pinned.cpu() <= pool()->concurrency);

      if (StaticThreadPool::member && StaticThreadPool::cpu == pinned.cpu()) {
```

eventuals/static-thread-pool.h
```diff
@@ -216,53 +220,43 @@ struct _StaticThreadPoolSchedule {
...
-      assert(pinned.core);
-
-      if (!(*pinned.core < pool()->concurrency)) {
-        CHECK(!adapted_);
-        if (interrupt_ != nullptr) {
-          k_.Register(*interrupt_);
-        }
-        k_.Fail("'" + context_->name() + "' required core is > total cores");
+      if (StaticThreadPool::member
+          && StaticThreadPool::cpu == pinned.cpu()) {
```
benh, line 232:
```suggestion
      CHECK(pinned.cpu() <= pool()->concurrency);

      if (StaticThreadPool::member && StaticThreadPool::cpu == pinned.cpu()) {
```

eventuals/static-thread-pool.h
```diff
@@ -276,41 +270,31 @@ struct _StaticThreadPoolSchedule {
...
```
benh, line 282:
```suggestion
      CHECK(pinned.cpu() <= pool()->concurrency);

      if (StaticThreadPool::member && StaticThreadPool::cpu == pinned.cpu()) {
```

eventuals/static-thread-pool.h
```diff
@@ -322,41 +306,31 @@ struct _StaticThreadPoolSchedule {
...
```
benh, line 318:
```suggestion
      CHECK(pinned.cpu() <= pool()->concurrency);

      if (StaticThreadPool::member && StaticThreadPool::cpu == pinned.cpu()) {
```

eventuals/static-thread-pool.h
```diff
@@ -369,49 +343,40 @@ struct _StaticThreadPoolSchedule {
...
```
benh, line 356:
```suggestion
      CHECK(pinned.cpu() <= pool()->concurrency);

      if (StaticThreadPool::member && StaticThreadPool::cpu == pinned.cpu()) {
```

eventuals/static-thread-pool.h
```diff
@@ -425,41 +390,31 @@ struct _StaticThreadPoolSchedule {
...
```
benh, line 402:
```suggestion
      CHECK(pinned.cpu() <= pool()->concurrency);

      if (StaticThreadPool::member && StaticThreadPool::cpu == pinned.cpu()) {
```

eventuals/event-loop.h
```diff
@@ -716,12 +691,19 @@ class EventLoop : public Scheduler {
...
-        k_(std::move(k)),
-        e_(std::move(e)) {}
+      : k_(std::move(k)),
+        e_(std::move(e)),
+        context_(
+            std::make_tuple(
+                CHECK_NOTNULL(loop),
+                std::move(name))) {}
+
+    // To avoid casting default 'Scheduler' to 'EventLoop' each time.
```
benh, line 703:
```suggestion
    // To avoid casting 'Scheduler*' to 'EventLoop*' each time.
```

benh (approved):
💪

PR #212, opened 2022-02-03
Rename '.then()' and '.otherwise()' for 'If()' to '.yes()' and '.no()'

benh:
@while-false as discussed!


onelxj:
What do you think about 'True' 'False'?

while-false (approved):
LGTM! 

In response to @onelxj : `True`/`False` were also discussed as candidates - so good instinct :+1:. The capitalization of True and False, however, breaks with the style used other-where for builder patterns.

PR #213, opened 2022-02-04
Set 'experimental_ui_max_stdouterr_bytes'

benh:
Need to set this to -1 so that we get output from GitHub Workflows.


FolMing (approved):

benh:
Thank you @FolMing!

PR #214, opened 2022-02-10
Make 'Catch' use builder

onelxj:


eventuals/catch.h
```diff
@@ -12,22 +13,38 @@ namespace eventuals {
...
+    template <typename F, typename... Args>
+    void BuildAndInvoke(F&& f, Args&&... args) {
+      using E = decltype(f(std::forward<Args>(args)...)
                              .template k<void>(_Then::Adaptor<K_>{k_}));
 
       // TODO(benh): propagate eventual errors so we don't need to
       // allocate on the heap in order to type erase.
       e_ = std::unique_ptr<void, Callback<void*>>(
-          new E(f_(std::forward<Args>(args)...)
+          new E(f(std::forward<Args>(args)...)
```
benh, line 47: We actually now have enough information that we don't need to do dynamic memory allocation _except_ in the all case! Let's chat synchronously about what this can look like.

eventuals/catch.h
```diff
@@ -42,6 +59,85 @@ struct _Catch {
...
+        typename CatchHandler_,
+        typename... Tail>
+    void UnpackInvokeHelper(
+        Error_&& error,
+        CatchHandler_&& handler,
+        Tail&&... tail) {
+      if constexpr (
+          IsSame<
+              std::decay_t<CatchHandler_>,
+              std::decay_t<Error_>>::value) {
```
benh, line 73: Let's do something like:
```suggestion
      if constexpr (Handler::Handles<Error>) {
```

eventuals/catch.h
```diff
@@ -42,6 +59,85 @@ struct _Catch {
       e->Start();
     }
 
+    template <
+        typename Error_,
+        typename CatchHandler_,
+        typename... Tail>
+    void UnpackInvokeHelper(
```
benh, line 66: We need to handle the case that `Error` is actually a `std::exception_ptr` and introduce an explicit `try` that catches the type we want. Shall we chat about this synchronously?

benh (changes requested):

test/catch.cc
```diff
@@ -0,0 +1,215 @@
...
+}
+
+TEST(CatchTest, All) {
+  auto checker = [](auto error) {
+    EXPECT_EQ(error, 10);
+  };
+
+  auto e = [&]() {
+    return Just(std::string("message"))
+        | Raise(10)
```
benh, line 75: If this line was removed then this code should no longer compile! That's a bug, your `std::string("message")` needs to be what the `Then` would get below if we don't do a `Raise`.

test/catch.cc
```diff
@@ -0,0 +1,215 @@
...
+    return expected() // Throwing 'std::exception_ptr' there.
+        | Catch()
+              .raised<int>([](auto&& error) {
+                ADD_FAILURE() << "Encountered an unexpected raised 'int'";
+                return 1;
+              })
+              // Receive 'Error' type there, that had been rethrowed from
+              // 'std::exception_ptr'.
+              .raised<Error>([](auto&& error) {
+                EXPECT_EQ(std::string("child exception"), error.what());
```
benh, line 121:
```suggestion
                EXPECT_STR_EQ("child exception", error.what());
```

test/catch.cc
```diff
@@ -0,0 +1,215 @@
...
+    return Just(1)
+        | Raise(1)
+        | Catch()
+              .raised<int>([](auto&& error) {
+                return Raise(1);
+              })
+              .raised<std::exception_ptr>([](auto&& error) {
+                ADD_FAILURE()
+                    << "Encountered an unexpected raised 'std::exception_ptr'";
+                return Raise(100);
```
benh, line 201: To really make sure that this test is doing what we want let's make one of the handlers _not_ do a `Raise()` but instead return just an `int`. I think that will mean that we actually need to do _more_ type unification to deal with the case that we have 'Raise()`. See `eventuals/conditional.h` for how to `Unify`!

test/catch.cc
```diff
@@ -0,0 +1,215 @@
...
+}
+
+TEST(CatchTest, ChildException) {
+  struct Error : public std::exception {
+    const char* what() const noexcept override {
+      return "child exception";
+    }
+  };
+
+  auto e = [&]() {
```
benh, line 48:
```suggestion
  auto e = []() {
```

benh, line 48:
```suggestion
  auto e = []() {
```

test/catch.cc
```diff
@@ -0,0 +1,215 @@
...
+
+  EXPECT_EQ(*e(), 100);
+}
+
+TEST(CatchTest, All) {
+  auto checker = [](auto error) {
+    EXPECT_EQ(error, 10);
+  };
+
+  auto e = [&]() {
```
benh, line 73: This is exactly why we want to have `Fail(Error&& error)` functions because then we don't have to jump through hoops like this because we have a `...`. It would be much better to just check inline, it's less complex!
```suggestion
  auto e = []() {
```

benh (changes requested):

benh:
This is coming along @onelxj! It was trickier than I realized!

Definitely check out `eventuals/conditional.h` for how we do type unification which you'll need to do to support a catch handler doing a re-raise. We can also talk more about this synchronously.

benh:
This is coming along @onelxj! It was trickier than I realized!

Definitely check out `eventuals/conditional.h` for how we do type unification which you'll need to do to support a catch handler doing a re-raise. We can also talk more about this synchronously.

benh:
@onelxj are you blocked on this? If not, would love to get this wrapped up!

eventuals/catch.h
```diff
@@ -12,38 +15,113 @@ namespace eventuals {
...
+        if constexpr (!std::is_same_v<Error_, std::exception_ptr>) {
+          ConstructAndStart(std::move(k), interrupt, std::forward<E>(e));
+        } else {
+          ConstructAndStart(
+              std::move(k),
+              interrupt,
+              make_exception_ptr_or_forward(std::forward<E>(e)));
+        }
+
+        return true;
```
benh, line 64: I'd still prefer to explicitly call out the first case related to the comment above, otherwise the comment gets lost and it's hard to discern how the "all" handler works:
```suggestion
      // When 'Error_' is 'std::exception_ptr' it indicates we're the
      // 'all' handler which catches everything but we need to pass a
      // 'std::exception_ptr' in case we don't already have one.
      if constexpr (std::is_same_v<Error_, std::exception_ptr>) {
        ConstructAndStart(
            std::move(k),
            interrupt,
            make_exception_ptr_or_forward(std::forward<E>(e)));
        
        return true;
      } else if constexpr (std::disjunction_v<
                        std::is_same<Error_, E>,
                        std::is_base_of<Error_, E>>) {
        ConstructAndStart(std::move(k), interrupt, std::forward<E>(e));
        return true;
```

eventuals/catch.h
```diff
@@ -12,38 +15,113 @@ namespace eventuals {
...
+      } else if constexpr (std::is_same_v<E, std::exception_ptr>) {
+        try {
+          std::rethrow_exception(e);
+        }
+        // Catch by reference because if 'Error_' is a polymorphic
+        // type it can't be caught by value.
+        catch (Error_& error) {
+          ConstructAndStart(std::move(k), interrupt, std::move(error));
```
benh, line 72:
```suggestion
          ConstructAndStart(std::move(k), interrupt, std::forward<E>(error));
```

eventuals/catch.h
```diff
@@ -12,38 +15,113 @@ namespace eventuals {
...
+
+      adapted_->Start(std::forward<E>(e));
+    }
+
+    template <typename E>
+    // NOTE: we're _must_ explicitly take 'k' and 'e' by rvalue reference here
+    // because we do _not_ want the compiler to move the 'k' and 'e' when
+    // calling this function in the event that we _don't_ end up handling the
+    // error.
+    bool Handle(K_&& k, Interrupt* interrupt, E&& e) {
```
benh, line 47: How about:
```suggestion
    bool TryHandle(K_&& k, Interrupt* interrupt, E&& e) {
```

eventuals/catch.h
```diff
@@ -12,38 +15,113 @@ namespace eventuals {
...
+      : f_(std::move(f)) {}
+
+    // Helper to convert a catch handler to a new 'K'.
+    template <typename K>
+    auto Convert() && {
+      return Handler<K, Error_, F_>{std::move(f_)};
+    }
+
+    template <typename E>
+    void ConstructAndStart(K_&& k, Interrupt* interrupt, E&& e) {
```
benh, line 32: How about:
```suggestion
    void Handle(K_&& k, Interrupt* interrupt, E&& e) {
```

eventuals/catch.h
```diff
@@ -12,38 +15,113 @@ namespace eventuals {
...
-        e->Register(*interrupt_);
-      }
+    template <typename Error>
+    void Fail(Error&& error) {
+      std::apply(
+          [&](auto&... catch_handler) {
+            // Using a fold expression to simplify the iteration.
+            ([&](auto& catch_handler) {
+              if (!handled_) {
+                handled_ = catch_handler.Handle(
```
benh, line 112:
```suggestion
                handled_ = catch_handler.TryHandle(
```

test/catch.cc
```diff
@@ -0,0 +1,232 @@
...
+  EXPECT_THROW(*e(), std::runtime_error);
+}
+
+TEST(CatchTest, ReRaise) {
+  auto e = []() {
+    return Just(1)
+        | Raise("10")
+        | Catch()
+              .raised<std::runtime_error>([](auto&& error) {
+                EXPECT_EQ(error.what(), std::string("10"));
```
benh, line 193:
```suggestion
                EXPECT_STREQ(error.what(), "10");
```

test/catch.cc
```diff
@@ -0,0 +1,232 @@
...
+              .all([](std::exception_ptr&& error) {
+                ADD_FAILURE() << "Encountered an unexpected all";
+                return Just(100);
+              })
+        | Then([](auto&&) {
+             return 200;
+           })
+        | Catch()
+              .raised<std::runtime_error>([](auto&& error) {
+                EXPECT_EQ(error.what(), std::string("1"));
```
benh, line 205:
```suggestion
                EXPECT_STREQ(error.what(), "1");
```

eventuals/catch.h
```diff
@@ -12,38 +15,113 @@ namespace eventuals {
...
-          });
-
-      auto* e = static_cast<E*>(e_.get());
-
-      if (interrupt_ != nullptr) {
-        e->Register(*interrupt_);
-      }
+    template <typename Error>
+    void Fail(Error&& error) {
+      std::apply(
```
benh, line 107:
```suggestion
    void Fail(Error&& error) {
      static_assert(
          std::disjunction_v<
              std::is_base_of<std::exception, std::decay_t<Error>>,
              std::is_same<std::exception_ptr, std::decay_t<Error>>>,
          "'Catch' expects a type derived from std::exception or a std::exception_ptr");

      std::apply(
```

test/catch.cc
```diff
@@ -0,0 +1,232 @@
...
+using eventuals::Then;
+using eventuals::Unexpected;
+using testing::MockFunction;
+
+TEST(CatchTest, RaisedRuntimeError) {
+  auto e = []() {
+    return Just(1)
+        | Raise(std::runtime_error("message"))
+        | Catch()
+              .raised<int>([](auto&& error) {
```
benh, line 31: Instead of using `int` or `double` or `std::string` or any other types for `raised<...>()` let's instead use things like `std::overflow_error` and `std::underflow_error` and anything that extends from `std::exception` as found [here](https://en.cppreference.com/w/cpp/error/exception).

To simplify let's change all the `ADD_FAILURE()` strings to just say `"Encountered unexpected matched raised"`.

Here and the rest of the file please!

benh (approved):
🕺🏼

PR #215, opened 2022-02-11
Fix bug in condition variable's waiters

while-false:
Before this commit a `Waiter` was added to the pending list of waiters no matter if they were done waiting or not, potentially leading to a use-after-free error.

Bonus commit: The PR contains an extra commit to add a `branch = main` to the `dev-tools` submodule.

### DIFF
--- a/eventuals/lock.h
+++ b/eventuals/lock.h
@@ -694,25 +694,40 @@ class ConditionVariable {
     return eventuals::Wait(
         lock_,
         [this, f = std::move(f), waiter = Waiter()](auto notify) mutable {
-          waiter.notify = std::move(notify);
-          if (head_ == nullptr) {
-            head_ = &waiter;
-          } else if (head_->next == nullptr) {
-            head_->next = &waiter;
-          } else {
-            auto* next = head_->next;
-            while (next->next != nullptr) {
-              next = next->next;
-            }
-            next->next = &waiter;
-          }
-          return [&]() {
-            if constexpr (std::is_same_v<F, EmptyCondition>) {
+          // Check if template `F` is for `EmptyCondition`.
+          constexpr bool using_empty_condition =
+              std::is_same_v<F, EmptyCondition>;
+
+          // Helper to determine if we need to wait or not.
+          auto should_wait = [&]() {
+            if constexpr (using_empty_condition) {
               return f(waiter);
             } else {
               return f();
             }
           };
+
+          // If we should wait, the `waiter` need to be
+          // enqueued with other waiters so it can later be notified.
+          if (should_wait()) {
+            // Assign `nofify` callback to `waiter` for later use.
+            waiter.notify = std::move(notify);
+
+            // Add `waiter` to list of waiters. The below might look convoluted
+            // at first but is a text book "append" to a linked list.
+            if (head_ == nullptr) {
+              head_ = &waiter;
+            } else if (head_->next == nullptr) {
+              head_->next = &waiter;
+            } else {
+              auto* next = head_->next;
+              while (next->next != nullptr) {
+                next = next->next;
+              }
+              next->next = &waiter;
+            }
+          }
+          return should_wait;
         });
   }
 
--- a/test/lock.cc
+++ b/test/lock.cc
@@ -378,3 +378,42 @@ TEST(LockTest, ConditionVariable) {
   future2.get();
   future3.get();
 }
+
+TEST(LockTest, ConditionVariable_UseAfterFree) {
+  // A bug was caught in the wild where `ConditionVariable` would enqueue a
+  // waiting `eventual` for later notification, even if the waiting condition
+  // was already met at the time of creation. This would result in a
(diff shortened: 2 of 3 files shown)


benh (changes requested):
Sweet! Please add a test and then I'll give a review!

These are the fun tests to write because the first time you write them they fail then you fix the code and rerun the test and it works! It's so satisfying!

test/lock.cc
```diff
@@ -347,34 +292,103 @@ TEST(LockTest, ConditionVariable) {
...
+    // Wait for `false`. This should be a no-op and not result in the
+    // `eventual` being enqueued.
+    auto [future, k] = Terminate(foo.WaitFor(false));
+    k.Start();
+
+    EXPECT_NE(std::future_status::timeout,
+              future.wait_for(std::chrono::seconds(0)));
+
+    bool condition_value = future.get();
+    EXPECT_EQ(condition_value, false);
```
benh, line 380: Since you're ultimately _waiting_ here (versus exiting early using one of the `ASSERT_*()` family of functions) then can I suggest you simplify the code:
```suggestion
    EXPECT_EQ(false, *foo.WaitFor(false));
```

test/lock.cc
```diff
@@ -347,34 +292,103 @@ TEST(LockTest, ConditionVariable) {
...
+  // in the condition variable, setting the condition variable to `false`
+  // (again) and attempting to notify the destructed object should make the
+  // test blow up.
+  auto [update_future, update_k] = Terminate(foo.SetCondition(false));
+  update_k.Start(); // In case of error, this blows up.
+
+  // Else, make sure we are not blocked, and terminate test.
+  EXPECT_NE(std::future_status::timeout,
+            update_future.wait_for(std::chrono::seconds(0)));
+  update_future.get();
```
benh, line 393: As commented above, since you're waiting here no matter what, how about just:
```suggestion
  *foo.SetCondition(false);
```

test/lock.cc
```diff
@@ -347,34 +292,103 @@ TEST(LockTest, ConditionVariable) {
...
+        condition_variable_.NotifyAll();
+        return value;
+      }));
+    }
+
+    auto WaitFor(bool value) {
+      return Synchronized(condition_variable_.Wait([this, value]() {
+        return condition_ != value;
+      }) | Just(value));
+    }
```
benh, line 358: I don't understand the value of the extra complexity in checking `condition_`? For the tests sake, why not just call `condition_variable_.Wait()` with a lambda that always returns `false`?

Likewise why not just simplify and rename `SetCondition` to something like `NotifyAll()` and drop the `bool condition_` member all together?

Then the test can then just do `*foo.Wait()` and then `*foo.NotifyAll()`.

There are other code paths here that are definitely not getting tested nor do I think we need/want to test and I'd like to not have to think about those when trying to understand what we are testing. 😉

eventuals/lock.h
```diff
@@ -694,20 +694,41 @@ class ConditionVariable {
...
-          } else if (head_->next == nullptr) {
-            head_->next = &waiter;
+          // Check if template `F` is for `EmptyCondition`.
+          constexpr bool using_empty_condition = std::is_same_v<F, EmptyCondition>;
+
+          // Determine current state of condition variable.
+          // Note: Initializing using an `?:` did not work, so here goes verbatim.
+          bool current_condition;
+          if constexpr (using_empty_condition) {
+            current_condition = f(waiter);
```
benh, line 704: May I suggest here:
```suggestion
          // Helper to determine if we need to wait or not.
          auto wait = [&]() {
            if constexpr (using_empty_condition) {
              return f(waiter);
            } else {
              return f();
            }
          };
```

eventuals/lock.h
```diff
@@ -694,20 +694,41 @@ class ConditionVariable {
...
           } else {
-            auto* next = head_->next;
-            while (next->next != nullptr) {
-              next = next->next;
+            current_condition = f();
+          }
+
+          if (current_condition == true) {
```
benh, line 709: Based on my comment in the code above (which I wasn't able to do correctly because GitHub won't let me make a suggestion on deleted lines 🤦🏻‍♂️), then this becomes:
```suggestion
          if (wait()) {
```

eventuals/lock.h
```diff
@@ -694,20 +694,41 @@ class ConditionVariable {
...
+            } else {
+              auto* next = head_->next;
+              while (next->next != nullptr) {
+                next = next->next;
+              }
+              next->next = &waiter;
             }
-            next->next = &waiter;
           }
           return [&]() {
```
benh, line 730: And now based on comments from above this can become:
```suggestion
          return wait;
```

benh (approved):
Thanks @while-false for fixing my bugs! 🤪

Looks like something went awry with clang-format? Your `.clang-format` should link to the one in `dev-tools`, maybe you have another that is overwriting that one?

Let me know if you'd like to talk about any of my comments before merging.

while-false (commented):
Thanks for you input @benh - simple is better than complex.

PR #221, opened 2022-02-17
Reorder eventuals continuation 'k_' to be last

benh:
By always storing 'k_' as the _last_ member it will be destructed
_first_ and thus we won't have any use-after-delete issues during
destruction of 'k_' if it holds any references or pointers to any (or
within any) of the previous continuations.
    
This fixes 3rdparty/eventuals-grpc#71 and allows us to turn on
building with 'asan' on this repository as well as eventuals-grpc.

### DIFF
--- a/eventuals/catch.h
+++ b/eventuals/catch.h
@@ -14,6 +14,10 @@ namespace eventuals {
 struct _Catch {
   template <typename K_, typename F_>
   struct Continuation {
+    Continuation(K_ k, F_ f)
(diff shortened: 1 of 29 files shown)


onelxj:
What about concurrent-ordered helpers?

benh:
> What about concurrent-ordered helpers?

Thanks!

onelxj (approved):
LGTM!

PR #223, opened 2022-02-20
Log warning to help debug when dereferencing an eventual that fails

benh:
@while-false this was inspired from our debugging session, will possibly help track down issues even before pulling out the debugger.


while-false (approved):
Excellent idea! Thanks @benh!

PR #225, opened 2022-02-21
Add virtual destructors and make classes be `final`

onelxj:


eventuals/concurrent-ordered.h
```diff
@@ -20,12 +20,14 @@ namespace eventuals {
...
 
-struct _ReorderAdaptor {
+struct _ReorderAdaptor final {
   template <typename K_, typename Value_>
-  struct Continuation : public TypeErasedStream {
+  struct Continuation final : public TypeErasedStream {
     Continuation(K_ k)
       : k_(std::move(k)) {}
 
+    Continuation(Continuation&& that) = default;
```
benh, line 29: Did a compiler ask for this?

eventuals/event-loop.h
```diff
@@ -420,7 +420,7 @@ class EventLoop : public Scheduler {
 
   EventLoop();
   EventLoop(const EventLoop&) = delete;
-  virtual ~EventLoop();
+  ~EventLoop();
```
benh, line 423:
```suggestion
  ~EventLoop() override;
```

eventuals/range.h
```diff
@@ -10,7 +10,7 @@ namespace eventuals {
 
 struct _Range {
```
benh, line 11: This one too?

eventuals/scheduler.h
```diff
@@ -53,7 +53,7 @@ class Scheduler {
 
     Context(Context&& that) = delete;
 
-    virtual ~Context() {}
+    virtual ~Context() = default;
```
benh, line 56: Let's actually make `Context` be `final`!

benh (changes requested):
Please add `override` to destructors of derived classes, thanks!

eventuals/concurrent-ordered.h
onelxj: Yep

eventuals/range.h
onelxj: Thanks!

eventuals/scheduler.h
onelxj: Let's chat about that synchronously.

benh, line 56: Doesn't need to be `virtual` because it's `final` and doesn't `override` any base class destructor!

benh (approved):

PR #226, opened 2022-02-21
Don't create nested `std::exception_ptr`

onelxj:


eventuals/terminal.h
```diff
@@ -200,6 +200,20 @@ struct StoppedException : public std::exception {
 
 ////////////////////////////////////////////////////////////////////////
 
+// Using to don't create nested 'std::exception_ptr'.
+template <typename... Args>
+auto make_exception_ptr_or_forward(Args&&... args) {
+  if constexpr (sizeof...(args) > 0) {
+    return std::make_exception_ptr(std::forward<Args>(args)...);
+  }
```
benh, line 208: The fact that you don't return anything in the case when `sizeof...(args) == 0` seems to mean that you always have an arg! So in that case let's instead have a `static_assert()` rather than have the possibility that this gets called with no args and leads to a tricky compiler error. And let's also be extra extra sure that we don't have a `std::exception_ptr` here too please:
```suggestion
  static_assert(sizeof...(args) > 0, "Expecting an error");
  static_assert(!std::is_same_v<std::decay_t<Args...>, std::exception_ptr>);
  return std::make_exception_ptr(std::forward<Args>(args)...);
```

eventuals/terminal.h
```diff
@@ -200,6 +200,20 @@ struct StoppedException : public std::exception {
...
+}
+
+inline auto make_exception_ptr_or_forward(std::exception_ptr&& error) {
```
benh, line 211: What happens if somewhere you don't have an rvalue reference? Will it try and call the above case? To make sure that this specialization will always be picked how about we take `error` by value instead.

benh (approved):

PR #227, opened 2022-02-23
Inline a few header methods

rjhuijsman:
If these methods are not inlined, they'll cause "duplicate definition"
errors when used in a program that has multiple translation units.

TESTED: caused build failures in upcoming `reboot-dev/respect` change
previously, build failures now resolved.

### DIFF
--- a/eventuals/rsa.h
+++ b/eventuals/rsa.h
@@ -208,7 +208,7 @@ namespace pem {
 
 // Returns an expected 'std::string' with the encoded private key in
 // PEM format or an unexpected.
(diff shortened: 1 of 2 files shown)


rjhuijsman:
I believe the windows build failure is due to an issue @benh is already debugging; I'll wait for a fix from @benh

benh (approved):
Thank you!

Unfortunately mergequeue can only be configured to a single organization right now, sorry.

And if you rebase you should be all set for building on Windows!

PR #228, opened 2022-02-23
Ensure errors derive from std::exception

benh:
This will make it much easier to use 'Catch()` and be able to easily
print out 'what()'.


eventuals/expected.h
```diff
@@ -187,9 +187,31 @@ auto Expected(T t) {
 
 ////////////////////////////////////////////////////////////////////////
 
-template <typename T>
-auto Unexpected(T t) {
-  return _Expected<_Unexpected>(make_exception_ptr_or_forward(std::move(t)));
+template <typename Error>
+auto Unexpected(Error error) {
```
onelxj, line 191: If 'Error' is 'std::runtime_error', 'std::exception_ptr' would be created, should we propagate 'std::runtime_error' as it is?

onelxj (approved):

PR #229, opened 2022-02-24
Prevent calling 'Fail' with no errors

onelxj:


eventuals/catch.h
```diff
@@ -25,6 +25,7 @@ struct _Catch final {
 
     template <typename... Args>
     void Fail(Args&&... args) {
+      static_assert(sizeof...(args) > 0, "Calling 'Fail' with no error");
```
benh, line 28: What if we just did the following everywhere instead?
```suggestion
    template <typename Error>
    void Fail(Error&& error) {
```

onelxj: In the future, we can provide multiple error types for `ErrorFrom`, we can imagine that there could be multiple types for `Fail`. If we do like this `Fail(Error&& error)`, I believe that we need to rewrite other functions like 'Start', 'Ended', etc.

By the way, I thought a bit about the next iteration for `Composable` and I believe that for `ErrorFrom` we need to provide multiple types, but only one type for 'ValueFrom'.

benh: It totally makes sense that we'll want multiple error _types_, but not multiple errors. And if someone does want to send multiple errors then they can just send them as a `std::tuple` and then people can `std::apply` to unpack themselves, or use our `Unpack()` from `eventuals/unpack.h`.

We really simplify the code when we know we always have a single error, so let's do it!

eventuals/closure.h
```diff
@@ -28,9 +28,9 @@ struct _Closure final {
       continuation().Start(std::forward<Args>(args)...);
     }
 
-    template <typename... Args>
-    void Fail(Args&&... args) {
-      continuation().Fail(std::forward<Args>(args)...);
+    template <typename Error>
+    void Fail(Error&& error) {
+      continuation().Fail(std::move(error));
```
benh, line 33: These should still be all `std::forward`!

benh (changes requested):

benh (approved):
Awesome! Please get this committed but can you also go ahead and see if any changes are necessary for `eventuals-grpc`? Thanks!

benh:
@onelxj you changed all of these from `std::forward` to `std::move`, they all still needed to be `std::forward`. I made that comment but I didn't double check that you correctly fixed them all. Please go find all places where you changed from `std::forward` and put them back to `std::forward` just without the `...`.

PR #230, opened 2022-02-24
Revert "Ensure errors derive from std::exception"

rjhuijsman:
Reverts 3rdparty/eventuals#228

This is [causing build failures](https://github.com/reboot-dev/respect-integration/runs/5324237959?check_suite_focus=true) when used with our integration tests:
```
external/com_github_3rdparty_eventuals/eventuals/terminal.h:215:12: error: static assertion failed: Expecting a type derived from std::exception
Target //tests/test_client_eventuals_server:test_container failed to build
```

This build failure is blocking progress for myself and @CodingCanuck, so we have to either roll forward or roll back. I had a cursory glance at whether I could roll forward, but can't quite wrap my head around what's going on. Therefore I'm opting to roll back to unblock progress, allowing for a more leisurely investigation.

TESTED: ran `respect-integration`'s `make tests/test_client_eventuals_server` to confirm that the build is fixed by rolling back this PR.


rjhuijsman:
FYI @benh

sp33drac3r (approved):
Get back. Get back. Get back to where you once belonged.
🔙

benh:
@rjhuijsman just needed to get https://github.com/3rdparty/eventuals-grpc/pull/78 merged then reboot-dev/respect builds without errors. But since https://github.com/3rdparty/eventuals-grpc/pull/78 depends on what was just reverted, now we need to revert the revert which I've done in #232.

CodingCanuck:
> @rjhuijsman just needed to get [3rdparty/eventuals-grpc#78](https://github.com/3rdparty/eventuals-grpc/pull/78) merged then reboot-dev/respect builds without errors. But since [3rdparty/eventuals-grpc#78](https://github.com/3rdparty/eventuals-grpc/pull/78) depends on what was just reverted, now we need to revert the revert which I've done in #232.

I'm just trying to understand what I think happened here: did respect-integration pull in a newer version of the eventuals repo (or try to), but the prerequisite matching fix in eventuals-grpc hadn't landed yet?

So if respect-integration = A, eventuals-grpc = B, and eventuals = C, A needs B and B needs C. A got a new version of C but didn't get a new version of B (which didn't exist / couldn't have been submitted before a new version of C was submitted), so that broke things?

This sounds like a counter-argument for directly including eventuals as a submodule in https://github.com/reboot-dev/respect-integration/tree/main/submodules: if the version of eventuals has to match what eventuals-grpc is built with, maybe we should go with your earlier suggestion of "inheriting" eventuals through eventuals-grpc.

But IMHO this flags a deeper module / build dependency question, similar to what Georg flagged. In the A -> B -> C example, we've wound up with both C and C' that are different versions of the same library, and we're trying to build both versions in incompatible ways. This feels brittle, but I'm not sure of how to solve it, especially for repos where we intentionally want releases to happen separately (i.e. you can depend on eventuals without depending on eventuals-grpc, which means you could depend on different unrelated releases of both [the same way we depend on different versions of absl and gRPC]).

This sounds like a great case study to dig into when we discuss submodules / polyrepos next week!

PR #231, opened 2022-02-24
Add Dazel configuration

rjhuijsman:
Before this PR the `eventuals` repo could only build if you had (the
right version of) `bazel` installed on your machine. To be more hermetic
we'd like to avoid this hope-you-installed-it-right type situations. In
fact, our GitHub Codespaces don't (currently) install `bazel` at all!

This PR adds a symlink to our standard `Dockerfile.dazel` in
the `dev-tools` subrepo, allowing us to use `dazel` to build in a
repeatable way. Dazel will, unfortunately, look for a `WORKSPACE` file
instead of a `WORKSPACE.dazel` file, so we add a symlink to that also.

This is a step towards https://github.com/reboot-dev/company/issues/34

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1,4 +1,4 @@
-# Workflow for building and testing eventuals on macOS, Ubuntu and Windows.  
+# Workflow for building and testing eventuals on macOS, Ubuntu and Windows.
 name: Build and Run all tests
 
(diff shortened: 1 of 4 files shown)


benh (approved):

rjhuijsman:
Mergequeue seems to be struggling with this, probably because the checks have changed since this PR was filed. Will squash and merge manually.

PR #233, opened 2022-02-26
Introduce initial support for "static" errors

benh:
Previously there was no way to know, or track, what errors might
propagate through a an eventual pipeline. This commit introduces the
initial components necessary to do that.

Next step is to add 'ErrorsFrom' to all eventual types and then update
'Composed' to not use 'ErrorsFromComposed' but instead assume that all
eventuals that are composed 'HaveErrorsFrom'.

### DIFF
--- a/eventuals/expected.h
+++ b/eventuals/expected.h
@@ -4,6 +4,7 @@
 
 #include "eventuals/eventual.h"
 #include "eventuals/terminal.h"
+#include "eventuals/type-traits.h"
 
 ////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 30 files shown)


eventuals/expected.h
```diff
@@ -1,9 +1,11 @@
 #pragma once
 
+#include <iostream>
```
onelxj, line 3: Do we need this?

eventuals/expected.h
```diff
@@ -26,54 +28,96 @@ Overloaded(Fs...) -> Overloaded<Fs...>;
...
   template <typename Arg>
   using ValueFrom = Value_;
 
-  // NOTE: providing a default constructor so this type can be used as
-  // a type parameter to 'std::promise' which fails on MSVC (see
-  // https://stackoverflow.com/questions/28991694 which also ran into
-  // this issue; very surprised this hasn't been fixed, apparently
-  // there aren't many users of 'std::promise' on Windows).
-  _Expected() {}
+  template <typename Arg, typename Errors>
```
onelxj, line 53: we can make
`template <typename Arg, typename... Errors>...` - that isn't require passing `std::tuple<>` in tests, but need to add a comment that it expects a `tuple` instead of a set of types.

test/concurrent-emit-fail-interrupt.cc
```diff
@@ -46,10 +46,12 @@ TYPED_TEST(ConcurrentTypedTest, EmitFailInterrupt) {
                })
         | this->ConcurrentOrConcurrentOrdered([&]() {
             return Map(Let([&](int& i) {
-              return Eventual<std::string>([&](auto& k) {
-                k.Fail(std::runtime_error("error"));
-                interrupt.Trigger();
-              });
+              return Eventual<std::string>()
```
onelxj, line 49: Let's make Eventual be configurated using different order (for example raises->interruptible->start, start->raises) because it depends and need to be sure that the future changes won't break it.

test/type-traits.cc
```diff
@@ -0,0 +1,110 @@
+#include "eventuals/type-traits.h"
```
onelxj, line 1: Thanks for creating a new file with tests, love it!

eventuals/expected.h
benh: I'm not following, can you elaborate?

test/concurrent-emit-fail-interrupt.cc
benh: There isn't anything preventing you from configuring in any order!

eventuals/expected.h
benh: Thanks!

eventuals/expected.h
benh: Let's do this in a separate PR!

onelxj (approved):

PR #237, opened 2022-03-02
`ErrorsFrom` for `Task`

onelxj:


onelxj:
Need to be rebased this after [this](https://github.com/3rdparty/eventuals/pull/237/commits/9eab308e4deada4c3183b9f166cd8dbb1cd31363) one.

test/task.cc
```diff
@@ -534,3 +580,23 @@ TEST(Task, RefSuccess) {
...
+
+TEST(Task, RaisesWith) {
+  auto e = []() {
+    return Task::Of<int>::Raises<std::runtime_error>::With<int, std::string>(
+        42,
+        "hello world",
+        [](auto i, auto s) {
+          EXPECT_EQ(s, "hello world");
+          return Just(i)
+              .template raises<std::runtime_error>();
```
benh, line 592: That's some kung fu magic! Let's be explicit for our poor readers, and then if the `Just()` implementation ever changes in the future this line of code won't break awkwardly.
```suggestion
return Eventual<int>()
    .template raises<std::runtime_error>()
    .start([&](auto& k) {
      return k.Start(i);
    });
```

test/task.cc
```diff
@@ -391,34 +416,40 @@ TEST(Task, Success) {
...
     }
   };
 
   struct Async : public Base {
-    Task::Of<int> GetTask() override {
+    Task::Of<int>::Raises<std::runtime_error> GetTask() override {
       return []() {
-        return Just(20);
+        return Just(20)
+            .raises<std::runtime_error>();
```
benh, line 446: See other comment!

test/task.cc
```diff
@@ -391,34 +416,40 @@ TEST(Task, Success) {
...
   struct Base {
-    virtual Task::Of<int> GetTask() = 0;
+    virtual Task::Of<int>::Raises<std::runtime_error> GetTask() = 0;
   };
 
   struct Sync : public Base {
-    Task::Of<int> GetTask() override {
-      return Task::Success(10);
+    Task::Of<int>::Raises<std::runtime_error> GetTask() override {
+      return Task::Success<int, std::runtime_error>(10);
```
benh, line 438: Let's make our API not require specifying the error type when we know we're successful!

eventuals/task.h
```diff
@@ -570,57 +598,66 @@ class TaskFromToWith {
...
   }
 
-  template <typename Value>
+  template <typename Value, typename... Errors>
   static auto Success(std::reference_wrapper<Value> value) {
-    return TaskFromToWith<void, Value&>(std::move(value));
+    return TaskFromToWith<
+        void,
+        Value&,
+        std::tuple<Errors...>>(std::move(value));
```
benh, line 652: Errors is always empty! See [here in expected.h](https://github.com/3rdparty/eventuals/blob/main/eventuals/expected.h#L193).

eventuals/task.h
```diff
@@ -570,57 +598,66 @@ class TaskFromToWith {
...
   }
 
+  template <typename... Errors>
   static auto Success() {
-    return TaskFromToWith<void, void>(std::monostate{});
+    return TaskFromToWith<
+        void,
+        void,
+        std::tuple<Errors...>>(std::monostate{});
```
benh, line 660: See comment above.

eventuals/task.h
```diff
@@ -570,57 +598,66 @@ class TaskFromToWith {
...
-
-      template <typename F>
-      To(F f)
-        : TaskFromToWith<From_, To_>(std::move(f)) {}
+template <
+    typename From_ = void,
+    typename To_ = void,
+    typename Errors_ = std::tuple<>,
+    typename... Args_>
+struct TaskBuilder final
```
benh, line 606: I LOVE THIS BUILDER PATTERN! GREAT IDEA!!!

Once we settle on the exact naming let's do the same in expected.h.

eventuals/task.h
```diff
@@ -570,57 +598,66 @@ class TaskFromToWith {
...
+  : public TaskFromToWith<From_, To_, Errors_, Args_...> {
+  template <typename T>
+  using From = TaskBuilder<T, To_, Errors_, Args_...>;
```
benh, line 609: What happens if someone does `From<>` more than once when building? Can we add `has_from_` fields and do:

```cpp
template <typename T>
using From = std::enable_if_t<!has_from_, TaskBuilder<...>>;
```

eventuals/task.h
```diff
@@ -570,57 +598,66 @@ class TaskFromToWith {
...
 
-      To(To&& that) = default;
+  template <typename T>
+  using To = TaskBuilder<From_, T, Errors_, Args_...>;
 
-      ~To() override = default;
-    };
+  template <typename T>
+  using Of = TaskBuilder<void, T, Errors_, Args_...>;
```
benh, line 615: Since `Of` is an alias for both `From` and `To` then we can check in the `std::enable_if_t` as appropriate!

eventuals/task.h
```diff
@@ -570,57 +598,66 @@ class TaskFromToWith {
...
+  using Raises = TaskBuilder<From_, To_, std::tuple<Errors...>, Args_...>;
 
-    From(From&& that) = default;
+  template <typename F>
+  TaskBuilder(Args_... args, F f)
+    : TaskFromToWith<
+        From_,
+        To_,
+        Errors_,
+        Args_...>(std::move(args)..., std::move(f)) {}
```
benh, line 629: We should be able to just replace all of this with:

```suggestion
    using TaskFromToWith<..., ..., etc>::TaskFromToWith;
```

See how we did this in [`expected.h`](https://github.com/3rdparty/eventuals/blob/main/eventuals/expected.h#L178).

eventuals/task.h
```diff
@@ -570,57 +598,66 @@ class TaskFromToWith {
...
 
-    ~From() override = default;
-  };
+  ~TaskBuilder() = default;
 
-  template <typename To_>
-  using Of = From<void>::To<To_>;
+  TaskBuilder(TaskBuilder&&) = default;
```
benh, line 633: Shouldn't need this because of comment above!

eventuals/task.h
```diff
@@ -570,57 +598,66 @@ class TaskFromToWith {
...
+  TaskBuilder(Args_... args, F f)
+    : TaskFromToWith<
+        From_,
+        To_,
+        Errors_,
+        Args_...>(std::move(args)..., std::move(f)) {}
 
-    ~From() override = default;
-  };
+  ~TaskBuilder() = default;
```
benh, line 631:
```suggestion
  ~TaskBuilder() override = default;
```

eventuals/task.h
```diff
@@ -570,57 +598,66 @@ class TaskFromToWith {
...
 
-  template <typename To_>
-  using Of = From<void>::To<To_>;
+  TaskBuilder(TaskBuilder&&) = default;
 
-  template <typename... Args_>
-  using With = From<void>::To<void>::With<Args_...>;
+  TaskBuilder(MonostateIfVoidOr<To_> value)
+    : TaskFromToWith<From_, To_, Errors_, Args_...>(std::move(value)) {}
```
benh, line 636: Again, shouldn't need this by pulling in parent constructors, didn't think of this when we were doing it before but it should work!

benh (changes requested):

test/generator.cc
```diff
@@ -322,7 +322,9 @@ TEST(Generator, TaskWithGenerator) {
     };
   };
 
-  auto task = [&]() -> Task::Of<std::vector<int>> {
+  // TODO: 'Raises' should be removed since we support 'ErrorsFrom' for all
+  // eventuals. Now it propagates 'std::exception' as general case.
+  auto task = [&]() -> Task::Of<std::vector<int>>::Raises<std::exception> {
```
benh, line 327: I don't understand why we have to use `Raises` if it doesn't raise anything?

eventuals/task.h
```diff
@@ -448,28 +472,73 @@ struct _TaskFromToWith final {
...
     typename... Args_>
-class TaskFromToWith {
+class TaskFromToWith final {
  public:
   template <typename Arg>
   using ValueFrom = To_;
 
-  TaskFromToWith(MonostateIfVoidOrReferenceWrapperOr<To_> value)
-    : e_(std::move(value)) {}
+  template <typename Arg, typename Errors = std::tuple<>>
```
benh, line 482: No default here, someone should always give us the errors to start with.
```suggestion
  template <typename Arg, typename Errors>
```

eventuals/task.h
```diff
@@ -448,28 +472,73 @@ struct _TaskFromToWith final {
...
     : e_(std::move(that.e_)) {
     CHECK(!k_.has_value()) << "moving after starting";
   }
 
-  virtual ~TaskFromToWith() = default;
+  ~TaskFromToWith() = default;
+
+  auto& e() {
+    return e_;
+  }
```
benh, line 534:
```suggestion
```

eventuals/task.h
```diff
@@ -448,28 +472,73 @@ struct _TaskFromToWith final {
...
     : e_(std::move(args)..., std::move(f)) {}
 
+  TaskFromToWith(MonostateIfVoidOrReferenceWrapperOr<To_> value)
+    : e_(std::move(value)) {}
+
+  template <
+      typename E,
+      std::enable_if_t<tuple_types_subset_v<E, Errors_>, int> = 0>
+  TaskFromToWith(TaskFromToWith<From_, To_, E, Args_...>&& that)
+    : e_(std::move(that.e().value_or_dispatch_), std::move(that.e().args_)) {
```
benh, line 521:
```suggestion
    : e_(std::move(that.e_.value_or_dispatch_), std::move(that.e_.args_)) {
```

eventuals/task.h
```diff
@@ -359,6 +364,12 @@ struct _TaskFromToWith final {
 
       using Value = typename E::template ValueFrom<From_>;
 
+      using ErrorsFromE = typename E::template ErrorsFrom<From_, std::tuple<>>;
+
+      static_assert(
+          eventuals::tuple_types_subset_v<ErrorsFromE, Errors_>,
+          "Specified errors can't be thrown from 'Task'");
```
benh, line 371:
```suggestion
          "Specified errors can't be raised within 'Task'");
```

eventuals/task.h
```diff
@@ -544,83 +626,29 @@ class TaskFromToWith {
...
   static auto Success(Value value) {
-    return TaskFromToWith<void, Value>(std::move(value));
+    return TaskFromToWith<
+        void,
+        Value,
+        std::tuple<>>(std::move(value));
   }
 
-  template <typename Value>
+  template <typename Value, typename... Errors>
```
benh, line 639: Success never has errors!
```suggestion
  template <typename Value>
```

benh (changes requested):

eventuals/task.h
onelxj: Should we make 'e_' public instead?

eventuals/task.h
```diff
@@ -493,16 +558,29 @@ class TaskFromToWith {
     k_->Start();
   }
 
-  template <typename Arg>
+  template <typename Error>
   void Fail(
-      Arg&& arg,
+      Error&& arg,
```
benh, line 563:
```suggestion
      Error&& error,
```

eventuals/task.h
```diff
@@ -448,28 +472,69 @@ struct _TaskFromToWith final {
 template <
     typename From_,
     typename To_,
+    typename Errors_,
     typename... Args_>
-class TaskFromToWith {
+class TaskFromToWith final {
```
benh, line 477:
```suggestion
class _Task final {
```

benh (approved):

PR #238, opened 2022-03-03
Fix concurrent tests

FolMing:
Not sure what happened while I was away, but still...

Trying to run concurrent tests while commenting everything out (Windows):
Before the fix:
image

After the fix:
image

### DIFF
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -49,6 +49,9 @@ cc_library(
         "//:eventuals",
         "@com_github_google_googletest//:gtest",
     ],
+    # Setting it to False or leaving it absent will prevent concurrent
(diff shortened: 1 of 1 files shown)


test/BUILD.bazel
```diff
@@ -49,6 +49,8 @@ cc_library(
         "//:eventuals",
         "@com_github_google_googletest//:gtest",
     ],
+    # Setting it to False will prevent concurrent tests from running.
```
benh, line 52:
```suggestion
    # Setting it to False or leaving it absent will prevent concurrent 
    # tests from running on Windows.
```

benh (approved):
Thanks @FolMing!

They're definitely still running on macOS for me!

PR #241, opened 2022-03-03
Add explicit `has_value` method

while-false:
This allows for explicit checks of `Expected::Of`s for whether the instance has a value of not instead of relying on the less explicit cast to bool.


onelxj:
What do you think about `success` instead of `has_value`, the reason that expected might store both `success value` and `std::exception_ptr` and in the future we can check if expected stores `std::exception_ptr` as `!success()` instead of `!has_value()` that brings some missunderstandings.

PR #245, opened 2022-03-07
`ErrorsFrom` for `Generator`

onelxj:


benh (changes requested):
Where are the tests that show that `Raises` works?

benh (approved):

PR #247, opened 2022-03-10
Update from gRPC 1.40.0 to 1.45.0-pre1

CodingCanuck:
This pulled in a newer version of rules_python (through an implicit
dependency via gRPC, which I don't like) which required porting off of
the `pip_import` rule which was deprecated in release 0.1.0 in Oct 2020
(!): https://github.com/bazelbuild/rules_python/releases/tag/0.1.0 and
deleted in release 0.6.0 in Jan 2022:
https://github.com/bazelbuild/rules_python/releases/tag/0.6.0.

This removes the need to depend on a 1-year old absl library (!).

It also unbreaks bazel 5.0: fixes https://github.com/3rdparty/eventuals-grpc/issues/65.

### DIFF
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -27,17 +27,11 @@ grpc_extra_deps()
 #   (2) They're only used for tests and we assume those depending on
 #       eventuals don't need to depend on the tests.
 
-load("@io_bazel_rules_python//python:pip.bzl", "pip_import", "pip_repositories")
+load("@rules_python//python:pip.bzl", "pip_install")
 
-pip_import(
+pip_install(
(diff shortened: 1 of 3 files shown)


rjhuijsman (approved):
This is very exciting! 🎉

CodingCanuck:
I confirmed that `make test` succeeds in the eventuals repo with this change (using our current bazel 4.2.x).

CodingCanuck:
> I confirmed that `make test` succeeds in the eventuals repo with this change (using our current bazel 4.2.x).

Note that the windows build failing is a known issue at HEAD: https://github.com/3rdparty/eventuals/issues/248

PR #251, opened 2022-03-11
Do not use secret credentials for forked repos on GitGub Actions

onelxj:


onelxj:
Actually I tried to create a new env variable and set it conditionally, but with no result, so decided to make another run with applicable arguments for now.

.github/workflows/build_and_test.yml
```diff
@@ -95,7 +108,6 @@ jobs:
             test/... \
             --test_arg=--gtest_shuffle \
             --test_arg=--gtest_repeat=100
-
```
rjhuijsman, line 98: Let's not drop this whitespace line - it's nice for visual separation.

.github/workflows/build_and_test.yml
```diff
@@ -95,7 +108,6 @@ jobs:
...
       - name: Debug using tmate (if failure)
         if: ${{ failure() }}
-        uses: mxschmitt/action-tmate@v3
```
rjhuijsman, line 101: Let's not drop this newline at the end of the file

rjhuijsman (approved):
A few requests, then feel free to submit! Thanks for fixing this!

CodingCanuck (approved):
Thanks, Zakhar!

PR #253, opened 2022-03-16
Add `ErrorsFrom` for `Loop`

onelxj:


eventuals/foreach.h
```diff
@@ -13,7 +14,9 @@ template <typename E, typename F>
 auto Foreach(E e, F f) {
   return std::move(e)
       | Map(std::move(f))
-      | Loop();
+      | Loop()
+            .raises_tuple<
+                ErrorsFromMaybeComposable<decltype(e), void, std::tuple<>>>();
```
benh, line 19: But this `Loop()` doesn't raise that ...

That will get raised from `E`, so anyone who does `ErrorsFrom` on what gets returned from `Foreach` will have the correct possible errors, but we shouldn't need to add that on `Loop`.

eventuals/concurrent.h
```diff
@@ -457,6 +457,7 @@ struct _Concurrent final {
                notify_egress_();
              }))
           | Loop()
+                .raises<std::exception>()
```
benh, line 460:
```suggestion
```

eventuals/concurrent.h
```diff
@@ -508,6 +509,7 @@ struct _Concurrent final {
              }))
           | Until([](bool done) { return done; })
           | Loop() // Eagerly try to get next value to run concurrently!
+                .raises<std::exception>()
```
benh, line 512:
```suggestion
```

eventuals/collect.h
```diff
@@ -15,6 +15,7 @@ template <typename Container>
 auto Collect() {
   return Loop<Container>()
       .context(Container())
+      .template raises<std::exception>()
```
benh, line 18: This `Loop` never calls `k.Fail(...)`!
```suggestion
```

test/stream.cc
```diff
@@ -54,6 +54,7 @@ TEST(StreamTest, Succeed) {
                })
         | Loop<int>()
               .context(0)
+              .raises<std::runtime_error>()
```
benh, line 57: Never raises.
```suggestion
```

test/stream.cc
```diff
@@ -259,6 +271,7 @@ TEST(StreamTest, InterruptLoop) {
         | Loop<int>()
               .context(Lazy<std::atomic<bool>>(false))
               .interruptible()
+              .raises()
```
benh, line 274: Why `raises()` instead of `raises<std::runtime_error>()` which is what it looks like it raises on line 297?

benh (changes requested):

eventuals/loop.h
```diff
@@ -158,15 +217,18 @@ struct _Loop final {
       typename Stop_,
       bool Interruptible_,
       typename Value_,
-      typename... Errors_>
+      typename Errors_>
   struct Builder final {
     template <typename Arg>
     using ValueFrom = Value_;
 
+    template <typename Arg, typename Errors = std::tuple<>>
```
benh, line 225:
```suggestion
    template <typename Arg, typename Errors>
```

test/stream.cc
```diff
@@ -157,6 +163,11 @@ TEST(StreamTest, Fail) {
               });
   };
 
+  static_assert(
+      eventuals::tuple_types_unordered_equals_v<
+          decltype(s())::ErrorsFrom<void, std::tuple<>>,
+          std::tuple<std::exception, std::runtime_error>>);
```
benh, line 169: Please commit the code before this so we can remove `std::exception` from the resulting list?
```suggestion
          std::tuple<std::runtime_error>>);
```

test/stream.cc
```diff
@@ -69,6 +69,11 @@ TEST(StreamTest, Succeed) {
               });
   };
 
+  static_assert(
+      eventuals::tuple_types_unordered_equals_v<
+          decltype(s())::ErrorsFrom<void, std::tuple<>>,
+          std::tuple<std::exception>>);
+
```
benh, line 76:
```suggestion
```

test/stream.cc
```diff
@@ -259,6 +270,7 @@ TEST(StreamTest, InterruptLoop) {
         | Loop<int>()
               .context(Lazy<std::atomic<bool>>(false))
               .interruptible()
+              .raises<std::runtime_error>()
```
benh, line 273: Add a `static_assert` for this one?

eventuals/loop.h
```diff
@@ -294,7 +356,34 @@ struct _Loop final {
...
+          std::move(context_),
+          std::move(begin_),
+          std::move(body_),
+          std::move(ended_),
+          std::move(fail_),
+          std::move(stop_));
+    }
+
+    template <typename Errors>
+    auto raises_tuple() && {
```
benh, line 381: Dead code!

eventuals/type-traits.h
```diff
@@ -47,6 +47,14 @@ struct HasEmplaceBack<
 
 ////////////////////////////////////////////////////////////////////////
 
+template <typename>
+struct is_tuple : std::false_type {};
+
+template <typename... T>
+struct is_tuple<std::tuple<T...>> : std::true_type {};
```
benh, line 54: Dead code?

benh (approved):
This should be submitted after `Stream` it seems so we don't have those `std::exception` in there!

PR #256, opened 2022-03-21
Fix (this == &that) case for Stream's Continuation's move assignment operator

FolMing:
Should be self-explanatory.


benh (approved):
There are other spots where this was missed, notably I can think of `eventual.h`, can you please do a search and fix this everywhere? Thanks!

PR #257, opened 2022-03-21
Use move semantic for Timer's methods

FolMing:
Notice how in EventLoop::Clock::Timer we are using std::move on const reference, lol!

### DIFF
--- a/eventuals/event-loop.h
+++ b/eventuals/event-loop.h
@@ -147,7 +147,7 @@ class EventLoop final : public Scheduler {
 
     std::chrono::nanoseconds Now();
 
-    auto Timer(const std::chrono::nanoseconds& nanoseconds);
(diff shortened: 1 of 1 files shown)


eventuals/event-loop.cc
```diff
@@ -57,7 +57,7 @@ void EventLoop::Clock::Resume() {
 
 ////////////////////////////////////////////////////////////////////////
 
-void EventLoop::Clock::Advance(const std::chrono::nanoseconds& nanoseconds) {
+void EventLoop::Clock::Advance(std::chrono::nanoseconds&& nanoseconds) {
```
benh, line 60: Doesn't seem like we need to change this one though given we don't need to copy or move so let's stick with `const &` so we simplify callers.

benh (approved):

PR #258, opened 2022-03-21
Try to minimize paths for bazel output files

### LINKED ISSUE #248: Build error on windows
Eventuals fails to build on windows after eventuals-grpc was merge into this repo in https://github.com/3rdparty/eventuals/pull/243.

The issue looks like too-long file paths hitting a Windows path length limit: @ArthurBandaryk is fixing this in https://github.com/3rdparty/eventuals/pull/244

FolMing:
Fixes #248.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -28,16 +28,20 @@ jobs:
             # build/test on Linux but on macOS!
             bazel-command: "bazelisk"
             bazel-config: ""
+            output-user_root: ""
           - os: ubuntu-latest
             sudo-command: "sudo"
             bazel-command: "dazel"
             bazel-config: "--config=asan"
+            output-user_root: ""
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -79,7 +79,9 @@ jobs:
         # when a workflow is triggered from a forked repository.
         if: github.event.pull_request.head.repo.full_name == '3rdparty/eventuals'
         run: |
-          ${{ matrix.bazel-command }} build \
+          ${{ matrix.bazel-command }} \
+            --output_user_root=C:/bzl \
```
benh, line 83: Need to make this:
```suggestion
            ${{ matrix.output-user_root }} \
```
So that we only do this for Windows!

benh (approved):
Thanks! 💪

PR #260, opened 2022-03-24
Add `ErrorsFrom` for `Catch`

onelxj:


eventuals/catch.h
```diff
@@ -178,10 +182,76 @@ struct _Catch final {
...
+    // Helper that creates the right tuple of 'Errors' for 'Catch'.
+    // If 'Catch' handles some type from upstream 'Errors' it won't be
+    // added to the result errors tuple.
+    template <
+        typename CatchErrors,
+        typename UpstreamErrors,
+        typename Processed = std::tuple<>,
+        size_t I = 0,
+        bool End = I == std::tuple_size_v<UpstreamErrors>>
+    struct Intersection {
```
benh, line 213: This isn't "intersection", it's set subtraction!

Let's move this to be a helper in `type-traits.h` and call it `tuple_types_subtract` and let's have it look like: `tuple_types_subtract_t<Errors, CaughtErrors>`. See [here](https://github.com/3rdparty/eventuals/blob/main/eventuals/type-traits.h#L171) for an example.

Also, let's implement this by using `tuple_types_contains`, that is, create a "new" set that only includes things from the left hand side `Errors` if the type is _not_ `tuple_types_contains` in the right hand side.

benh (changes requested):

eventuals/catch.h
onelxj: Actually it uses `tuple_types_contains` at `std::disjunction` now.

eventuals/type-traits.h
```diff
@@ -172,6 +173,32 @@ using tuple_types_union_t = typename tuple_types_union<Left, Right>::type;
...
+
+template <typename Left, typename... Lefts, typename Right>
+struct tuple_types_subtract<
+    std::tuple<Left, Lefts...>,
+    Right> {
+  using type = tuple_types_concatenate_t<
+      std::conditional_t<
+          std::disjunction_v<
+              tuple_types_contains<std::exception, Right>,
+              tuple_types_contains<std::exception_ptr, Right>,
```
benh, line 189: You're writing a _generic_ function called `tuple_types_subtract`, what does that have to do with `std::exception` and `std::exception_ptr`!?

eventuals/type-traits.h
```diff
@@ -172,6 +173,32 @@ using tuple_types_union_t = typename tuple_types_union<Left, Right>::type;
...
+              tuple_types_contains<Left, Right>>,
+          std::tuple<>,
+          std::tuple<Left>>,
+      typename tuple_types_subtract<std::tuple<Lefts...>, Right>::type>;
+};
+
+template <typename Errors, typename CaughtErrors>
+using tuple_types_subtract_t =
```
benh, line 197: Where are the tests for this in `test/type-traits.cc`!?

eventuals/type-traits.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <stdexcept>
```
benh, line 3:
```suggestion
```

eventuals/catch.h
```diff
@@ -178,10 +183,37 @@ struct _Catch final {
...
+          std::tuple<typename CatchHandler::Error>,
+          typename CatchErrorsUnpacker<Tail...>::Errors>;
+    };
+
+    template <typename CatchHandler>
+    struct CatchErrorsUnpacker<CatchHandler> {
+      using Errors = std::tuple<typename CatchHandler::Error>;
+    };
+
+    ////////////////////////////////////////////////////////////////////////
```
benh, line 205: Isn't this just:
```suggestion
    using CatchErrors_ = std::tuple<typename CatchHandlers_::Error...>;
```

benh (changes requested):

eventuals/catch.h
```diff
@@ -182,6 +187,16 @@ struct _Catch final {
     template <typename Arg>
     using ValueFrom = Arg;
 
+    using CatchErrors_ = std::tuple<typename CatchHandlers_::Error...>;
+
+    template <typename Arg, typename Errors>
+    using ErrorsFrom = std::conditional_t<
+        std::disjunction_v<
+            tuple_types_contains<std::exception, CatchErrors_>,
+            tuple_types_contains<std::exception_ptr, CatchErrors_>>,
```
benh, line 196:
```suggestion
            // If 'std::exception' is caught then all exceptions will be caught.
            tuple_types_contains<std::exception, CatchErrors_>,
            // A 'std::exception_ptr' implies we have a '.all(...)' and all exceptions are caught.
            tuple_types_contains<std::exception_ptr, CatchErrors_>>,
```

test/type-traits.cc
```diff
@@ -108,3 +109,41 @@ static_assert(
         std::tuple<int>>);
 
 ////////////////////////////////////////////////////////////////////////
+
+static_assert(std::is_same_v<
```
benh, line 113:
```suggestion
static_assert(
    std::is_same_v<
```
Let's use the same formatting as above to be consistent in the file. Here and the rest of what you added please.

benh (approved):

Issue #271, opened 2022-03-28
`//test:grpc` exhibits flaky timeouts


CodingCanuck:
`//test:grpc` seems to timeout randomly.

Sample failure: https://github.com/3rdparty/eventuals/runs/5723441118?check_suite_focus=true

It looks like the actual test isn't that slow, running in under 1 second on average:
```
2022-03-28T16:34:08.6709730Z [----------] 19 tests from EventualsGrpcTest (894 ms total)
```

But we run it 100 times: https://github.com/3rdparty/eventuals/blob/main/.github/workflows/build_and_test.yml#L126-L127

It's easy for a ~1s test run 100x to take over 60s.

We use the `--gtest_repeat` flag to run tests `N` times. We could instead use bazel's `--runs_per_test` tag to parallelize these tests runs, but I think that's not what we want: we're also setting the `--gtest_shuffle` flag, and that combined with `--gtest_repeat` guarantees that a different random seed is used for each test: https://github.com/google/googletest/blob/main/docs/advanced.md#shuffling-the-tests

We could just increase the test timeout here, but it's not obvious to me that our current setup (exactly 100 tests run in sequent) is a locally optimal setup.

PR #273, opened 2022-03-28
Increase `//test:grpc`'s timeout

CodingCanuck:
Fixes https://github.com/3rdparty/eventuals/issues/271, though I added a
`TODO` to find a way to fix this via speedups rather than a longer
timeout.


CodingCanuck:
Closing in favor of https://github.com/3rdparty/eventuals/pull/280

Issue #281, opened 2022-03-28
protoc-gen-eventuals needs to generate code that uses `Task::Raises`


benh:
For example, [here](https://github.com/3rdparty/eventuals/blob/main/protoc-gen-eventuals/templates/eventuals.h.j2#L23) and [here](https://github.com/3rdparty/eventuals/blob/main/protoc-gen-eventuals/templates/eventuals.h.j2#L37) and [here](https://github.com/3rdparty/eventuals/blob/main/protoc-gen-eventuals/templates/eventuals.h.j2#L66) and [here](https://github.com/3rdparty/eventuals/blob/main/protoc-gen-eventuals/templates/eventuals.cc.j2#L30).

See also https://github.com/reboot-dev/respect/issues/230.

PR #282, opened 2022-03-28
Add necessary 'Raises<std::exception>' for protoc-gen-eventuals generated code

benh:
This is missing tests because all of the generated code is missing tests at this point in time.


benh:
Fixes #281

CodingCanuck (approved):
Drive by LGTM!

Issue #283, opened 2022-03-29
Make pre-merge checks fail faster (in under an hour, currently 3h)


CodingCanuck:
This pre-merge check has been running for more than 3 hours: https://github.com/3rdparty/eventuals/runs/5728226500?check_suite_focus=true

It failed in under 23 minutes. After that it started the "Debug using tmate (if failure)" step which has been running for 2h41m.

The workflow doesn't report failure results until after the entire workflow (including the tmate step) completes. In this case, it would've been useful to receive an error report 2h40m ago so I could've quickly retried the failure (it looks flaky).

Can we make these checks fail faster? Options might include:
1) Make tmate run asynchronously / in a non-blocking check
2) Stop running tmate on failures, perhaps switching it to something that can be optionally enabled (e.g. https://github.com/mxschmitt/action-tmate#manually-triggered-debug )
3) Reduce the tmate timeout ( https://github.com/mxschmitt/action-tmate#timeout )

CodingCanuck:
Fixed by https://github.com/3rdparty/eventuals/pull/284

PR #284, opened 2022-03-29
Make tmake debugging opt-in for faster failures

CodingCanuck:
See https://github.com/mxschmitt/action-tmate#manually-triggered-debug

Fixes https://github.com/3rdparty/eventuals/issues/283

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -14,6 +14,17 @@ on:
   pull_request:
     paths-ignore:
       - "**.md"
+  workflow_dispatch:
+    inputs:
+      # Live debug failures using tmate by toggling input parameter
+      # 'debug_enabled':
+      # https://github.com/mxschmitt/action-tmate#manually-triggered-debug
+      # When manually running this workflow:
+      # https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
+      debug_enabled:
+        description: 'Run the build with tmate debugging enabled (https://github.com/marketplace/actions/debugging-with-tmate)'
+        required: false
+        default: false
 
 jobs:
   build-and-test:
@@ -127,6 +138,8 @@ jobs:
             --test_arg=--gtest_repeat=100
 
       - name: Debug using tmate (if failure)
-        if: ${{ failure() }}
         uses: mxschmitt/action-tmate@v3
-        
+        # Optionally enable tmate debugging if the workflow was manually-triggered
+        # with `debug_enabled` set to `true`.
+        # https://github.com/mxschmitt/action-tmate#manually-triggered-debug
+        if: ${{ failure() && github.event_name == 'workflow_dispatch' && github.event.inputs.debug_enabled }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -14,6 +14,14 @@ on:
   pull_request:
     paths-ignore:
       - "**.md"
+  # Optionally use tmate to live debug failures:
+  # https://github.com/mxschmitt/action-tmate#manually-triggered-debug
+  workflow_dispatch:
+    inputs:
```
benh, line 20:
```suggestion
  workflow_dispatch:
    inputs:
      # Live debug failures using tmate by toggling input parameter 'debug_enabled':
      # https://github.com/mxschmitt/action-tmate#manually-triggered-debug
      debug_enabled:
```
Please put the comment on `debug_enabled` instead of the generic `workflow_dispatch`. I also suggested an alternative comment to help convey that you need to manually toggle the input parameter because that wasn't obvious to me at first, but feel free to change to anything that you think better conveys how to accomplish! Including maybe even a link directly to [Manually running a workflow](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) which IMO is more valuable then the link to the tmate documentation because that's more about doing what you did vs helping people that come across this code figure out how to run a workflow with `debug_enabled` toggled.

.github/workflows/build_and_test.yml
```diff
@@ -127,6 +135,7 @@ jobs:
             --test_arg=--gtest_repeat=100
 
       - name: Debug using tmate (if failure)
-        if: ${{ failure() }}
         uses: mxschmitt/action-tmate@v3
-        
\ No newline at end of file
+        # Optionally enable tmate debugging of manually-triggered workflows.
```
benh, line 139: Maybe:
```suggestion
        # Optionally enable tmate debugging if the workflow was manually-triggered 
        # with `debug_enabled` set to `true`.
```

.github/workflows/check_code_style.yml
```diff
@@ -10,10 +10,18 @@ on:
...
-      - "**.md"       
+      - "**.md"
+  # Optionally use tmate to live debug failures:
+  # https://github.com/mxschmitt/action-tmate#manually-triggered-debug
+  workflow_dispatch:
+    inputs:
+      debug_enabled:
+        description: 'Run the build with tmate debugging enabled (https://github.com/marketplace/actions/debugging-with-tmate)'
+        required: false
+        default: false
```
benh, line 24: Actually, how about we remove support for debugging here all together!? In my experience thus far any failures are printed out and we basically just need to go (re)run `clang-format` or `buildifier` so seems like we can simplify here and remove all together.

benh (approved):

.github/workflows/check_code_style.yml
CodingCanuck: Sold: removed.

.github/workflows/build_and_test.yml
CodingCanuck: Applied, and added a link to manually running a workflow.

PR #287, opened 2022-03-29
Add a `make test` command that uses a build cache

CodingCanuck:
Add a `Makefile` that implements a `make test` command which runs dazel
using a remote build cache.

The cache lives on Google Cloud Compute storage, so it requires a
pre-build authorization step that's awkward for users to run manually.
`make test` prompts users for authorization as needed.

On my fresh GitHub Codespace (where I've never run eventuals) this
reduces build / test time to just a couple of minutes.

### DIFF
--- a/.dazelrc
+++ b/.dazelrc
@@ -1,0 +1,4 @@
+# Use a unique Dazel image and instance name so it doesn't conflict with other
+# Dazel instances (e.g. ones in other repos).
+DAZEL_IMAGE_NAME = "dazel-eventuals"
+DAZEL_INSTANCE_NAME = "dazel-eventuals"
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,40 @@
+# Makefile for running eventuals tests.
+
+# GCloud credentials will be required to be present for any target that runs
+# a bazel command using a remote cache, which (since it's a Google Cloud
+# Storage bucket) requires those credentials to function.
+#
+# We normally have the user do a synchronous authentication and place the
+# resulting credentials in a  place where it can get shared with the Docker
+# containers.
+GOOGLE_APPLICATION_CREDENTIALS ?= /tmp/gcloud-access-token.json
+GOOGLE_APPLICATION_CREDENTIALS_DIR := ${shell dirname ${GOOGLE_APPLICATION_CREDENTIALS}}
+
+DAZEL_VOLUMES := ${GOOGLE_APPLICATION_CREDENTIALS_DIR}:${GOOGLE_APPLICATION_CREDENTIALS_DIR}
+
+
+BAZEL_BUILD_ARGS := --verbose_failures
+
+# Use a remote build cache so we can all share our build artifacts.
+# That'll save a new workstation a few minutes of build-time.
+BAZEL_BUILD_ARGS += --remote_cache=https://storage.googleapis.com/reboot-workstations-buildcache --google_default_credentials
+
+# Compute the Google Application Credentials; see a description of what they
+# are at the variable definition above.
+#
+# It is recommended to make this an "order-only" target (use "|" before the
+# target name) to avoid rebuilding artifacts when our credentials change.
+${GOOGLE_APPLICATION_CREDENTIALS}:
+	gcloud auth application-default login
+	mkdir -p $(dir ${GOOGLE_APPLICATION_CREDENTIALS})
+	cp ${HOME}/.config/gcloud/application_default_credentials.json ${GOOGLE_APPLICATION_CREDENTIALS}
+
+.PHONY: test
+test: | ${GOOGLE_APPLICATION_CREDENTIALS}
+	DAZEL_ENV_VARS="`echo GOOGLE_APPLICATION_CREDENTIALS=${GOOGLE_APPLICATION_CREDENTIALS}`" \
+	DAZEL_VOLUMES="${DAZEL_VOLUMES}" \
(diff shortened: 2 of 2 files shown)


rjhuijsman (approved):
Note that this doesn't solve the problem for anyone who's contributing from outside our organization, since their GCloud credentials won't allow them access to the remote cache bucket.

benh:
@CodingCanuck @rjhuijsman I'd like to revisit this approach because I don't want to add yet another build system into the mix.

I'm assuming the reason we didn't do it the same way we do in [respect](https://github.com/reboot-dev/respect/blob/main/.bazelrc#L6) is because not everyone has gcloud access? Is the idea then that those with gcloud access run `make test` and everyone else keeps running `bazel test`? That's not a future I want to encourage.

What about doing something similar to what we did with respect except telling folks to use their own `users.bazelrc`? This would still require doing your own `gcloud auth` but that's what we already do in respect.

I think we need to return to the problem we're trying to solve so we can brainstorm tradeoffs of different approaches.

Ideally everyone can come and just run `bazel` and they can do what the need/want. And if they run `dazel` the only different will be it will use a Docker container (but otherwise work the same as `bazel`).

CodingCanuck:
> @CodingCanuck @rjhuijsman I'd like to revisit this approach because I don't want to add yet another build system into the mix.

+1! Also, you might've heard me mention that I don't like Makefiles :)

> 
> I'm assuming the reason we didn't do it the same way we do in [respect](https://github.com/reboot-dev/respect/blob/main/.bazelrc?rgh-link-date=2022-03-29T20%3A17%3A11Z#L6) is because not everyone has gcloud access?

I think you've exposed a bit of a mess that looks leftover from some churn on how we access the remote build cache.

The remote build cache used to be configured in the `Makefile`: https://github.com/reboot-dev/respect-integration/pull/15/files

Then we re-added it in `.bazelrc`: https://github.com/reboot-dev/respect-integration/commit/ce5c07042ef0deacd7399c027794d7bff5a8532f

Then we deduplicated by moving this config out of the `Makefile`: https://github.com/reboot-dev/respect-integration/commit/2a92c87f0186dc6751bfbfab54868fc0a01e7c99

Then we *re-added* it to the `Makefile` in order to support Codespaces, which don't have Google Application Default Credentials by default: https://github.com/reboot-dev/respect-integration/commit/efa00763c1a5a1ddc373fb7ef33e135f5e857e0a

Then we removed the original non-Makefile use of the .bazelrc by calling the relevant library (now `routing_hints.py`) directly from tests invoked via `make test`: https://github.com/reboot-dev/respect-integration/commit/12a74aa99f2e16d4e608ab3a1cc0f8f733aa32b7

At this point, I'm not sure if those lines you linked to work at all (or work as intended): at least on Codespaces, they might only work if you've first run a `make test`.

This PR is largely ripped from what respect's Makefile does: https://github.com/reboot-dev/respect/blob/main/Makefile#L109-L120


Note also that our respect/README.md specifies yet another (possibly outdated) method of setting credentials: https://github.com/reboot-dev/respect/#gcloud-preliminaries


So I'm not sure what the "blessed" way of getting GCP credentials is at this point.

> Is the idea then that those with gcloud access run `make test` and everyone else keeps running `bazel test`? That's not a future I want to encourage.

I agree that this split sucks. I don't have a good understanding of what we want, or what we want to encourage.

Do we want any non-org contributor to be able to use our build cache? RJ's been giving me the impression that our answer is "no" for security reasons (our GCP bucket is just ~random storage, so we probably don't want any internet user to have write-access to that).

If we don't want non-org contributors to be able to use our build cache, then I sure hope that we can use it for fast builds! I think we're then stuck with two ways of doing things. Then the open questions are around UX: can everyone run a single command and get the fastest build available for them (cache-enabled if possible, non-cached otherwise)? Can the command give obvious errors or hints for broken (using a cache you can't access) or sub-optimal (building cache-less when you have cache access) usage?

> 
> What about doing something similar to what we did with respect except telling folks to use their own `users.bazelrc`? This would still require doing your own `gcloud auth` but that's what we already do in respect.

This is worth discussing. At a minimum, I think we'd need additional config to set up the remote cache (you both need auth *and* flags to turn on caching): perhaps we can put opt-in / auth-required stuff into a .bazelrc-full-access file or similar that users import.

This is non-optimal for Codespaces, though, where you're regularly starting with a new machine that you haven't configured. We could probably improve that (e.g. with [dotfiles](https://docs.github.com/en/codespaces/customizing-your-codespace/personalizing-codespaces-for-your-account#dotfiles)) but it feels pretty penguiny.
> 
> I think we need to return to the problem we're trying to solve so we can brainstorm tradeoffs of different approaches.
> 
> Ideally everyone can come and just run `bazel` and they can do what the need/want. And if they run `dazel` the only different will be it will use a Docker container (but otherwise work the same as `bazel`).

Agreed, but I'd add the requirement that "it should be hard for users to use a non-optimal setup" (meaning users who have cache access should ideally use the cache by default, and it should be super obvious [via warnings or similar] when they're not using the cache, and they shouldn't be able to accidentally opt-out of caching [e.g. by forgetting to log in on a new codespace]).

WDYT is the best forum for improving things here? I can file an Eventuals issue, and maybe one on Respect too.

CodingCanuck:
@rjhuijsman pointed out offline that for the respect repo, we might currently be relying on the double declaration of cache-related flags: it might be that our `Makefile`-set cache flags are used when building envoy and our `.bazelrc` cache flags are used when building tests.

Both ways would rely on auth being set up before the cache is used.

Issue #289, opened 2022-03-29
Generated code is broken due to mismatch with hand-written `server.h`


rjhuijsman:
At [`eventuals/grpc/server.h:54`](https://github.com/3rdparty/eventuals/blob/35e1fcf4e197f0c3fa72179b19db4c6f0fd672cb/eventuals/grpc/server.h#L54) we find the following:
```cpp
class Service {
 public:
  // [...]
  virtual Task::Of<void>::Raises<std::runtime_error> Serve() = 0;
```

While since #282 the code generated by `protoc-gen-eventuals` looks as follows:
```cpp
class TypeErasedService : public ::eventuals::grpc::Service {
 public:
  ::eventuals::Task::Of<void>::Raises<std::exception> Serve() override;
```

The mismatch between `std::runtime_error` and `std::exception` causes compilation failures, such as in https://github.com/reboot-dev/respect/issues/230#issuecomment-1082196695:
```
In file included from ./object-store/object-store.h:8,
                 from object-store/main.cc:13:
bazel-out/k8-fastbuild/bin/external/com_github_reboot_dev_respect_api/respect/v1/object_store.eventuals.h:20:57: error: invalid covariant return type for 'virtual eventuals::_Task<void, void, std::tuple<> >::Raises<std::exception> respect::v1::eventuals::ObjectStore::TypeErasedService::Serve()'
   20 |     ::eventuals::Task::Of<void>::Raises<std::exception> Serve() override;
      |                                                         ^~~~~
In file included from object-store/main.cc:8:
external/com_github_3rdparty_eventuals/eventuals/grpc/server.h:54:54: note: overridden function is 'virtual eventuals::_Task<void, void, std::tuple<> >::Raises<std::runtime_error> eventuals::grpc::Service::Serve()'
   54 |   virtual Task::Of<void>::Raises<std::runtime_error> Serve() = 0;
      |                                                      ^~~~~
Target //object-store:image.tar failed to build
```

CodingCanuck:
Is it correct that this build error was found by another repo which included eventuals ( https://github.com/reboot-dev/respect/issues/230#issuecomment-1082196695 ) ?

Why wasn't this break caught by tests in the eventuals repo? I suspect this points to some missing test coverage: can you please file an issue for the tests we should have to catch breaks like this going forward?

rjhuijsman:
[...]

PR #292, opened 2022-03-29
Improve too-constraining static assert.

rjhuijsman:
Before this PR the static assert in `task.h` that checks that a given
`Task` only raises the errors specified in its `::Raises` clause didn't
accommodate the possiblity of specifying `::Raises<std::exception>` as
a catch-all.

This PR makes it possible to use that catch-all, allowing us to modify
`server.h` and `server.cc` to match the `std::exception` currently being
raised by generated code.

Fixes https://github.com/3rdparty/eventuals/issues/289

Tested: built `reboot-dev/respect:respect-object-store.tar`.

### DIFF
--- a/eventuals/grpc/server.cc
+++ b/eventuals/grpc/server.cc
@@ -115,7 +115,7 @@ Server::Server(
     serve->service->Register(this);
 
     serve->task.emplace(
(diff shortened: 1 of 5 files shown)


benh (approved):
🙏

Issue #293, opened 2022-03-30
Use fast bazel settings (e.g. remote cache) whenever possible, and make it difficult to use non-fast settings


CodingCanuck:
The current [eventuals build instructions](https://github.com/3rdparty/eventuals#buildingtesting) runs all builds locally. This is often slower (and ~never faster) than using the remote build cache that we've set up for this repo.

Users currently have to perform some undocumented steps in order to use the remote bazel cache for local builds. We should figure out how to give more users better build performance. Things to think about:

1) We probably can't make all build optimizations work for all users (e.g. we can't turn on world/anonymous write access to the GCP buckets which we use for bazel remote caches).
2) Builds should work without errors for users who aren't part of our organizations who check out our repositories
3) Frequent contributors to our codebases should not waste time on slow builds (meaning they should probably use build features like the bazel remote cache)
4) Users should have a hard time accidentally using non-optimal settings (e.g. if you can use the cache, you *should* use the cache)

It's unclear the best way to go about this. Maybe we could offer something like a one-time setup that users perform via something like a setup script to configure caches to be used (or not used), though we'd have to make sure that this setup happens before the first bazel build.

A big open question: how do other open source projects that use bazel handle this sort of bazel config, where some users use build features (e.g. bazel features like [remote caching](https://bazel.build/docs/remote-caching) and [remote execution](https://bazel.build/docs/remote-execution)) that are only available to an authorized subset of authenticated users?

CodingCanuck:
See longer discussion on https://github.com/3rdparty/eventuals/pull/287

CodingCanuck:
Some Googling leaves me empty handed: I can't find any examples of open source codebases that use bazel with a remote cache.

Some projects use it in what I think are CI environments ([example](https://github.com/GoogleCloudPlatform/kms-integrations/blob/ed71842310687b174b79cda58fcbec09b32873c2/.kokoro/windows_build.bat#L29))

One tiny project makes remote cache usage a `--config` (like `--config=asan`) and instructs users to manually add that to their personal `.bazelrc` file ([example](https://github.com/mattsoulanille/NovaJS/tree/6a5c0478a08e7e617185cdf7b16e87d3ab94cddd#optional-read-from-the-bazel-remote-cache))

[...]

PR #294, opened 2022-03-30
Revert "Add a `make test` command that uses a build cache"

CodingCanuck:
Reverts 3rdparty/eventuals#287

Per https://github.com/3rdparty/eventuals/issues/293, it's not clear that this addition of a `make` layer is the right way to speed up builds.


CodingCanuck:
@benh FYI

rjhuijsman (approved):

PR #298, opened 2022-04-01
Use a custom dazel name to avoid name collisions

CodingCanuck:
Progress towards https://github.com/reboot-dev/respect-integration/issues/142

### DIFF
--- a/.dazelrc
+++ b/.dazelrc
@@ -1,0 +1,20 @@
+#!/usr/bin/env python3
+# This configuration file is effectively a python module loaded by dazel.
+# Source: https://github.com/nadirizr/dazel/blob/master/dazel.py#L600-L601
+
+import os.path
+
+# Instead of using the host's Bazel cache (at ~/.cache/bazel), use a directory
+# that Dazel has to itself. This is essential to avoid a recurring problem with
+# compilation failures due to corrupt caches. Even Dazel's maintainers think
+# using the same cache is a mistake:
+#   https://github.com/nadirizr/dazel/issues/43
+DAZEL_BAZEL_USER_OUTPUT_ROOT = os.path.expanduser("~/.cache/dazel/tests")
+
+DAZEL_DOCKERFILE = "Dockerfile.dazel"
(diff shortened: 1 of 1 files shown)


rjhuijsman (approved):

.dazelrc
```diff
@@ -0,0 +1,20 @@
...
+# Source: https://github.com/nadirizr/dazel/blob/master/dazel.py#L600-L601
+
+import os.path
+
+# Instead of using the host's Bazel cache (at ~/.cache/bazel), use a directory
+# that Dazel has to itself. This is essential to avoid a recurring problem with
+# compilation failures due to corrupt caches. Even Dazel's maintainers think
+# using the same cache is a mistake:
+#   https://github.com/nadirizr/dazel/issues/43
+DAZEL_BAZEL_USER_OUTPUT_ROOT = os.path.expanduser("~/.cache/dazel/tests")
```
benh, line 12: Is `tests` here necessary? If so, can we document why? Otherwise can we just do `~/.cache/dazel/tests`?

.dazelrc
```diff
@@ -0,0 +1,20 @@
...
+
+DAZEL_DOCKERFILE = "Dockerfile.dazel"
+DAZEL_VOLUMES = ["/var/run/docker.sock:/var/run/docker.sock"]
```
benh, line 15: Can we leave a comment as to why we need/want to expose the host's Docker? IIUC the Ubuntu build which uses `dazel` has been working before this so it's not clear why we need to add it now and a would be great!

benh (commented):
Was catching up on merged PRs and had a few questions on this one!

.dazelrc
CodingCanuck: You removed this in https://github.com/3rdparty/eventuals/pull/314 ! :D

.dazelrc
CodingCanuck: Nope, and you cleaned this up in https://github.com/3rdparty/eventuals/pull/315 ! :D

CodingCanuck (commented):
Closing out old comments for posterity

PR #301, opened 2022-04-02
Add basic compilation tests for generated code

CodingCanuck:
Fixes https://github.com/3rdparty/eventuals/issues/291

As expected, this catches when backwards-incompatible code-gen updates
(like https://github.com/reboot-dev/pyprotoc-plugin/commit/ad976383919a1d96e6d1e6a2c86c0ff95623c8ee)
fail to compile. Nice!

PR #302, opened 2022-04-02
Update to latest `pyprotoc-plugin`

CodingCanuck:
This required upgrading from bazel 5.0.0 to 5.1.0 to pick up the
[new starlark strings functions](https://blog.bazel.build/2022/03/24/bazel-5.1.html#starlark).


CodingCanuck:
Formatting fixes are in https://github.com/reboot-dev/pyprotoc-plugin/pull/16

rjhuijsman:
Looks like this also introduces a test service? Is that intentional?

It also looks like this uncovers an eventuals-related build breakage? Is there an issue for that yet?

CodingCanuck:
TL;DR let's do a quick sync after your meeting to catch you up on the cross-repo juggling here.

> Looks like this also introduces a test service? Is that intentional?
> 
> It also looks like this uncovers an eventuals-related build breakage? Is there an issue for that yet?

No to all of the above.

I'm currently basing this PR on top of https://github.com/3rdparty/eventuals/pull/301 : *that* PR adds a test service, that's intentional, and it fails to build for reasons being discussed in that other PR.

This PR modifies the call to `cc_eventuals_library` used by the test service (this is to handle the pyprotoc-plugin API change I made in https://github.com/reboot-dev/pyprotoc-plugin/pull/13 : this was expected. It's part of https://github.com/reboot-dev/respect-api/issues/17 . I didn't file an issue to fix the break this will cause in other repos by updating them to this new format: I could file one issue per repo, but we'd need to file 4-5 issues [see list](https://github.com/reboot-dev/respect/issues/266#issuecomment-1086542590) and that doesn't seem high ROI)

This call to `cc_eventuals_library` doesn't exist at HEAD / without PR 301: once this PR is approved, I'll just rebase it onto main if 301 isn't approved yet.

while-false (approved):
LGTM! :tada:

CodingCanuck:
> TL;DR let's do a quick sync after your meeting to catch you up on the cross-repo juggling here.
> 
> > Looks like this also introduces a test service? Is that intentional?
> > It also looks like this uncovers an eventuals-related build breakage? Is there an issue for that yet?
> 
> No to all of the above.
> 
> I'm currently basing this PR on top of #301 : _that_ PR adds a test service, that's intentional, and it fails to build for reasons being discussed in that other PR.
> 
> This PR modifies the call to `cc_eventuals_library` used by the test service (this is to handle the pyprotoc-plugin API change I made in [reboot-dev/pyprotoc-plugin#13](https://github.com/reboot-dev/pyprotoc-plugin/pull/13) : this was expected. It's part of [reboot-dev/respect-api#17](https://github.com/reboot-dev/respect-api/issues/17) . I didn't file an issue to fix the break this will cause in other repos by updating them to this new format: I could file one issue per repo, but we'd need to file 4-5 issues [see list](https://github.com/reboot-dev/respect/issues/266#issuecomment-1086542590) and that doesn't seem high ROI)
> 
> This call to `cc_eventuals_library` doesn't exist at HEAD / without PR 301: once this PR is approved, I'll just rebase it onto main if 301 isn't approved yet.

https://github.com/3rdparty/eventuals/pull/301 is still pending: rebasing onto main and merging with @while-false 's LGTM

PR #312, opened 2022-04-06
Use typed input for enabling tmate debugging

benh:
Before this commit we used a string which required users to know they needed to add `true` or `false` and instead a typed input should make this more user friendly.


CodingCanuck (approved):
clap

[The new one is much better](https://www.youtube.com/watch?v=dBqhIVyfsRg&ab_channel=cheedogg).

PR #313, opened 2022-04-07
Remove useless colon

benh:
GitHub renders the checkbox to the left of the description text which makes the colon useless. But more importantly, we probably shouldn't rely on the rendering in case they change it in the future!


CodingCanuck (approved):
Please ensure that respect benchmarks are unchanged by this PR

Also rerun integration tests

Also I'll need a TPS report detailing the consumer-facing impact of this PR on my desk by tomorrow 8am

PR #314, opened 2022-04-07
Remove unnecessary DAZEL_VOLUMES

benh:
Perhaps we'll need to add this in the future if/when we create images and specifically want to use a Docker that is outside of the running Dazel container, but for now we're removing this so that it doesn't cause an unexpected bug or send people in the wrong direction when debugging.


CodingCanuck (approved):
Thanks, this is strictly simpler!

PR #315, opened 2022-04-07
Use a different directory for Bazel caching

benh:
Also added a comment explaining why we picked the directory we did and why it's best effort.


CodingCanuck (approved):
FYI, the "multiple clones" thing is the most central sales pitch for Codespaces for me. (It's also why I lean towards often firing up single-use Codespaces)

PR #317, opened 2022-04-07
Type erase `Task` for `Serve` withoud any `Raises`

onelxj:


eventuals/grpc/server.h
```diff
@@ -746,8 +747,10 @@ auto UnaryEpilogue(ServerCall<Request, Response>& call) {
               return ::grpc::Status(::grpc::UNKNOWN, e.what());
             })
       | Then([&](auto&& status) {
-           return call.Finish(status)
-               | call.WaitForDone();
+           return call.Finish(status);
+         })
+      | Finally([&](auto&& expected) {
+           return call.WaitForDone();
          });
```
benh, line 754: Fix up here and below please! By embedding the `Finally()` we are more easily able to catch mistakes where by accident an error would propagate and we haven't properly done a catch, because we only know about the error propagating from `Finish()` so let's bind the catch closely!

```suggestion
           return call.Finish(status)
               | Finally([&](std::optional<std::exception_ptr>&&) {
                    return call.WaitForDone();
                });
```

Note that we really should improve the `Finally()` API to not give us an exception_ptr but instead to give us the equivalent of doing a `Catch().raised<std::exception>())` so that we get a `std::optional<std::exception>` so that we can easily just do an `exception->what()` to log the error message. Can you update `Finally()` in a different review please? Thanks!

eventuals/grpc/server.h
```diff
@@ -767,8 +770,10 @@ auto StreamingEpilogue(ServerCall<Request, Response>& call) {
               return ::grpc::Status(::grpc::UNKNOWN, e.what());
             })
       | Then([&](auto&& status) {
-           return call.Finish(status)
-               | call.WaitForDone();
+           return call.Finish(status);
+         })
+      | Finally([&](auto&& expected) {
+           return call.WaitForDone();
```
benh, line 776: Fix up this one too please!

benh, line 776: Fix up this one too please!

test/helloworld.eventuals.cc
```diff
@@ -40,7 +40,7 @@ Task::Of<void>::Raises<std::exception> Greeter::TypeErasedService::Serve() {
                      return UnaryPrologue(call)
                          | Then(Let([&](auto& request) {
                               return Then(
-                                  [this,
+                                  [&,
```
benh, line 43: Capturing `this` is explicitly _less_ so I'd prefer it! It's not clear why you needed to change this?

benh, line 43: Capturing `this` is explicitly _less_ so I'd prefer it! It's not clear why you needed to change this?

benh (approved):

eventuals/finally.h
```diff
@@ -13,6 +13,9 @@ namespace eventuals {
 struct _Finally final {
   template <typename K_, typename Arg_>
   struct Continuation final {
+    Continuation(K_&& k)
+      : k_(std::move(k)) {}
+
```
benh, line 18: Necessary or just from your debugging?

eventuals/raise.h
```diff
@@ -17,7 +19,12 @@ struct _Raise final {
 
     template <typename... Args>
     void Start(Args&&...) {
-      k_.Fail(std::move(t_));
+      if constexpr (std::is_same_v<T_, std::optional<std::exception_ptr>>) {
+        CHECK(t_.has_value());
+        k_.Fail(std::move(t_.value()));
+      } else {
+        k_.Fail(std::move(t_));
+      }
```
benh, line 27: ???
```suggestion
```

eventuals/grpc/server.h
```diff
@@ -766,9 +780,21 @@ auto StreamingEpilogue(ServerCall<Request, Response>& call) {
...
               return ::grpc::Status(::grpc::UNKNOWN, e.what());
             })
-      | Then([&](auto&& status) {
+      | Then([&](::grpc::Status status) {
            return call.Finish(status)
-               | call.WaitForDone();
+               | Finally([&](std::optional<std::exception_ptr>&& e) {
+                    return If(e.has_value())
+                               .yes(
+                                   Raise(std::move(e))
```
benh, line 788: But you shouldn't be able to raise a `std::optional`!!!

eventuals/grpc/server.h
```diff
@@ -745,9 +747,21 @@ auto UnaryEpilogue(ServerCall<Request, Response>& call) {
...
+                                   | Catch()
+                                         .raised<std::exception>(
+                                             [](std::exception&& error) {
+                                               EVENTUALS_GRPC_LOG(1)
+                                                   << error.what();
```
benh, line 760: Nicer error message please!

test/finally.cc
```diff
@@ -127,3 +134,36 @@ TEST(Finally, VoidStop) {
       std::rethrow_exception(exception.value()),
       eventuals::StoppedException);
 }
+
+TEST(Finally, FinallyInsideThen) {
```
benh, line 138: Test the path that goes through the `Raise` please!!!

benh (changes requested):

test/finally.cc
```diff
@@ -127,3 +134,37 @@ TEST(Finally, VoidStop) {
...
+                                                       error.what(),
+                                                       "error");
+                                                 });
+                                 })
+                                 .no([]() {
+                                   return Raise("another error");
+                                 })
+                          | Eventual<void>([](auto& k) {
+                               k.Start();
+                             });
```
benh, line 164: Is this necessary?

benh (approved):

PR #320, opened 2022-04-11
Use callables instead of eventuals for 'If.yes()' and 'If.no()'

benh:
Because most folks are not as used to expression based langauges they
run into issues where they end up doing things in the same eventuals
pipeline expression that they weren't expecting because they were
thinking like it was multiple statements. This is true of 'If.yes()'
and 'If.no()'. By making these take callables (e.g., a lambda) instead
then they are forced to treat the bodies of each not as expressions
and thus will be required to think (a little more) about how to
'std::move()' or copy values.

### DIFF
--- a/test/if.cc
+++ b/test/if.cc
@@ -27,8 +27,8 @@ TEST(IfTest, Yes) {
     return Just(1)
         | Then([](int i) {
              return If(i == 1)
-                 .yes(Just("yes"))
-                 .no(Just("no"));
+                 .yes([]() { return Just("yes"); })
+                 .no([]() { return Just("no"); });
            });
   };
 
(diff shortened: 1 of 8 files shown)


ArthurBandaryk (approved):

test/if.cc
```diff
@@ -117,8 +119,8 @@ TEST(IfTest, Raise) {
     return Just(1)
         | Then([](int i) {
              return If(i == 1)
-                 .yes(Just(i))
-                 .no(Raise("raise"));
+                 .yes([&]() { return Just(i); })
```
onelxj, line 122: What if we send `i` as an argument to this function, so we can `move` some args to eventuals

onelxj (commented):
What if we propagate args from `Start` to `Then` it might be helpful in some cases and we keep it flexible by using 
`if constexpr (std::is_invocable<F>){k.Start()}
else {k.Start(std::forward<Args>(args)...)}
`

PR #321, opened 2022-04-11
Move concurrent tests into a subfolder

CodingCanuck:


test/concurrent/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
+# We build "concurrent" tests in a separate target to significantly speed up
+# linking on platforms which prefer shared libraries (e.g., macos).
+
+load("@rules_cc//cc:defs.bzl", "cc_test")
+load("//bazel:copts.bzl", "copts")
+
+cc_test(
+    name = "test_concurrent",
```
benh, line 8: We run all the `eventuals` tests directly via the gtest binary for easily using the debugger so please keep this as the `concurrent-tests` library and re-add as a dependency of the eventuals `cc_test`! Plan was to eventually integrate the grpc tests too but given some of the current flakyness we just haven't gotten to it yet.

benh (approved):

test/concurrent/BUILD.bazel
CodingCanuck: Can you help me understand this: why does having everything in a single cc_test make it easier to use the debugger?

(Is there something about our debugger setup that makes it hard to work with multiple tests? This sounds highly fixable, and a kitchen sync binary [tests are just fancy binaries] that links in every library is an anti-pattern: separate tests build faster and can get better bazel cache utilization)

benh: (1) Chaos testing: when we are using the debugger it's almost always to sort out some non-deterministic issues from running the tests _repeatedly_ and with _shuffle_ because some internal part of eventuals is broken. These bugs are far more difficult (if not impossible) to trigger we run every test independently. For now I sleep better at night knowing we are injecting _some_ chaos into our testing to help trigger any bugs.

(2) Streamlining editor breakpoint debugging: some editors allow you to configure how to launch a binary for debugging in order to add debug symbols, such as Visual Studio Code (here's an [example](https://stackoverflow.com/questions/57825915/how-to-debug-bazel-c-google-test-in-visual-studio-code)). In my experience it's nice to be able to have a singular one of these for all the tests where you can quickly add breakpoints and then just run _debug_ without tracking down which test landed in which binary and then updating the `launch.json` appropriately or figuring out which `launch.json` to run.

The build time tradeoff right now has been very manageable when doing iterative `printf` style debugging (which I'm not encouraging, it's just the most resource intensive approach so it's a good proxy for the build time pain or lack thereof). And to be clear, I'm very much in favor of separate libraries as this can very much improve build times.

I'm happy for us to consider alternative approaches, but let's not change how we're doing this until we've had those discussions please!

CodingCanuck: TL;DR this sounds like we're optimizing for the wrong thing such that the tail is wagging the dog.

But to your point, let's separate out this discussion from this code move: switched this to a `cc_library` target that's a dependency of `//test:eventuals`.

> (1) Chaos testing: when we are using the debugger it's almost always to sort out some non-deterministic issues from running the tests _repeatedly_ and with _shuffle_ because some internal part of eventuals is broken. These bugs are far more difficult (if not impossible) to trigger when we run every test independently. For now I sleep better at night knowing we are injecting _some_ chaos into our testing to help trigger any bugs.

Have we encountered any bugs where one test affects another? That would require some sort of shared state across tests, and I though that eventuals was entirely functional/stateless. Specifically, there's nothing like a singleton or any shared memory where eventuals test A could affect eventuals test B.

If it's correct that all eventuals implementations are ~pure functional, then I don't see how combining tests "adds chaos" or increases coverage. *Repeating* tests, sure (to flush out things like race condition), but repeating individual tests should be as effective as repeating combined/mega-tests.

Is my understanding here correct? I suspect I'm missing something here since I'm still quite unfamiliar with eventuals.

But if it helps you sleep, then sure!

> 
> (2) Streamlining editor breakpoint debugging: some editors allow you to configure how to launch a binary for debugging in order to add debug symbols, such as Visual Studio Code (here's an [example](https://stackoverflow.com/questions/57825915/how-to-debug-bazel-c-google-test-in-visual-studio-code)). In my experience it's nice to be able to have a singular one of these for all the tests where you can quickly add breakpoints and then just run _debug_ without tracking down which test landed in which binary and then updating the `launch.json` appropriately or figuring out which `launch.json` to run.

Ah, this makes sense: I'll admit that I haven't fired up a debugger in ages, though I admit that I remember Visual Studio's debugger being absolutely top notch a decade ago! I also very likely wasn't working in a multi-binary world in that era.

Internally at Google, there's very little support for debuggers. I initially found this to be an enormous culture clash: didn't all of these SWEs realize what useful debuggers are out there (especially the Visual Studio debugger)?!? But I slowly came around to the style that Google preferred: most of the time (not all of the time), debuggers help you identify your current bug quickly, but that's non-investment work (firing up that debugger today doesn't help you with next week's novel bug). In contrast, writing more tests to catch more bugs (or adding new test infrastructure like `asan`, `msan`, or `tsan`) is *investment* work: that work *does* help you catch next week's novel bug.

In particular, it seems that the majority of bugs produced by "chaos testing" are readily caught by `asan`/`msan`/`tsan` in a way that doesn't require the use of a debugger. Sanitizers as a specific investment are also often more readable debugging aids than debuggers (debuggers allow you to inspect all current state, sanitizers highlight *bad* state), and they can be much faster, too (e.g. if we run all tests under asan by default, or run msan/tsan on Actions Runners, you get a detailed error before you thought to fire up the debugger!)

So I definitely agree that it's a good goal to make debugging easy and ergonomic. It might be most productive to prioritize good, fast, comprehensive test coverage more highly than easy debugger use, though, and those might sometimes pull in different directions (e.g. well divided tests == more concise log output == faster to find the error message that points you directly to a bug).

> 
> The build time tradeoff right now has been very manageable when doing iterative `printf` style debugging (which I'm not encouraging, it's just the most resource intensive approach so it's a good proxy for the build time pain or lack thereof). And to be clear, I'm very much in favor of separate libraries as this can very much improve build times.
> 
> I'm happy for us to consider alternative approaches, but let's not change how we're doing this until we've had those discussions please!

Issue #274, opened 2022-03-28
`EventualsGrpcTest.ServerDeathTest` is flaky


CodingCanuck:
Sample flaky failure: https://github.com/3rdparty/eventuals/runs/5727035543?check_suite_focus=true

```
[ RUN      ] EventualsGrpcTest.ServerDeathTest

Warning:  external/com_github_google_googletest/googletest/src/gtest-death-test.cc:1121:: Death tests use fork(), which is unsafe particularly in a threaded context. For this test, Google Test detected 2 threads. See https://github.com/google/googletest/blob/master/docs/advanced.md#death-tests-and-threads for more explanation and suggested solutions, especially if this is the last message you see before your test times out.
W20220328 20:39:18.128109 328424960 terminal.h:282] WARNING: exception thrown while dereferencing eventual: Failed to write
unknown file: Failure
C++ exception with description "Failed to write" thrown in the test body.
[  FAILED  ] EventualsGrpcTest.ServerDeathTest (37 ms)
```

I'm not sure if the `failed to write` error is related to the warning about `fork()`, but I wonder:
1) Do we need to use fork here? (Can we instead manually start a non-forked server subprocess, then kill that by e.g. sending it a SIGTERM?)
2) What is this test testing? Does the server need to actually die in order to test the interesting logic?

CodingCanuck:
https://github.com/3rdparty/eventuals-grpc/issues/51 might have some historical context (though it might also be a different issue with a similar symptom)

CodingCanuck:
Anecdotally, this seems to be happening regularly on macos-lates, e.g.: https://github.com/3rdparty/eventuals/runs/5730033688?check_suite_focus=true

PR #322, opened 2022-04-11
Rewrite ServerDeathTest to avoid fork()

CodingCanuck:
I'm hoping this should resolve the flaky failures seen in
https://github.com/3rdparty/eventuals/issues/274. At the very least,
using subprocesses instead of threads, making process lifetime explicit
(rather than relying on gtest's death test infrastructure), and ensuring
that an RPC succeeds before terminating the server should make failures
easier to debug.

Should fix https://github.com/3rdparty/eventuals/issues/274.

### DIFF
--- a/test/grpc/death-server.cc
+++ b/test/grpc/death-server.cc
@@ -1,5 +1,12 @@
+// A dummy server used in tests.
+// Successfully responds to the 1st RPC then exits with an error on the 2nd.
+
+#include <iostream>
+
 #include "eventuals/grpc/server.h"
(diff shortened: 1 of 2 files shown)


CodingCanuck:
Per offline discussion: closing in favor of https://github.com/3rdparty/eventuals/pull/338 (we think this fixes the flaky test error) and https://github.com/3rdparty/eventuals/pull/339 (test refactor to use `std::system` to both make this test windows-compatible and to avoid warnings from multithreaded use of `ASSERT_DEATH`).

Issue #275, opened 2022-03-28
test/streaming.cc takes minutes to compile


CodingCanuck:
This file: [https://github.com/3rdparty/eventuals-grpc/blob/main/test/streaming.cc](https://github.com/3rdparty/eventuals-grpc/blob/main/test/streaming.cc?rgh-link-date=2022-02-17T00%3A52%3A23Z)

Takes several minutes to compile as part of this build:

```
bazel test //test:grpc
```

Sample output from a not-yet-done bazel build with default compilation option (5 minutes and counting!):

```
[22 / 24] Compiling test/streaming.cc; 307s processwrapper-sandbox
```

My wild guess as to what's going on: we have some long eventual chains in this file. Does that result in slow compilation times due to heavy template nesting? If so, here's a resource on how to profile time spent in the compiler: [stackoverflow.com/questions/15818281/profiling-template-metaprogram-compilation-time](https://stackoverflow.com/questions/15818281/profiling-template-metaprogram-compilation-time)

It looks like we're building using GCC. I wonder if Clang's any faster?

(This issue is a recreated version of https://github.com/3rdparty/eventuals-grpc/issues/73 now that eventuals-grpc is a read-only repo)

CodingCanuck:
From an offline discussion with @benh : a likely hypothesis is that the long eventuals chains in streaming.cc are indeed causing slow compilation times.

One workaround he proposes is to split streaming.cc into many files (one file per test) so there's at most one long eventual per test (a similar thing was done for eventuals tests in https://github.com/3rdparty/eventuals/pull/182 ).

CodingCanuck:
Compiling the output files generated by `cc_eventuals_library` also takes a super long time, e.g.: https://github.com/reboot-dev/respect/runs/5839858123?check_suite_focus=true

```
[1,964 / 2,126] Compiling external/com_github_reboot_dev_respect_api/respect/v1/object_store.eventuals.cc; 312s local, remote-cache ... (4 actions running)
```

CodingCanuck:
Hmm, I'm no longer able to reproduce these multi-minute compile times.

clang (3 runs):
```
3 runs of:
echo "" >> test/streaming.cc && time dazel build --config=clang //test:grpc

real    0m30.302s
user    0m0.143s
sys     0m0.062s

real    0m29.914s
user    0m0.128s
sys     0m0.083s

real    0m30.148s
user    0m0.123s
sys     0m0.086s
```

gcc (3 runs):
```
[...]

PR #323, opened 2022-04-11
Compile with clang by default

CodingCanuck:
This should speed up template compilation times and improve
https://github.com/3rdparty/eventuals/issues/275.

This also ensures that clang is used on all platforms. Before this PR,
our CI servers were building with gcc.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -6,13 +6,20 @@ build --enable_platform_specific_config
 # Verbose failure logs when something goes wrong.
 build --verbose_failures
 
+# Build with clang (rather than the default of gcc) since clang compiles
+# templates faster and produces better error messages.
+# Prefer repo_env over action_env to avoid invalidating the action graph:
+# https://github.com/bazelbuild/bazel/issues/8074
+build --repo_env=CC=clang
+
 # Use Clang as a default compiler on Windows.
-# TODO: use Clang as a default compiler on every platform.
+# TODO(xander): Does repo_env above cover the windows case?
 build:windows --compiler="clang-cl"
 
-# Compiler options.
-build:gcc --action_env=CC=gcc
-build:clang --action_env=CC=clang
+# Make it easy to use gcc.
+# Prefer repo_env over action_env to avoid invalidating the action graph:
+# https://github.com/bazelbuild/bazel/issues/8074
+build:gcc --repo_env=CC=gcc
 
 # Build eventuals in C++17 mode.
 build:linux --cxxopt=-std=c++17
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -116,7 +116,6 @@ jobs:
             ${{ matrix.bazel-config }} \
             ${{ env.BAZEL_REMOTE_CACHE }} \
             --verbose_failures \
-            --spawn_strategy=local \
             -c dbg \
             --strip="never" \
             ...
@@ -130,7 +129,6 @@ jobs:
             --experimental_ui_max_stdouterr_bytes=-1 \
             ${{ env.BAZEL_REMOTE_CACHE }} \
             --verbose_failures \
-            --spawn_strategy=local \
             -c dbg \
             --strip="never" \
             --test_output=errors \


.bazelrc
```diff
@@ -3,9 +3,12 @@ try-import submodules/dev-tools/.bazelrc
...
 build --enable_platform_specific_config
 
-build:windows --compiler="clang-cl"
+# Build with clang (rather than the default of gcc) since clang compiles
+# templates faster and produces better error messages.
+build --action_env=CC=clang
 
-build:clang --action_env=CC=clang
+# Make it easy to use gcc.
+build:clang --action_env=CC=gcc
```
FolMing, line 11: Shouldn't that be `build:gcc`?

CodingCanuck: :/ absolutely, yep.

Done.

.bazelrc
```diff
@@ -3,9 +3,12 @@ try-import submodules/dev-tools/.bazelrc
 # Specific Bazel build/test options.
 build --enable_platform_specific_config
 
-build:windows --compiler="clang-cl"
```
FolMing, line 6: Is there a specific reason why you have removed this line?
It looks like it's now compiling with MSVC.

CodingCanuck: Reverted, and tested that the current settings use clang on windows. (Though to be fair I didn't do a similar test removing this line)

.bazelrc
```diff
@@ -6,13 +6,16 @@ build --enable_platform_specific_config
...
 # Use Clang as a default compiler on Windows.
-# TODO: use Clang as a default compiler on every platform.
+# TODO(xander): Does repo_env above cover the windows case?
 build:windows --compiler="clang-cl"
 
-# Compiler options.
-build:gcc --action_env=CC=gcc
-build:clang --action_env=CC=clang
+# Make it easy to use gcc.
+build:gcc --repo_env=CC=gcc
```
FolMing, line 18: Can you please leave a comment why we have to use `--repo_env` instead of `--action_env`?

CodingCanuck: Done.

FolMing (commented):
`.bazelrc` file looks ok.
I'm not sure about the workflow changes, so either let's wait for @rjhuijsman's review or pls walk me through them (preferably in Gather).

CodingCanuck:
> `.bazelrc` file looks ok. I'm not sure about the workflow changes, so either let's wait for @rjhuijsman's review or pls walk me through them (preferably in Gather).

Happy to do either one!

I'll likely be working on something like CEST / German time for the next few weeks, feel free to grab me whenever (unless RJ reviews this first!)

.github/workflows/build_and_test.yml
```diff
@@ -90,6 +90,28 @@ jobs:
...
+            || true
+
+      # Clean up any artifacts from previous builds to ensure that we aren't
+      # mixing state across novel and old builds. This is especially important
+      # if new builds use different compiler settings than old builds.
+      - name: Remove Dazel cache
+        if: ${{ matrix.bazel-command == 'dazel' }}
+        run: |
+          sudo rm -rf ~/.cache/dazel
+
```
rjhuijsman, line 114: I'm surprised that this is necessary. This is the `eventuals` repository, and it uses only GitHub-hosted Action Runners. Those should be hermetic. Why is it necessary to stop the Dazel instances and remove the cache manually?

CodingCanuck: Excellent question! It's not necessary at all: I tried several magic incantations and failed to isolate what the fix actually is.

You're right, the cleanup is unnecessary because each actions runner should run in its own VM (per some reading I did, though I didn't find a super strong conclusion on this).

What actually matters is removing the `--spawn_strategy=local` bazel flag. Per https://bazel.build/docs/user-manual#execution-strategy , the `local` flag runs compilation steps as a local subprocess (thereby accessing whatever files / libraries are available locally). The default when not specifying a flag is to instead run in some sort of sandbox which restricts where it can look for libraries. I'm not sure why we were previously preferring the local spawn strategy, and since removing it fixes the build let's do that :)

FolMing: > the local flag runs compilation steps as a local subprocess (thereby accessing whatever files / libraries are available locally). The default when not specifying a flag is to instead run in some sort of sandbox which restricts where it can look for libraries.

Huh? That doesn't sound right. There's no way it just stops accessing the local libraries. There should be something else.

FolMing (approved):

PR #326, opened 2022-04-11
Add a unit test for `Then<void>()`

CodingCanuck:


test/then.cc
```diff
@@ -42,6 +44,20 @@ TEST(ThenTest, Succeed) {
...
+  bool ran = false;
+  auto e = [&]() {
+    return Just()
+        | Then([&]() {
+             ran = true;
+           });
+  };
+
+  // A Then<void>() doesn't return anything.
+  static_assert(std::is_same_v<decltype(*e()), void>);
```
benh, line 57:
```suggestion
  static_assert(std::is_void_v<decltype(*e())>);
```

test/then.cc
```diff
@@ -42,6 +44,20 @@ TEST(ThenTest, Succeed) {
...
+TEST(ThenTest, SucceedVoid) {
+  bool ran = false;
+  auto e = [&]() {
+    return Just()
+        | Then([&]() {
+             ran = true;
+           });
+  };
+
+  // A Then<void>() doesn't return anything.
```
benh, line 56: So as not to confuse folks that `Then` can take a `void` type parameter, can I suggest something like:
```suggestion
  // Passing a callable (e.g., a lambda) that doesn't return anything to 'Then()' also doesn't return anything.
```

benh (approved):

test/then.cc
CodingCanuck: Done, with a phrasing tweak:

```
  // Then() doesn't return anything when its callable (in this case, its
  // lambda) doesn't return anything.
```

PR #330, opened 2022-04-11
BUILD cleanup: one BUILD file per directory

CodingCanuck:
This compartmentalizes our packages, making each `BUILD.bazel` file
smaller, making cross-package dependencies more explicit, and follows
best practice from https://bazel.build/rules/build-style#recursive :

> It is good practice to author a BUILD file in each directory and define a dependency graph between them.


benh (approved):
🤙

PR #335, opened 2022-04-18
Introduce Submodules-sync for `eventuals`

nizamutdinovb:
This PR is going to cover the issue https://github.com/reboot-dev/company/issues/65
However, the workflow won't work until token for Rebot account is added to the `3rdparty` organization secrets as it is done for `reboot-dev` org.

@CodingCanuck @rjhuijsman could you please add Rebot token to the '3rdparty' organization secrets as PRIVATE_REPO_ACCESS_AS_REBOT_TOKEN. I would add it by myself, but I don't know its value.
Thank you.


CodingCanuck (approved):
Nice!

PR #339, opened 2022-04-18
Use standard C++ to run subprocesses: avoid posix

### LINKED ISSUE #337: Avoid `execl` in tests for windows compatibility
We use `execl()` in two unit tests:

https://github.com/3rdparty/eventuals/blob/23ff9177164816b744e7d2905d07ac21ee4fcc26/test/grpc/server-death-test.cc#L63

https://github.com/3rdparty/eventuals/blob/23ff9177164816b744e7d2905d07ac21ee4fcc26/test/grpc/client-death-test.cc#L67

This isn't windows-compatible, since `execl()` is a non-standard / `posix` function.

Let's rewrite these tests with the following pattern:
- start a `std::thread` (tests already do this)
- call `std::system` as a cross-platform way to launch the subprocess (today, tests instead call `ASSERT_DEATH` which emits warnings when invoked from a thread)
- that thread will block until the server completes
- that thread should EXPECT / assert on the server's return code, which should be a well-known return code emitted by the server when dying
- the test should then cleanly join the thread (today, the tests detach the thread)

CodingCanuck:
Use cross-platform standard C++ to run subprocesses. This is an
incremental fix for building and running tests on windows. It also
avoids warnings emitted by `ASSERT_DEATH` caused by `fork()`-ing from a
multithreaded process.

Fixes #337.

### DIFF
--- a/test/grpc/death-client.cc
+++ b/test/grpc/death-client.cc
@@ -2,6 +2,7 @@
 #include "eventuals/terminal.h"
 #include "eventuals/then.h"
 #include "examples/protos/helloworld.grpc.pb.h"
+#include "test/grpc/death-constants.h"
 
 using helloworld::Greeter;
 using helloworld::HelloReply;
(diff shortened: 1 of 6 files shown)


test/grpc/death-server.cc
```diff
@@ -73,3 +51,24 @@ int main(int argc, char** argv) {
...
+// used to send the bound port of the gRPC server.
+//
+// See 'server-death-test.cc' for more details.
+int main(int argc, char** argv) {
+  // TODO(benh): use stout-flags!
+  CHECK_EQ(argc, 2) << "expecting 'pipe' to be passed as an argument";
+
+  int pipe = atoi(argv[1]);
+
+  ::eventuals::grpc::RunServer(pipe);
```
benh, line 72: Similar to the comment earlier, please don't put tests/functions in the `eventuals::grpc` namespace since that's not currently consistent with the rest of the codebase (only made the comment here, but same for the client death test).

test/grpc/death-client.cc
```diff
@@ -2,76 +2,49 @@
...
   };
 
   *call();
+}
+} // namespace
+} // namespace eventuals::grpc
+
+// Should only be run from tests!
+//
+// Expects two arguments.
```
benh, line 37:
```suggestion
// Expects one arguments.
```

benh (approved):
This is a great improvement! Thank you!

CodingCanuck (commented):
Thanks for the high quality and thoughtful review!

PR #341, opened 2022-04-19
Introduce an 'Executor' for "fire and forget" tasks

benh:
See all commits as some clean up commits along the way.

### DIFF
--- a/eventuals/executor.h
+++ b/eventuals/executor.h
@@ -1,0 +1,119 @@
+#pragma once
+
+#include "eventuals/catch.h"
+#include "eventuals/concurrent.h"
+#include "eventuals/eventual.h"
+#include "eventuals/interrupt.h"
+#include "eventuals/just.h"
+#include "eventuals/lock.h"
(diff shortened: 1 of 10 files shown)


eventuals/executor.h
```diff
@@ -0,0 +1,119 @@
...
+  auto Wait() {
+    return TypeCheck<void>(
+        Synchronized(
+            wait_until_finished_.Wait(this->Borrow([this]() {
+              return /* wait = */ !finished_;
+            }))));
+  }
+
+ private:
+  std::string name_;
```
ArthurBandaryk, line 107: Just for my knowledge: why should we store this name?:)

ArthurBandaryk (approved):
Looks great to me! So you can merge it. Anyway later I will run with debugger and logs to understand much more better! Thanks!

eventuals/executor.h
benh: Because we need the other members from `Executor` so it's easiest to just capture `this`.

PR #342, opened 2022-04-19
Introduce support for polling a file descriptor

benh:


eventuals/event-loop.h
```diff
@@ -706,6 +722,274 @@ class EventLoop final : public Scheduler {
...
+        using Tuple = std::tuple<decltype(this), Error>;
+        auto tuple = std::make_unique<Tuple>(
+            this,
+            std::forward<Error>(error));
+
+        // Submitting to event loop to avoid race with interrupt.
+        loop_.Submit(
+            this->Borrow([tuple = std::move(tuple)]() {
+              std::apply(
+                  [](auto* continuation, auto&&... args) {
```
onelxj, line 881: Does it make sense `(auto* continuation, Error&&... args)`?

eventuals/event-loop.h
```diff
@@ -706,6 +722,274 @@ class EventLoop final : public Scheduler {
...
+                  std::move(*tuple));
+            }),
+            &context_);
+      }
+
+      void Stop() {
+        // Submitting to event loop to avoid race with interrupt.
+        loop_.Submit(
+            this->Borrow([this]() {
+              k_.Stop();
```
onelxj, line 894: Should be `CHECK(!closed_ && !completed_)` there?`

onelxj (approved):

PR #344, opened 2022-04-19
Fix bug on Windows with poll initialization for sockets

benh:


eventuals/event-loop.h
```diff
@@ -797,7 +803,8 @@ class EventLoop final : public Scheduler {
             &context_);
       }
 
-      void Next() override {
+      void
```
onelxj, line 806: Restore?

onelxj (approved):

eventuals/event-loop.h
```diff
@@ -781,9 +781,15 @@ class EventLoop final : public Scheduler {
               if (!completed_) {
                 CHECK(!started_);
                 started_ = true;
-
+#if !defined(_WIN32)
```
onelxj, line 784: Maybe `#ifndef` could be better there?

FolMing:
Come on, guys, why approve? This doesn't work at all!

1. GetFileType requires #include <fileapi.h>
2. GetFileType expects HANDLE as a parameter and not a random integer.

Even after fixing the above things (using reinterpret_cast for the 2.) the test executable crashes:
image

eventuals/event-loop.h
```diff
@@ -781,9 +781,15 @@ class EventLoop final : public Scheduler {
...
                 error_ = uv_poll_init(loop_, poll(), fd_);
-
+#else
+                if (GetFileType(fd_) == FILE_TYPE_PIPE) {
```
FolMing, line 787: 1. GetFileType requires #include <fileapi.h>
2. GetFileType expects HANDLE as a parameter and not a random integer.

FolMing (changes requested):

FolMing:
#346 should fix the issues above.

FolMing (approved):

PR #345, opened 2022-04-19
Pass `Args` from `Task::With` and `Generator::with` as lvalue

onelxj:


benh (approved):
Good stuff! Great tests!

benh:
@onelxj looks like asan has found a bug, please create an issue and copy over the debug output.

I don't think this has to do with your changes, let's try running the tests again to see if it's non-deterministic.

benh:
@onelxj I'm realizing now that this won't work if you try and take a pointer as an argument to `With`, you should be able to confirm with a test.

So we need to think through that before we commit!

ArthurBandaryk (approved):

PR #349, opened 2022-04-20
Move tests into named namespaces

CodingCanuck:
Per https://google.github.io/styleguide/cppguide.html#Namespaces :

> With few exceptions, place code in a namespace.

In addition to making code more style-guide compliant, this reduces
noise by making many `using` statements unnecessary.

Move tests into an unnamed namespace to give them internal linkage:
https://google.github.io/styleguide/cppguide.html#Internal_Linkage

See discussion in https://github.com/3rdparty/eventuals/pull/339

### DIFF
--- a/test/callback.cc
+++ b/test/callback.cc
@@ -3,7 +3,8 @@
 #include "gtest/gtest.h"
 #include "stout/borrowed_ptr.h"
 
-using eventuals::Callback;
+namespace eventuals::test {
+namespace {
 
 TEST(Callback, Destructor) {
   struct Foo {
@@ -64,3 +65,6 @@ TEST(Callback, BorrowedCallable) {
 
   EXPECT_EQ(foo.borrows(), 0);
 }
+
+} // namespace
+} // namespace eventuals::test
--- a/test/catch.cc
+++ b/test/catch.cc
@@ -14,16 +14,9 @@
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
-using eventuals::Catch;
-using eventuals::Conditional;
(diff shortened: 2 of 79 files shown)


onelxj (approved):
Thanks for your patience!

benh:
Let's talk about this synchronously before proceeding please! I'll schedule some time @CodingCanuck.

test/callback.cc
```diff
@@ -3,8 +3,8 @@
 #include "gtest/gtest.h"
 #include "stout/borrowed_ptr.h"
 
-using eventuals::Callback;
-
+namespace eventuals {
+namespace {
 TEST(Callback, Destructor) {
```
benh, line 8: To stay consistent with the style in the code base please keep a newline between `namespace` and anything else!

test/callback.cc
```diff
@@ -3,8 +3,8 @@
 #include "gtest/gtest.h"
 #include "stout/borrowed_ptr.h"
 
-using eventuals::Callback;
-
+namespace eventuals {
```
benh, line 6: As discussed offline, let's put all of these in the `*::test` namespace, e.g., `eventuals::test`.

test/callback.cc
```diff
@@ -64,3 +64,5 @@ TEST(Callback, BorrowedCallable) {
 
   EXPECT_EQ(foo.borrows(), 0);
 }
+} // namespace
```
benh, line 67: Newlines here too please to be consistent with codebase! Thanks!

benh (approved):

PR #350, opened 2022-04-20
Remove unused using-directives

CodingCanuck:
In addition to being unused, these are discouraged by the style guide:
https://google.github.io/styleguide/cppguide.html#Namespaces


onelxj (approved):
Nice!

PR #351, opened 2022-04-20
Move most of HttpMockSever into a .cc file

CodingCanuck:
By defining a separate library for this helper, we're forced to make
`http-mock-server.h` self-contained (which is also required by the style
guide: https://google.github.io/styleguide/cppguide.html#Self_contained_Headers).

### DIFF
--- a/test/http-mock-server.cc
+++ b/test/http-mock-server.cc
@@ -1,0 +1,276 @@
+#include "test/http-mock-server.h"
+
+#include <memory>
+#include <optional>
+#include <string>
+#include <thread>
+
+#include "asio.hpp"
+#include "asio/ssl.hpp"
+#include "eventuals/http.h"
+#include "eventuals/rsa.h"
+#include "eventuals/x509.h"
+#include "gtest/gtest.h"
+
+namespace {
+
+constexpr static size_t kBufferSize = 4096;
+
+// Implementation of an insecure socket, i.e, no TLS/SSL.
+class InsecureSocket final : public HttpMockServer::Socket {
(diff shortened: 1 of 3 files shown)


test/http-mock-server.cc
```diff
@@ -0,0 +1,274 @@
...
+
+#include "asio.hpp"
+#include "asio/ssl.hpp"
+#include "eventuals/http.h"
+#include "eventuals/rsa.h"
+#include "eventuals/x509.h"
+#include "gtest/gtest.h"
+
+namespace {
+constexpr static size_t kBufferSize = 4096;
```
benh, line 16:
```suggestion
namespace {

constexpr static size_t kBufferSize = 4096;
```

test/http-mock-server.cc
```diff
@@ -0,0 +1,274 @@
...
+    stream_.next_layer().close(error);
+    if (error) {
+      ADD_FAILURE() << "Failed to close the socket: " << error.message();
+    }
+  }
+
+ private:
+  asio::ssl::stream<asio::ip::tcp::socket> stream_;
+};
+} // namespace
```
benh, line 94:
```suggestion
};

} // namespace
```

benh (approved):

test/http-mock-server.cc
CodingCanuck: Done.

test/http-mock-server.cc
CodingCanuck: Done.

PR #353, opened 2022-04-21
Introduce mergequeue config

nizamutdinovb:
This PR is about to cover https://github.com/reboot-dev/company/issues/66
As per offline discussion with @rjhuijsman we will merge this PR and see how `mergequeue` config will work.
After that we will proceed with similar configuration on the rest of repositories.


rjhuijsman (approved):
Cool! Let's see how this works!

PR #375, opened 2022-04-28
Add ServerStatus.ToString() and operator<<

CodingCanuck:
This makes it easier to print bad status messages in tests, so we can
change:

`ASSERT_TRUE(status.ok());`

to:

`ASSERT_TRUE(status.ok()) << status;`

This follows the pattern used by `::absl::Status`:
https://github.com/abseil/abseil-cpp/blob/master/absl/status/status.cc#L316-L319

Note that this is dissimilar to `::grpc::Status` which provides no such
convenience functions:
https://grpc.github.io/grpc/cpp/classgrpc_1_1_status.html

### DIFF
--- a/eventuals/grpc/server.h
+++ b/eventuals/grpc/server.h
@@ -2,6 +2,7 @@
 
 #include <cassert>
 #include <deque>
+#include <string_view>
 #include <thread>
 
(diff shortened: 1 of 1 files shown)


rjhuijsman (approved):
👍🏻

PR #376, opened 2022-04-28
Print bad status details on `ASSERT_TRUE(status)`

CodingCanuck:
When a status is unexpectedly OK, it's easier to debug the problem when
the failure messages includes more details about the bad status.


rjhuijsman (approved):
🙏🏻

PR #377, opened 2022-04-28
Google remote cache to work in read-only mode from local workstations

nizamutdinovb:
Refer to issue https://github.com/reboot-dev/company/issues/71


rjhuijsman (approved):
🎉

PR #394, opened 2022-05-10
Split compiling `TypeErased{method}` into different .cc files

onelxj:


protoc-gen-eventuals/protoc-gen-eventuals.py
```diff
@@ -34,6 +34,22 @@ def analyze_file(self, proto_file: FileDescriptorProto) -> dict:
...
+            for method in service['methods']:
+              for client_streaming in [True, False]:
+                for server_streaming in [True, False]:
+                  if method['server_streaming'] == server_streaming and method['client_streaming'] == client_streaming:
+                    header_type[(client_streaming, server_streaming)] = True
+
+        header_type['has_bidirectional_streaming'] = header_type.pop((True, True), False)
+        header_type['has_server_streaming'] = header_type.pop((False, True), False)
+        header_type['has_client_streaming'] = header_type.pop((True, False), False)
+       	header_type['has_no_streaming'] = header_type.pop((False, False), False)
```
while-false, line 49: I know I suggested these names, but I only now realize that they somewhat clash with the `method['server_streaming']` and `method['client_streaming']`...

Can we rename these variables (here and in the template) to the slightly longer but also more descriptive `bidirectional_streaming_methods_exists` (for the bidirectional case)?
I think this makes it clearer that it is not about the particular method but about the general existence of a particular type of methods.

To be clear, the naming of the templates I think is fine, it is just the variable name that could be a little crisper.

while-false (changes requested):

protoc-gen-eventuals/protoc-gen-eventuals.py
```diff
@@ -41,31 +41,51 @@ def process_file(self, proto_file: FileDescriptorProto):
...
 
         header_file_name = proto_file_name.replace('.proto', '.eventuals.h')
         source_file_name = proto_file_name.replace('.proto', '.eventuals.cc')
 
         eventuals_data = {
             'eventuals_header': header_file_name,
-            'grpc_pb_header': proto_file_name.replace('.proto', '.grpc.pb.h'),
+            'grpc_pb_header': '../' + proto_file_name.replace('.proto', '.grpc.pb.h'),
+            'service_number': 0,
+            'method_number': 0,
```
while-false, line 61: Declare these `None` or omit them here. They are not incremented, but assigned later. 
In jinja, using `None` would also allow for patterns like `service_number | default 0`.

protoc-gen-eventuals/protoc-gen-eventuals.py
```diff
@@ -41,31 +41,51 @@ def process_file(self, proto_file: FileDescriptorProto):
...
-
+        # Generate only base files, without rpc methods implementation
+        for file_name, template_name in general_outputs:
             template = load_template(template_name)
             content = template.render(**template_data)
+            self.response.file.add(name=file_name, content=content)
+
+        generate_methods_template_name = 'grpc-method-eventuals.cc.j2'
+
+        for service_index, service in enumerate(template_data['services'], start = 0):
```
while-false, line 79: Drop the `start = 0` here and in next line.

It's the standard "we always start counting at 0" and `start=0` is indeed the default value.

Further, keyword arguments have no spaces according to the style guide. I.e. it would have been `start=N` and not `start = N`. :)

protoc-gen-eventuals/protoc-gen-eventuals.py
while-false: Actually, reading further, I would omit these variables here. The rest of the variables in this `dict` is generic, these are changing non-generic ones. There is little won in terms of performance in merging the two, there is some lost clarity and maintainability in merging, so let's keep it seperate.

protoc-gen-eventuals/protoc-gen-eventuals.py
```diff
@@ -41,31 +41,51 @@ def process_file(self, proto_file: FileDescriptorProto):
...
+            for method_index, method in enumerate(service['methods'], start = 0):
+                template = load_template(generate_methods_template_name)
+                template_data['service_number'] = service_index
+                template_data['method_number'] = method_index
```
while-false, line 83: Given the comment above, I'd change these lines to:
```python
service_data = {
    'service_number': service_number,
    'method_number': method_number,
}

File: protoc-gen-eventuals/protoc-gen-eventuals.py [hunk 5]
```diff
@@ -41,31 +41,51 @@ def process_file(self, proto_file: FileDescriptorProto):
...
+
+                content = template.render(**template_data)
```
while-false, line 85: ... and then change this line to:
```python
content = template.render(**template_data, **service_data)
```

Making it (hopefully very) clear that we are rendering the template based on a combination of generic data (from `template_data` and some service specific data from `service_data`.

bazel/cc_eventuals_library.bzl
```diff
@@ -6,5 +6,4 @@ load("@com_github_reboot_dev_pyprotoc_plugin//:rules.bzl", "create_protoc_plugin
 
 cc_eventuals_library = create_protoc_plugin_rule(
     "@com_github_3rdparty_eventuals//protoc-gen-eventuals:protoc-gen-eventuals",
-    extensions = (".eventuals.h", ".eventuals.cc"),
```
while-false, line 9: I assume this relies on the `declare_directory` approach that we discussed? Did you get it to work and does it pick up changes automatically?

protoc-gen-eventuals/templates/grpc-method-eventuals.cc.j2
```diff
@@ -0,0 +1,128 @@
+{%- set service = services[service_number] %}
+{%- set method = service.methods[method_number] %}
```
while-false, line 2: Are you using `service_number` or `method_number` anywhere else?

If not, you can change skip all the `enumerate(...)` you do in python and just do:

```python
for service in template_data['services']:
    for method in service['methods']:
    	# ...
	service_data = {
	    'service': service,
	    'method': method,
	}
	# ...
	content = template.render(**template_data, **service_data)
```

and this would give you the variable `service` and `method` in your jinja.

while-false (changes requested):
Hey @onelxj, it in general looks good - thanks for writing this up!

Most of the comments are mostly targeting your python and not so much the functionality.

As a reader and reviewer, I had to jump around between files to figure out what you might have thought. I thing a think that would have helped me greatly would be some strategic comments in the code explaining some of the "Why's" (as opposed to the "What's") of the code. Does that make sense?

For instance, why the `enumerate` and the `method_index`? For what purpose are we extracting or passing these on?

protoc-gen-eventuals/protoc-gen-eventuals.py
```diff
@@ -41,31 +41,51 @@ def process_file(self, proto_file: FileDescriptorProto):
...
+
+        proto_file_name = proto_file_path[index:]
 
         header_file_name = proto_file_name.replace('.proto', '.eventuals.h')
         source_file_name = proto_file_name.replace('.proto', '.eventuals.cc')
 
         eventuals_data = {
             'eventuals_header': header_file_name,
-            'grpc_pb_header': proto_file_name.replace('.proto', '.grpc.pb.h'),
+            'grpc_pb_header': '../' + proto_file_name.replace('.proto', '.grpc.pb.h'),
```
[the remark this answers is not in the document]
benh: We should not use `../` in includes please!

protoc-gen-eventuals/templates/grpc-method-eventuals.cc.j2
benh: I like this a lot!

benh (changes requested):
Thanks @while-false for the great review!

while-false (approved):
Thanks @onelxj ! :)

protoc-gen-eventuals/protoc-gen-eventuals.py
```diff
@@ -41,31 +41,51 @@ def process_file(self, proto_file: FileDescriptorProto):
...
-        proto_file_name = file_digest.pop('proto_file_name')
+        proto_file_path = file_digest.pop('proto_file_name')
+
+        index = len(proto_file_path)
+        for ch in reversed(proto_file_path):
+            if (ch == '/'):
+                break
+            index -= 1
+
+        proto_file_name = proto_file_path[index:]
```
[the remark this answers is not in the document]
onelxj: it is outdated =)

benh (approved):

PR #403, opened 2022-05-19
Introduce google failure signal handler

ArthurBandaryk:
Earlier we had absolutely horrible stack trace on `glog` check failure, like in example below:
stack

This PR introduces a convenient [signal handler](https://github.com/google/glog#failure-signal-handler) that will dump useful information when the program crashes on certain signals such as SIGSEGV. The signal handler can be installed by `google::InstallFailureSignalHandler()` from `glog` library.

### DIFF
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -61,6 +61,8 @@ cc_test(
         "eventual.cc",
         "executor.cc",
         "expected.cc",
+        "failure-signal-handler.cc",
+        "failure-signal-handler.h",
         "filesystem.cc",
         "filter.cc",
         "finally.cc",
--- a/test/failure-signal-handler.cc
+++ b/test/failure-signal-handler.cc
@@ -1,0 +1,8 @@
+#include "failure-signal-handler.h"
+
+// For setting up gtest environment in order to use
+// `glog` failure signal handler:
+// https://github.com/google/glog#failure-signal-handler
+static testing::Environment* const signal_handler_failure_env =
+    testing::AddGlobalTestEnvironment(
+        new InstallFailureSignalHandlerEnvironment{});
--- a/test/failure-signal-handler.h
+++ b/test/failure-signal-handler.h
@@ -1,0 +1,13 @@
+#include "glog/logging.h"
+#include "gtest/gtest.h"
+
+class InstallFailureSignalHandlerEnvironment : public ::testing::Environment {
+ public:
+  ~InstallFailureSignalHandlerEnvironment() override {}
+
+  void SetUp() override {
+    google::InstallFailureSignalHandler();
+  }
+
+  void TearDown() override {}
+};


test/BUILD.bazel
```diff
@@ -46,6 +46,17 @@ cc_library(
     ],
 )
 
+cc_library(
+    name = "environment",
```
onelxj, line 50: What do you think about `test_environment`?

test/signal-handler-env.h
```diff
@@ -0,0 +1,15 @@
+#include "glog/logging.h"
+#include "gtest/gtest.h"
+
+class SignalHandlerEnvironment : public ::testing::Environment {
+ public:
+  ~SignalHandlerEnvironment() override {}
+
+  // Override this to define how to set up the environment.
```
onelxj, line 8: remove please

test/signal-handler-env.h
```diff
@@ -0,0 +1,15 @@
...
+  void SetUp() override {
+    google::InstallFailureSignalHandler();
+  }
+
+  // Override this to define how to tear down the environment.
```
onelxj, line 13: remove please

test/stack-trace-check.cc
```diff
@@ -0,0 +1,30 @@
+#include "glog/logging.h"
+#include "gtest/gtest.h"
+
+// For setting up gtest environment in order to use
+// `glog` failure signal handler:
+// https://github.com/google/glog#failure-signal-handler
+#include "signal-handler-env.h"
+
+// Including eventuals stuff.
```
onelxj, line 9: remove please

test/stack-trace-check.cc
```diff
@@ -0,0 +1,28 @@
...
+// `glog` failure signal handler:
+// https://github.com/google/glog#failure-signal-handler
+#include "signal-handler-env.h"
+
+using namespace eventuals;
+
+static testing::Environment* const signal_handler_failure_env =
+    testing::AddGlobalTestEnvironment(new SignalHandlerEnvironment{});
+
+TEST(StackTrace, CheckFail) {
```
benh, line 16: This file is great for testing but let's not include it when we merge, thanks!

test/stack-trace-check.cc
```diff
@@ -0,0 +1,28 @@
...
+
+// For setting up gtest environment in order to use
+// `glog` failure signal handler:
+// https://github.com/google/glog#failure-signal-handler
+#include "signal-handler-env.h"
+
+using namespace eventuals;
+
+static testing::Environment* const signal_handler_failure_env =
+    testing::AddGlobalTestEnvironment(new SignalHandlerEnvironment{});
```
benh, line 14: This will need to be in a different `.cc` in the tests.

benh (changes requested):

test/BUILD.bazel
```diff
@@ -46,6 +46,17 @@ cc_library(
     ],
 )
 
+cc_library(
```
benh, line 49: Why does this need to be its own library?

benh (changes requested):

PR #404, opened 2022-05-20
Prefer constructor to extra 'Lazy' function

ArthurBandaryk:


eventuals/lazy.h
```diff
@@ -12,17 +12,20 @@ namespace eventuals {
 ////////////////////////////////////////////////////////////////////////
 
 template <typename T, typename... Args>
-class _Lazy final {
+class Lazy final {
```
benh, line 15:
```suggestion
template <typename T_, typename... Args_>
class Lazy final {
 public:
   template <typename... Args>
   using Args = std::enable_if_t<sizeof...(Args_) == 0, Lazy<T_, Args...>>;
   
   Lazy(Args_&&... args)
    : args_(std::forward<Args_>(args)...) {}
    
  Lazy(Lazy&& that)
    : args_(std::move(that.args_)) {
    CHECK(!t_) << "'Lazy' can not be moved after using";
  }
```

eventuals/lazy.h
```diff
@@ -12,17 +12,20 @@ namespace eventuals {
...
  public:
   template <typename Tuple>
-  _Lazy(Tuple args)
+  Lazy(Tuple args)
```
benh, line 18: Can you see if we can delete this one and instead always pass the 'args' and construct the tuple from that?

benh (changes requested):

onelxj (approved):
LGTM!

benh (changes requested):

onelxj (approved):

eventuals/lazy.h
```diff
@@ -11,19 +11,26 @@ namespace eventuals {
...
-template <typename T, typename... Args>
-class _Lazy final {
+template <typename T_, typename... Args_>
+struct _Lazy final {
  public:
-  template <typename Tuple>
-  _Lazy(Tuple args)
-    : args_(std::move(args)) {}
+  template <typename... Args>
+  using Args = std::enable_if_t<sizeof...(Args_) == 0, _Lazy<T_, Args...>>;
```
benh, line 18: To ensure someone can't do `Lazy::Of<Foo>::Args<>::Args<int>`:

```suggestion
  template <typename Arg, typename... Args>
  using Args = std::enable_if_t<sizeof...(Args_) == 0, _Lazy<T_, Arg, Args...>>;
```

benh (approved):

PR #422, opened 2022-05-31
Refactor interrupts

onelxj:


eventuals/interrupt.h
```diff
@@ -39,67 +38,45 @@ class Interrupt final {
...
 
     Interrupt* interrupt_ = nullptr;
-    Callback<void()> callback_;
-    Handler* next_ = nullptr;
+    std::optional<Callback<void()>> callback_;
   };
 
-  Interrupt()
-    : handler_(this, []() {}) {}
+  Interrupt() = default;
```
benh, line 54:
```suggestion
  Interrupt() = default;
  
    ~Interrupt() {
      // TO BE FINISHED!
      auto* handler = handler_.exchange(&placeholder_handler_, ........);      
      if (handler != nullptr) {
        handler->state_.Update(INSTALLED, UNINSTALLED));        
      }
    }
```

eventuals/interrupt.h
```diff
@@ -39,67 +38,45 @@ class Interrupt final {
...
   }
 
-  // To simplify the implementation we signify a triggered interrupt
-  // by storing nullptr in head_. Thus, when an interrupt is first
-  // constructed we store a "placeholder" handler that we ignore when
-  // executing the rest of the handlers.
-  Handler handler_;
-
-  std::atomic<Handler*> head_ = &handler_;
+  std::atomic<Handler*> handler_ptr_ = nullptr;
```
benh, line 79:
```suggestion
  Handler* status_ = &placeholder_handler_;
  std::atomic<Handler*> handler_ = nullptr;
```

benh (changes requested):

eventuals/interrupt.h
```diff
@@ -35,71 +47,97 @@ class Interrupt final {
...
+      Handler::State state = handler->state_.state();
+      handler->state_.Update(state, Handler::State::UNINSTALLED);
+    }
+  }
 
-    handler->next_ = head_.load(std::memory_order_relaxed);
+  bool Install(Handler* handler) {
+    if (handler_.load(std::memory_order_relaxed) != nullptr) {
+      Handler::State state = handler->state_.state();
+      handler->state_.Update(state, Handler::State::INSTALLED);
```
benh, line 99:
```suggestion
      CHECK_EQ(state, Handler::State::UNINSTALLED);
      CHECK(handler->state_.Update(state, Handler::State::INSTALLED));
```

eventuals/interrupt.h
```diff
@@ -35,71 +47,97 @@ class Interrupt final {
...
+        handler_.exchange(&placeholder_handler_, std::memory_order_relaxed);
+    if (handler != nullptr) {
+      Handler::State state = handler->state_.state();
+      handler->state_.Update(state, Handler::State::UNINSTALLED);
+    }
+  }
 
-    handler->next_ = head_.load(std::memory_order_relaxed);
+  bool Install(Handler* handler) {
+    if (handler_.load(std::memory_order_relaxed) != nullptr) {
```
benh, line 97: Just doing a `load()` is insufficient here because you are racing with a possible call to `Trigger()`.

You need to do a `load()`, then a `compare_exchange()`. The semantics here are that EITHER the `compare_exchange()` is successful or IT MUST be the case that `Trigger()` was called and therefore `handler_` is now set to `nullptr` so we should `CHECK()` that. That's because with our new semantics if you do a `load()` and then someone else tries to do an `Install()` then we must have a bug in our code because only one scheduler context should be installing on the same interrupt at a time.

eventuals/interrupt.h
```diff
@@ -35,71 +47,97 @@ class Interrupt final {
...
+    auto* handler =
+        handler_.exchange(&placeholder_handler_, std::memory_order_relaxed);
+    if (handler != nullptr) {
+      Handler::State state = handler->state_.state();
+      handler->state_.Update(state, Handler::State::UNINSTALLED);
+    }
+  }
 
-    handler->next_ = head_.load(std::memory_order_relaxed);
+  bool Install(Handler* handler) {
```
benh, line 96:
```suggestion
  [[nodiscard]] bool Install(Handler* handler) {
```

eventuals/interrupt.h
```diff
@@ -35,71 +47,97 @@ class Interrupt final {
...
-    do {
-      // Check if the interrupt has already been triggered.
-      if (handler->next_ == nullptr) {
-        return false;
-      }
-    } while (!head_.compare_exchange_weak(
-        handler->next_,
+  void Uninstall(Handler* handler) {
+    Handler::State state = handler->state_.state();
+    handler->state_.Update(state, Handler::State::UNINSTALLED);
```
benh, line 109:
```suggestion
    Handler::State state = handler->state_.state();
    CHECK_EQ(state, Handler::State::UNINSTALLING);
    CHECK(handler->state_.Update(state, Handler::State::UNINSTALLED));
```

eventuals/interrupt.h
```diff
@@ -35,71 +47,97 @@ class Interrupt final {
...
-        handler->next_ = nullptr;
-        handler = next;
-      }
+      handler->Invoke();
+    }
+
+    Interrupt* linked = linked_interrupt_.load(std::memory_order_relaxed);
+    if (linked) {
+      linked->Trigger();
     }
```
benh, line 129: Let's talk about this synchronously.

test/concurrent/emit-interrupt-fail.cc
```diff
@@ -23,29 +23,27 @@ TYPED_TEST(ConcurrentTypedTest, EmitInterruptFail) {
...
 
-  static_assert(
-      eventuals::tuple_types_unordered_equals_v<
-          typename decltype(e())::template ErrorsFrom<void, std::tuple<>>,
-          std::tuple<std::runtime_error>>);
+  static_assert(eventuals::tuple_types_unordered_equals_v<
+                typename decltype(e())::template ErrorsFrom<
+                    void,
+                    std::tuple<>>,
+                std::tuple<std::runtime_error>>);
```
benh, line 46: Please revert this formatting change, it's inconsistent with the rest!

test/concurrent/emit-interrupt-fail.cc
```diff
@@ -23,29 +23,27 @@ TYPED_TEST(ConcurrentTypedTest, EmitInterruptFail) {
...
                  }
                })
-        | this->ConcurrentOrConcurrentOrdered([]() {
-            return Map([](int i) {
-              return std::to_string(i);
-            });
-          })
+
+        | this->ConcurrentOrConcurrentOrdered(
+            []() { return Map([](int i) { return std::to_string(i); }); })
```
benh, line 38: Please revert this formatting change. And please review your PRs in the future to undo mistakes like this so I don't have to! Thank you!

test/concurrent/emit-interrupt-fail.cc
```diff
@@ -23,29 +23,27 @@ TYPED_TEST(ConcurrentTypedTest, EmitInterruptFail) {
                .raises<std::runtime_error>()
                .interruptible()
                .begin([](auto& k, Interrupt::Handler& handler) {
-                 handler.Install([&k]() {
-                   k.Fail(std::runtime_error("error"));
-                 });
+                 handler.Install(
+                     [&k]() { k.Fail(std::runtime_error("error")); });
```
benh, line 27: Why this formatting change? It's inconsistent with the rest.

test/concurrent/emit-interrupt-stop.cc
```diff
@@ -18,29 +18,25 @@ TYPED_TEST(ConcurrentTypedTest, EmitInterruptStop) {
     return Stream<int>()
                .interruptible()
                .begin([](auto& k, Interrupt::Handler& handler) {
-                 handler.Install([&k]() {
-                   k.Stop();
-                 });
+                 handler.Install([&k]() { k.Stop(); });
```
benh, line 21: Please don't introduce formatting changes unless they are to make the codebase more consistent.

benh (changes requested):
Please split this review out into multiple different PRs:

For the first PR, let's just add the interrupt handler to `next()`, `fail()`, etc for `Stream` and `Loop`. When you update all of the tests just take an `auto&` for the handler since we'll be updating the type in the next PR.

For the second PR, let's never dereference `*handler_` when calling any of the functions mentioned above and instead always pass the handler in its `std::optional` form.

For the third PR, let's update `Interrupt` to only ever have a single interrupt not the linked list. Does this break tests? If so, then lets combine the next couple of PRs but as separate commits.

For the fourth PR (or second commit in the last PR), let's update `Interrupt` to have an uninstall which gets called by the `~Handler()`. 

For the fifth PR (or third commit in the last PR), let's update all existing code that installs interrupt handlers to install them in the correct place. We need to talk synchronously about this.

For the last PR, let's add support for interrupts to `Wait()`.

PR #424, opened 2022-06-01
Add 'expected' dependency for eventuals

ArthurBandaryk:
Since we have [`bazelified`](https://github.com/3rdparty/bazel-rules-expected.git) `expected` [repo](https://github.com/TartanLlama/expected.git), now we can make eventuals also depend on this library too.
# DO NOT MERGE IT UNTIL THE PR IN `3rdparty/bazel-rules-expected` WILL BE MERGED!!!


eventuals/BUILD.bazel
```diff
@@ -131,5 +131,6 @@ cc_library(
         ":events",
         ":http",
         "@com_github_3rdparty_stout//:stout",
+        "@com_github_expected//:expected",
```
benh, line 134: We don't need this just yet! And also, seeing that name name, let's update it to `com_github_tl_expected` in the other PR. Thanks!

benh (approved):

PR #426, opened 2022-06-05
Replace 'Expected' with 'tl::expected'

benh:
Use the C++23 proposed 'std::expected' instead of our own.

### DIFF
--- a/eventuals/compose.h
+++ b/eventuals/compose.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <type_traits>
+
 #include "eventuals/type-traits.h"
 #include "glog/logging.h"
 
--- a/eventuals/expected.h
+++ b/eventuals/expected.h
@@ -1,10 +1,8 @@
 #pragma once
 
-#include <variant>
-
+#include "eventuals/compose.h"
 #include "eventuals/eventual.h"
-#include "eventuals/terminal.h"
-#include "eventuals/type-traits.h"
+#include "tl/expected.hpp"
 
 ////////////////////////////////////////////////////////////////////////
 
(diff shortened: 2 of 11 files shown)


eventuals/finally.h
```diff
@@ -16,9 +16,12 @@ struct _Finally final {
     template <typename... Args>
     void Start(Args&&... args) {
       if constexpr (std::is_void_v<Arg_>) {
-        k_.Start(std::optional<std::exception_ptr>());
+        k_.Start(
+            std::optional<std::exception_ptr>());
```
onelxj, line 20: Why?

eventuals/expected.h
```diff
@@ -1,224 +1,156 @@
...
+// 'eventuals::Status' is an eventuals specific error type. We might
+// then also have 'eventuals::grpc::expected<T>' be an alias for
+// 'std::expected<T, eventuals::grpc::Status>', and so forth and so
+// on.
+//
+// For now, an 'eventuals::expected<T>' uses a 'std::string' as the
+// default error type to simplify calls to 'make_unexpected("...")'
+// that take strings which is the majority (if not all) of calls.
+template <typename Value_, typename Error_ = std::string>
+class expected : public tl::expected<Value_, Error_> {
```
onelxj, line 78: What do you think about adding one more syntax sugar for expected for `std::exception_ptr`, I think it might be useful to the endusers since they are not worried about the error

eventuals/compose.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <type_traits>
```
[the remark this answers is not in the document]
benh: We should "include what you use", which is why I've added this.

eventuals/expected.h
```diff
@@ -1,224 +1,156 @@
...
-    : error(std::move(error)) {}
-
-  Error_ error;
-};
+// Helper for creating an eventual from an 'expected'.
+template <typename T, typename E>
+auto ExpectedToEventual(tl::expected<T, E>&& expected) {
+  // TODO(benh): support any error type that can be "stringified".
+  static_assert(
+      std::disjunction_v<
```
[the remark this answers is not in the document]
benh: Let's keep the explicitness for now since it's not too many places.

eventuals/expected.h
benh: Yes I totally agree! But let's think more about it and figure out what we want it to look like and then do that in a subsequent review.

onelxj (approved):
Thanks!

PR #340, opened 2022-04-19
Add eventuals-grpc `RouteGuide` tests

onelxj:

PR #428, opened 2022-06-06
Split generated

onelxj:


.bazelrc
```diff
@@ -5,7 +5,8 @@ build --enable_platform_specific_config
 
 build:macos --cxxopt='-std=c++17'
 
-build:linux --cxxopt='-std=c++17'
+# Specify repo_env std version for protoc generated files.
+build:linux --cxxopt='-std=c++17' --repo_env=BAZEL_CXXOPTS="-std=c++17"
```
benh, line 9: So does this mean that all repositories will need to do this? Because that will still break people. We didn't need to do anything before and we _definitely_ had C++17 code, why do we need to do something now?

benh:
What's the status of this PR?

onelxj:
> What's the status of this PR?

This one is included and a bit changed [here](https://github.com/3rdparty/eventuals/pull/340)

PR #430, opened 2022-06-11
Add configs from Tensorflow's .bazelrc file

FolMing:
[Tensorflow's .bazelrc file](https://github.com/tensorflow/tensorflow/blob/e9f76ac8a0e8767f0a16716e84599354d6044e69/.bazelrc) provides various useful configs to make life easier for the developers.

1) It sets [--verbose_failures](https://bazel.build/reference/command-line-reference#flag--verbose_failures) as a default setting when building. In case of a build error with this option enabled bazel will try to provide as much information as possible.
2) It defines [WIN32_LEAN_AND_MEAN](https://docs.microsoft.com/en-us/windows/win32/winprog/using-the-windows-headers#faster-builds-with-smaller-header-files) which also helped me in the past when I was dealing with various BoringSSL issues. By default including "Windows.h" will include pretty much everything that Windows API has, including various declarations of the same structures that BoringSSL and OpenSSL have which produces compilation errors. Defining [WIN32_LEAN_AND_MEAN](https://docs.microsoft.com/en-us/windows/win32/winprog/using-the-windows-headers#faster-builds-with-smaller-header-files) disables the whole Windows.h's "include everything" strategy. NOGDI pretty much does the same thing.
3) I'm not particularly sure about "--copt=/Zc:__cplusplus" thing, but while dealing with ASIO we encountered several errors regarding memory allocation and deallocation (even though we saw it happen only on Linux, I would like to remind you that we only have AddressSanitizer enabled in our Linux workflow, so we don't really know what is happening on Windows and MacOS), so I still included it into the config. At least it doesn't break our build or test and it looks like [MSVC team also recommends setting this option](https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/).

I also refactored our C++17 configs, now they look similar to Tensorflow's .bazelrc file.

I added an additional config for building eventuals with GCC, just in case.


ArthurBandaryk (approved):
LGTM! Thanks!

onelxj:
--host_cxxopt would be used only for genrules, do we need that one?

.bazelrc
```diff
@@ -3,16 +3,34 @@ try-import submodules/dev-tools/.bazelrc
...
+build --verbose_failures
 
+# Use Clang as a default compiler on Windows.
 build:windows --compiler="clang-cl"
 
+# Compiler options.
+build:gcc --action_env=CC=gcc
 build:clang --action_env=CC=clang
 
+# By default, build eventuals in C++17 mode.
```
CodingCanuck, line 16: I'd remove the "by default" bit (it doesn't add much: this whole file is defining defaults)

.bazelrc
```diff
@@ -3,16 +3,34 @@ try-import submodules/dev-tools/.bazelrc
...
+build:macos --cxxopt=-std=c++17
+build:macos --host_cxxopt=-std=c++17
+build:windows --cxxopt=/std:c++17
+build:windows --host_cxxopt=/std:c++17
+
+# On Windows, `__cplusplus` is wrongly defined without this switch.
+# See https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
+build:windows --copt=/Zc:__cplusplus
+
+# Make sure to include as little of windows.h as possible.
```
CodingCanuck, line 28: Please elaborate: why? (compilation speed, correctness, other?)

(Whatever the reason, avoiding most of `windows.h` sounds great! "Why we did this" comments tend to be more useful than "What we did here" comments for maintainers looking at this code several years from now, though)

CodingCanuck, line 28: Are there any references we can link to which recommend these flags?

I don't think I've seen `NOGDI` before: given that you mention that it's similar to [`WIN32_LEAN_AND_MEAN`](https://docs.microsoft.com/en-us/windows/win32/winprog/using-the-windows-headers#faster-builds-with-smaller-header-files) I'm wondering if we need both.

FolMing: I can't really find any information about those macros on msdn, but after some digging for half a minute through my own windows.h file I found a bunch of them:
```
/*  If defined, the following flags inhibit definition
 *     of the indicated items.
 *
 *  NOGDICAPMASKS     - CC_*, LC_*, PC_*, CP_*, TC_*, RC_
 *  NOVIRTUALKEYCODES - VK_*
 *  NOWINMESSAGES     - WM_*, EM_*, LB_*, CB_*
 *  NOWINSTYLES       - WS_*, CS_*, ES_*, LBS_*, SBS_*, CBS_*
 *  NOSYSMETRICS      - SM_*
 *  NOMENUS           - MF_*
 *  NOICONS           - IDI_*
 *  NOKEYSTATES       - MK_*
 *  NOSYSCOMMANDS     - SC_*
 *  NORASTEROPS       - Binary and Tertiary raster ops
 *  NOSHOWWINDOW      - SW_*
 *  OEMRESOURCE       - OEM Resource values
 *  NOATOM            - Atom Manager routines
 *  NOCLIPBOARD       - Clipboard routines
 *  NOCOLOR           - Screen colors
 *  NOCTLMGR          - Control and Dialog routines
 *  NODRAWTEXT        - DrawText() and DT_*
 *  NOGDI             - All GDI defines and routines
 *  NOKERNEL          - All KERNEL defines and routines
 *  NOUSER            - All USER defines and routines
 *  NONLS             - All NLS defines and routines
 *  NOMB              - MB_* and MessageBox()
 *  NOMEMMGR          - GMEM_*, LMEM_*, GHND, LHND, associated routines
 *  NOMETAFILE        - typedef METAFILEPICT
 *  NOMINMAX          - Macros min(a,b) and max(a,b)
 *  NOMSG             - typedef MSG and associated routines
 *  NOOPENFILE        - OpenFile(), OemToAnsi, AnsiToOem, and OF_*
 *  NOSCROLL          - SB_* and scrolling routines
 *  NOSERVICE         - All Service Controller routines, SERVICE_ equates, etc.
 *  NOSOUND           - Sound driver routines
 *  NOTEXTMETRIC      - typedef TEXTMETRIC and associated routines
 *  NOWH              - SetWindowsHook and WH_*
 *  NOWINOFFSETS      - GWL_*, GCL_*, associated routines
 *  NOCOMM            - COMM driver routines
 *  NOKANJI           - Kanji support stuff.
 *  NOHELP            - Help engine interface.
 *  NOPROFILER        - Profiler interface.
 *  NODEFERWINDOWPOS  - DeferWindowPos routines
 *  NOMCX             - Modem Configuration Extensions
 */
```

.bazelrc
```diff
@@ -3,16 +3,34 @@ try-import submodules/dev-tools/.bazelrc
...
 
-build:macos --cxxopt='-std=c++17'
-
-build:linux --cxxopt='-std=c++17'
-
-build:windows --cxxopt="/std:c++17"
+# Verbose failure logs when something goes wrong.
+build --verbose_failures
 
+# Use Clang as a default compiler on Windows.
```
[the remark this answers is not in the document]
FolMing: Added!

.bazelrc
FolMing: Added a comment!

PR #435, opened 2022-06-21
Introduce second constructor for grpc::Client

rjhuijsman:


eventuals/grpc/client.h
```diff
@@ -305,6 +305,12 @@ class Client {
     : channel_(::grpc::CreateChannel(target, credentials)),
       pool_(std::move(pool)) {}
 
+  Client(
+      std::shared_ptr<::grpc::Channel> channel,
```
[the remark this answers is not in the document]
FolMing: Oh wow, I found the answer myself, didn't know that it's actually better to just make a copy and move it inside!
Some compilers even trigger a warning stating that we should pass `shared_ptr` by value: https://docs.microsoft.com/en-us/cpp/code-quality/c26416?view=msvc-170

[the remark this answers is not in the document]
rjhuijsman: Interesting! I found some conflicting info: https://stackoverflow.com/questions/41871115/why-would-i-stdmove-an-stdshared-ptr

But let's not run afoul of compiler warnings - if we find that this `shared_ptr` pass-by-value is a performance pain we'll deal with that later. :)

eventuals/grpc/client.h
```diff
@@ -305,6 +305,12 @@ class Client {
     : channel_(::grpc::CreateChannel(target, credentials)),
       pool_(std::move(pool)) {}
 
+  Client(
```
CodingCanuck, line 308: Generally, all constructors should be `explicit` (probably the constructor above, too): https://abseil.io/tips/142

test/grpc/unary.cc
```diff
@@ -19,7 +19,8 @@ using helloworld::HelloRequest;
 
 using stout::Borrowable;
 
-TEST(UnaryTest, Success) {
+void TestUnaryWithClient(
+    std::function<Client(Borrowable<CompletionPool>&&, int)> client_factory) {
```
CodingCanuck, line 23: When declaring a function which takes an `std::function` which it only calls *but does not store*, prefer to take it by `const &`.

std::function can be ~heavy to copy. (Think of it like a ~big functor struct)

test/grpc/unary.cc
```diff
@@ -93,5 +90,26 @@ TEST(UnaryTest, Success) {
   server->Wait();
 }
 
+TEST(UnaryTest, SuccessWithDefaultChannel) {
+  TestUnaryWithClient([](Borrowable<CompletionPool>&& pool, int port) {
```
CodingCanuck, line 94: optional: `port` can be const, here and below

test/grpc/unary.cc
FolMing: > std::function can be ~heavy to copy.
What about passing it as rvalue?

CodingCanuck: Why? (What's the advantage over passing by const ref, where there's no requirement to construct or copy an object at all?)

eventuals/grpc/client.h
FolMing: That's a great advice! We should also add `explicit` in other parts of eventuals as well.

rjhuijsman: 👍🏻

test/grpc/unary.cc
rjhuijsman: 👍🏻

test/grpc/unary.cc
rjhuijsman: 👍🏻

PR #440, opened 2022-06-22
Use 'absl::StripAsciiWhitespace' to trim strings in eventuals/http.h

FolMing:
Since we use WIN32_LEAN_AND_MEAN we also do not need to order our includes in a specific way.


CodingCanuck (approved):
So much nicer!! Thanks :)

PR #441, opened 2022-06-23
Make `TypeCheck` work when mixing constness

CodingCanuck:
Specifically, make this code pattern compile:
```c++
auto Method() {
    return ::eventuals::TypeCheck<std::unique_ptr<const MyType>>(
        ::eventuals::Then([]() {
            std::unique_ptr<MyType> ptr = std::make_unique<MyType>();
            // Return a non-const ptr that can be assigned to const ptr.
            return ptr;
        })
    );
}
```

Add a bunch of const-conversion tests.

### DIFF
--- a/eventuals/type-check.h
+++ b/eventuals/type-check.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <utility>
+
 #include "eventuals/eventual.h"
 
 ////////////////////////////////////////////////////////////////////////
@@ -19,8 +21,10 @@ struct _TypeCheck {
   template <typename Arg, typename K>
   auto k(K k) && {
     static_assert(
-        std::is_same_v<T_, ValueFrom<Arg>>,
-        "Failed to type check; expecting type on left, found type on right");
+        std::disjunction_v<
+            std::is_assignable<T_, ValueFrom<Arg>>,
+            std::is_convertible<ValueFrom<Arg>, T_>>,
+        "Failed to type check: Cannot return type on right into type on left");
 
     return std::move(e_).template k<Arg>(std::move(k));
   }
--- a/test/type-check.cc
+++ b/test/type-check.cc
@@ -1,34 +1,91 @@
 #include "eventuals/type-check.h"
 
+#include <memory>
 #include <vector>
 
 #include "eventuals/collect.h"
 #include "eventuals/iterate.h"
+#include "eventuals/just.h"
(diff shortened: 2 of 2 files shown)


test/type-check.cc
```diff
@@ -1,34 +1,89 @@
...
+TEST(TypeCheck, UniquePtr) {
+  auto s = [&]() {
+    return TypeCheck<std::unique_ptr<int>>(Just(std::make_unique<int>(4)));
+  };
+  EXPECT_EQ(4, **s());
+}
+
+TEST(TypeCheck, ConstUniquePtr) {
+  auto s = [&]() {
+    return TypeCheck<std::unique_ptr<const int>>(Just(std::make_unique<const int>(4)));
```
onelxj, line 72: Refactor this to be not more then 80 characters please =)

test/type-check.cc
```diff
@@ -1,34 +1,89 @@
...
   };
+  EXPECT_EQ(4, **s());
+}
 
-  EXPECT_EQ(std::vector<int>({5, 12}), *s());
+TEST(TypeCheck, ConstUniquePtrFromNonConstUniquePtr) {
+  auto s = [&]() {
+    return TypeCheck<std::unique_ptr<const int>>(Just(std::make_unique<int>(4)));
```
onelxj, line 79: Refactor this to be not more then 80 characters please =)

eventuals/type-check.h
```diff
@@ -19,8 +21,8 @@ struct _TypeCheck {
   template <typename Arg, typename K>
   auto k(K k) && {
     static_assert(
-        std::is_same_v<T_, ValueFrom<Arg>>,
-        "Failed to type check; expecting type on left, found type on right");
+        std::is_assignable_v<T_, ValueFrom<Arg>> || std::is_convertible_v<ValueFrom<Arg>, T_>,
```
onelxj, line 24: Refactor this to be not more then 80 characters please =)

try this with `std::disjunction_v<...>`

test/type-check.cc
CodingCanuck: Done.

test/type-check.cc
CodingCanuck: Done.

eventuals/type-check.h
CodingCanuck: Great tip, thanks! Done.

PR #443, opened 2022-06-24
Introduce 'backward-cpp' integration

ArthurBandaryk:
This PR integrates [`backward-cpp`](https://github.com/bombela/backward-cpp)
to print out a stack trace on failure.


test/BUILD.bazel
```diff
@@ -53,8 +55,8 @@ cc_test(
         "eventual.cc",
         "executor.cc",
         "expected.cc",
-        "failure-signal-handler.cc",
-        "failure-signal-handler.h",
+        #"failure-signal-handler.cc",
```
onelxj, line 58: restore please

test/backward-stacktrace.h
```diff
@@ -0,0 +1,42 @@
+#include <csignal>
+
+#include "backward.hpp"
+#include "glog/logging.h"
+#include "gtest/gtest.h"
+
+class BackwardStackTrace : public ::testing::Environment {
```
CodingCanuck, line 7: We should probably put this in a specific namespace, rather than implicitly adding it to the global namespace: https://google.github.io/styleguide/cppguide.html#Namespaces

CodingCanuck, line 7: Please add a comment documenting this class. (Why do we declare it in a header, rather than in the one .cc file that uses it?)

test/backward-stacktrace.h
```diff
@@ -0,0 +1,42 @@
...
+  void SetUp() override {
+    // '+' converts lambda to C-style function pointer.
+    // Nik has already used this here:
+    // (check_line_length skip)
+    // https://github.com/3rdparty/eventuals/blob/00f4ac7fe5a3ebf1749845274de5b23e7aca2806/eventuals/http.h#L720
+    //
+    // You can read about this using the link below:
+    // (check_line_length skip)
+    // https://stackoverflow.com/questions/70505162/why-would-one-want-to-put-a-unary-plus-operator-in-front-of-a-c-lambda/70506136#70506136
+    static auto signal_handler = +[](int signal) {
```
CodingCanuck, line 20: Please document what this handler is supposed to do. (Why does it exit?)

test/test_signal.cc
```diff
@@ -0,0 +1,14 @@
...
+void f(int i) {
+  if (i >= 10) {
+    CHECK(false);
+  } else {
+    f(i + 1);
+  }
+};
+
+TEST(SignalTestSTD, CheckFail) {
+  f(20);
```
CodingCanuck, line 13: Document what this test is supposed to do.

It looks like this test will fail, which looks undesirable. Do we want something like `EXPECT_DEATH` to ensure that this crashes with good output?

If that doesn't work, maybe we want this test to run a subprocess which crashes, then we could inspect its log output to make sure that it printed a good stack trace. Example of running a subprocess that we expect to crash: https://github.com/3rdparty/eventuals/blob/main/test/grpc/client-death-test.cc#L64

CodingCanuck (changes requested):

test/backward-stacktrace.cc
```diff
@@ -0,0 +1,8 @@
+#include "backward-stacktrace.h"
+
+// For setting up gtest environment in order to
+// provide our own stack traces for both LOG(FATAL)
+// and an arbitrary signal being raised.
+static testing::Environment* const backward_env =
```
CodingCanuck, line 6: `kBackwardEnv`

https://google.github.io/styleguide/cppguide.html#Constant_Names

test/backward-stacktrace.h
```diff
@@ -0,0 +1,31 @@
...
+class BackwardStackTrace : public ::testing::Environment {
+ public:
+  ~BackwardStackTrace() override {}
+
+  void SetUp() override {
+    // A simple helper class that registers for you the most
+    // common signals and other callbacks to segfault,
+    // hardware exception, un-handled exception etc.
+    // https://github.com/bombela/backward-cpp#signalhandling
+    static backward::SignalHandling sh;
```
CodingCanuck, line 25: Please comment on why this is static.

Can it instead be a member variable of BackwardStackTrace?

The style guide prohibits static variables with non-trivial destructors: https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables if this can't be a member variable, then next best would be to make this a static pointer (that never gets deleted)

test/backward-stacktrace.h
```diff
@@ -0,0 +1,31 @@
...
+// the tests if there was no fatal failures, and finally
+// calls TearDown() of the environment object. If there
+// was any failure - a simple helper class
+// `backward::SignalHandling` will register this for us
+// and print out consistent stacktrace.
+// (check_line_length skip)
+// https://github.com/YOU-i-Labs/googletest/blob/master/googletest/docs/V1_7_AdvancedGuide.md#global-set-up-and-tear-down
+class BackwardStackTrace : public ::testing::Environment {
+ public:
+  ~BackwardStackTrace() override {}
```
CodingCanuck, line 18: I would not add empty overrides that don't do anything, so remove this destructor and remove `TearDown()`

test/test_fail.cc
```diff
@@ -0,0 +1,16 @@
...
+void f(int i) {
+  if (i >= 10) {
+    CHECK(false);
+  } else {
+    f(i + 1);
+  }
+};
+
+// This file is just for demonstrating stacktrace on any failure.
+// Will be removed when merging!
```
CodingCanuck, line 13: Should we have a non-removed test, e.g. a `EXPECT_DEATH` test that checks for a stack trace? https://github.com/YOU-i-Labs/googletest/blob/master/googletest/docs/V1_7_AdvancedGuide.md#death-tests

CodingCanuck (approved):

PR #448, opened 2022-07-01
Propagate `Errors` through `k`

onelxj:


test/finally.cc
```diff
@@ -41,13 +41,11 @@ TEST(Finally, Fail) {
...
 
-  expected<int, std::exception_ptr> result = *e();
+  expected<int, std::variant<StoppedException, std::runtime_error>> result = *e();
 
   ASSERT_FALSE(result.has_value());
 
-  EXPECT_THAT(
-      [&]() { std::rethrow_exception(result.error()); },
-      ThrowsMessage<std::runtime_error>(StrEq("error")));
+  EXPECT_STREQ(std::get<1>(result.error()).what(), "error");
```
benh, line 48:
```suggestion
  ASSERT_EQ(result.error().index(), 1);
  EXPECT_STREQ(std::get<1>(result.error()).what(), "error");
```

eventuals/grpc/server.h
```diff
@@ -822,8 +850,19 @@ template <typename Request, typename Response>
...
+                                                   EVENTUALS_GRPC_LOG(1)
+                                                       << std::get<
+                                                              0>(e)
+                                                              .what();
+                                                 } else {
+                                                   EVENTUALS_GRPC_LOG(1)
+                                                       << std::get<
+                                                              1>(e)
+                                                              .what();
+                                                 }
```
benh, line 865: This is the core of what you're trying to do so no need to `Raise()` above! We were doing the `Raise` before because we had a `std::exception_ptr` but now we can just do this without the `Raise()` and `Catch()`. Both places in this file please!

eventuals/finally.h
```diff
@@ -45,15 +43,28 @@ struct _Finally final {
...
     template <typename Arg, typename Errors>
     using ErrorsFrom = std::tuple<>;
 
-    template <typename Arg, typename K>
+    template <typename Arg, typename Errors, typename K>
     auto k(K k) && {
-      return Continuation<K, Arg>{std::move(k)};
+      return Continuation<
+          K,
+          typename AddStopped<typename TupleToVariant<Errors>::type>::type,
```
benh, line 66: This will have to be tweaked as per above. How about a `using Expected = ` type alias so we don't have to repeat the long `std::conditional_t` and then we can just have `using ValueFrom = Expected<Arg>;`.

eventuals/finally.h
```diff
@@ -11,30 +11,28 @@ namespace eventuals {
 ////////////////////////////////////////////////////////////////////////
 
 struct _Finally final {
-  template <typename K_, typename Arg_>
+  template <typename K_, typename Errors_, typename Arg_>
```
benh, line 14: `Arg_` then `Errors_`!

eventuals/expected.h
```diff
@@ -96,10 +99,22 @@ class expected : public tl::expected<Value_, Error_> {
...
   auto k(K k) && {
     return ExpectedToEventual(std::move(*this))
-        .template k<Value_>(std::move(k));
+        .template k<
+            Value_,
+            tuple_types_union_t<
+                std::tuple<
+                    std::conditional_t<
+                        std::disjunction_v<
+                            CheckErrorsTypesForVariant<std::decay_t<Error_>>,
```
benh, line 111: I don't want to propagate `std::variant` for error types right now, so this basically needs to be a union of `Errors` and the error type of the `expected`, i.e., `Error_`. In the case we have `std::variant<E1, E2>` as the error type of the `expected` then we should extract each of those into a tuple and then union that with `Errors`. Make sense?

eventuals/expected.h
```diff
@@ -28,6 +29,7 @@ auto ExpectedToEventual(tl::expected<T, E>&& expected) {
       .template raises<
           std::conditional_t<
               std::disjunction_v<
+                  CheckErrorsTypesForVariant<std::decay_t<E>>,
                   std::is_base_of<std::exception, std::decay_t<E>>,
                   std::is_base_of<std::exception_ptr, std::decay_t<E>>>,
               E,
```
benh, line 35: I never want to "raise" a `std::variant` for now, so instead, let's extract out all of the error types. This is hard because `Eventual().raises<E1, E2>()` doesn't take a `std::tuple`, so to simplify your life how about we provide the following overload `ExpectedToEventual(tl::expected<T, std::variant<Errors...>>&& expected)` and then deal with the `std::variant` case explicitly!

eventuals/conditional.h
```diff
@@ -105,11 +106,19 @@ struct _Conditional {
...
+                             _Then::Adaptor<K_>>()));
 
-    using ElseAdapted_ = decltype(std::declval<ElseE_>().template k<void>(
-        std::declval<_Then::Adaptor<K_>>()));
+    using ElseAdapted_ =
+        decltype(std::declval<
+                     ElseE_>()
+                     .template k<void, Errors_>(
+                         std::declval<
+                             _Then::Adaptor<K_>>()));
```
benh, line 121: For less whitespace can you not wrap the `std::declval<ThenE_>()` and `std::declval<ElseE_>()`?

benh (changes requested):
Let's add a follow-up PR that eliminates propagating `std::exception_ptr` unless it's explicitly done via `.raises<std::exception_ptr>()` or `Raise(make_exception_ptr(...))`, etc. The only places we should be calling `std::make_exception_ptr` is when doing `promise.set_exception(...)`. We should be able to delete `make_exception_ptr_or_forward()`. Please let me know if this doesn't make sense!

benh (changes requested):
Anything that takes a different path if there is an error (e.g., `If`) should not need to propagate the errors!

Do we need to review all `ErrorsFrom` and either add more `tuple_types_union_t` or remove some in places where we are not correctly using the sub-continuation that we pass control  to.

What is is going to take to eliminate `std::exception_ptr`?

eventuals/generator.h
```diff
@@ -305,13 +307,13 @@ struct _Generator final {
...
-  template <typename From_, typename To_, typename Errors_, typename... Args_>
+  template <typename From_, typename To_, typename Catches_, typename Errors_, typename... Args_>
   struct Composable final {
-    template <typename Arg>
+    template <typename Arg, typename Errors>
     using ValueFrom = To_;
 
     template <typename Arg, typename Errors>
-    using ErrorsFrom = tuple_types_union_t<Errors, Errors_>;
+    using ErrorsFrom = tuple_types_union_t<Errors_, tuple_types_unique_t<Errors, Catches_>>;
```
benh, line 316: We want to subtract `Catches_` from `Errors` and then union that with `Errors_`!

test/type-traits.cc
```diff
@@ -278,6 +311,43 @@ static_assert(
...
+              std::variant<std::runtime_error, std::overflow_error>,
+              std::tuple<std::overflow_error, std::runtime_error>>);
+
+static_assert(!check_variant_types_in_tuple_v<
+              std::variant<
+                  std::runtime_error,
+                  std::overflow_error,
+                  std::bad_cast>,
+              std::tuple<std::overflow_error, std::runtime_error>>);
+
```
benh, line 350: Dead code!
```suggestion
```

eventuals/type-traits.h
```diff
@@ -235,6 +239,9 @@ struct tuple_types_union<
 template <typename Left, typename Right>
 using tuple_types_union_t = typename tuple_types_union<Left, Right>::type;
 
+template <typename Left, typename Right>
+using tuple_types_unique_t = typename tuple_types_union<Left, Right>::unique_left;
```
benh, line 243: Let's triple check that anything we add we use and test. Please keep this issue unresolved until we're about to commit this PR, thank you!

benh (changes requested):

eventuals/event-loop.h
```diff
@@ -1338,11 +1339,11 @@ struct _EventLoopSchedule final {
...
     Interrupt* interrupt_ = nullptr;
 
-    using Value_ = typename E_::template ValueFrom<Arg_>;
+    using Value_ = typename E_::template ValueFrom<Arg_, std::tuple<>>;
 
-    using Adapted_ = decltype(std::declval<E_>().template k<Arg_>(
+    using Adapted_ = decltype(std::declval<E_>().template k<Arg_, std::tuple<>>(
         std::declval<_Reschedule::Composable>()
-            .template k<Value_>(std::declval<_Then::Adaptor<K_>>())));
+            .template k<Value_, std::tuple<>>(std::declval<_Then::Adaptor<K_>>())));
```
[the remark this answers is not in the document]
benh: Same as what you did in `static-thread-pool.h`.

eventuals/generator.h
```diff
@@ -16,12 +17,12 @@ namespace eventuals {
...
 ////////////////////////////////////////////////////////////////////////
 
-template <typename E_, typename From_, typename To_>
+template <typename E_, typename From_, typename To_, typename Errors_>
 struct HeapGenerator final {
   struct Adaptor final {
     Adaptor(
         Callback<void(TypeErasedStream&)>* begin,
-        Callback<void(std::exception_ptr)>* fail,
+        Callback<void(typename StoppedOrVariantFrom<Errors_>::type)>* fail,
```
benh, line 25:
```suggestion
        Callback<function_type_t<void, variant_from_tuple_or_void_t<Errors_>>)>* fail,
```

eventuals/generator.h
benh: `tuple_types_subtract`

test/generator.cc
```diff
@@ -254,10 +254,9 @@ TEST(Generator, FailStream) {
             functions.done.Call();
           })
           .fail([&](auto& k, auto&& error) {
-            // No need to specify 'raises' because type of error is
-            // 'std::exception_ptr', that just propagates.
+            // Will not be invoked since we do not specify any of
+            // 'Catches' or 'Raises'.
             functions.fail.Call();
```
benh, line 259: If this won't be invoked then let's put in a `FAIL()` here instead! And I think this is just because we don't specify `Catches` right? Doesn't matter if we specify `Raises`.
```suggestion
            FAIL() << "Should not get called because we did not specify any 'Catches'";
```

test/grpc/helloworld.eventuals.cc
```diff
@@ -66,7 +66,7 @@ Task::Of<void> Greeter::TypeErasedService::Serve() {
                >> Loop())
         >> Finally([&](auto&& expected) {
              if (!expected.has_value()) {
-               LOG(WARNING) << "Failed to serve: " << What(expected.error());
+               //  LOG(WARNING) << "Failed to serve: " << What(expected.error());
```
benh, line 69: Did you forget to uncomment this?

test/task.cc
```diff
@@ -76,6 +77,37 @@ TEST(Task, Void) {
...
+  auto f = []() -> Task::Of<int>::Catches<std::runtime_error> {
+    return []() {
+      return Finally([](expected<
+                         void,
+                         std::variant<
+                             Stopped,
+                             std::runtime_error>>&& expected) {
+               return Just(100);
+             })
+          >> Just(10);
```
benh, line 90: How about we remove this so that we are confident that we execute the `Finally`?
```suggestion
             });
```

test/BUILD.bazel
```diff
@@ -97,7 +97,7 @@ cc_test(
         ":http-mock-server",
         ":promisify-for-test",
         "//eventuals",
-        "//test/concurrent",
+        # "//test/concurrent",
```
benh, line 100: Let's not forget to resolve this before we commit!

eventuals/concurrent.h
```diff
@@ -158,16 +158,17 @@ struct _Concurrent final {
     // addressed before doing so and we should ensure that we see
     // further speed ups that are beyond what we get already from
     // putting all of these in 'TypeErasedAdaptor'.
-    [[nodiscard]] auto CreateOrReuseFiber() {
-      return Synchronized(Then([this]() {
+    template <typename Exception>
+    [[nodiscard]] auto CreateOrReuseFiber(const Exception& exception) {
```
benh, line 162:
```suggestion
    template <typename... Errors>
    [[nodiscard]] auto CreateOrReuseFiber(
        stout::borrowed_ref<
            std::optional<typename StoppedOrVariantFrom<Errors...>::type>>&& stopped_or_error) {
```

eventuals/concurrent.h
```diff
@@ -158,16 +158,17 @@ struct _Concurrent final {
...
+      return Synchronized(Then([this, &exception]() {
```
benh, line 163:
```suggestion
      return Synchronized(Then([this, stopped_or_error = std::move(stopped_or_error)]() {
```

eventuals/concurrent.h
```diff
@@ -158,16 +158,17 @@ struct _Concurrent final {
...
         // As long as downstream isn't done, or we've been interrupted,
         // or have encountered an error, then add a new fiber if none
         // exist, otherwise trim done fibers from the front, then look
         // for a done fiber to reuse, or if all fibers are not done add
         // a new one.
         TypeErasedFiber* fiber = nullptr;
 
-        if (!(downstream_done_ || interrupted_ || exception_)) {
+        if (!(downstream_done_ || interrupted_ || exception)) {
```
benh, line 171:
```suggestion
        if (!(downstream_done_ || interrupted_ || stopped_or_error->has_value())) {
```

eventuals/concurrent.h
```diff
@@ -213,7 +214,8 @@ struct _Concurrent final {
     // NOTE: this could also be moved to a .cc file but see the
     // comments above 'CreateOrReuseFiber()' above for what's
     // preventing us from doing so.
-    [[nodiscard]] auto IngressEpilogue() {
+    template <typename Exception>
+    [[nodiscard]] auto IngressEpilogue(Exception& exception) {
```
benh, line 218:
```suggestion
    template <typename... Errors>
    [[nodiscard]] auto IngressEpilogue(
        stout::borrowed_ref<
            std::optional<typename StoppedOrVariantFrom<Errors...>::type>>&& stopped_or_error) {
```

eventuals/concurrent.h
```diff
@@ -213,7 +214,8 @@ struct _Concurrent final {
...
       return Synchronized(
           Eventual<void>()
               .start([this](auto& k) {
```
benh, line 221:
```suggestion
          Eventual<void>()
              .context(std::move(stopped_or_error))
              .start([this](auto& /* stopped_or_error */, auto& k) {
```

eventuals/concurrent.h
```diff
@@ -228,12 +230,11 @@ struct _Concurrent final {
 
                 k.Start(); // Exits the synchronized block!
               })
-              .fail([this](auto& k, auto&& error) {
+              .fail([this, &exception](auto& k, auto&& error) {
```
benh, line 233:
```suggestion
              .fail([this](auto& stopped_or_error, auto& k, auto&& error) {
```

eventuals/concurrent.h
```diff
@@ -228,12 +230,11 @@ struct _Concurrent final {
...
                 upstream_done_ = true;
 
-                if (!exception_) {
-                  exception_ = make_exception_ptr_or_forward(
-                      std::forward<decltype(error)>(error));
+                if (!exception) {
+                  exception.emplace(std::forward<decltype(error)>(error));
```
benh, line 237:
```suggestion
                if (!stopped_or_error->has_value()) {
                  stopped_or_error->emplace(std::forward<decltype(error)>(error));
```

eventuals/concurrent.h
```diff
@@ -245,12 +246,11 @@ struct _Concurrent final {
 
                 k.Start(); // Exits the synchronized block!
               })
-              .stop([this](auto& k) {
+              .stop([this, &exception](auto& k) {
```
benh, line 249:
```suggestion
              .stop([this](auto& stopped_or_error, auto& k) {
```

eventuals/concurrent.h
```diff
@@ -245,12 +246,11 @@ struct _Concurrent final {
...
                 upstream_done_ = true;
 
-                if (!exception_) {
-                  exception_ = std::make_exception_ptr(
-                      eventuals::StoppedException());
+                if (!exception) {
+                  exception.emplace(eventuals::Stopped());
```
benh, line 253:
```suggestion
                if (!stopped_or_error->has_value()) {
                  stopped_or_error->emplace(eventuals::Stopped());
```

eventuals/concurrent.h
```diff
@@ -270,7 +270,8 @@ struct _Concurrent final {
     // NOTE: this could also be moved to a .cc file but see the
     // comments above 'CreateOrReuseFiber()' above for what's
     // preventing us from doing so.
-    [[nodiscard]] auto FiberEpilogue(TypeErasedFiber* fiber) {
+    template <typename Exception>
+    [[nodiscard]] auto FiberEpilogue(TypeErasedFiber* fiber, Exception& exception) {
```
benh, line 274:
```suggestion
    template <typename... Errors>
    [[nodiscard]] auto FiberEpilogue(
        TypeErasedFiber* fiber, 
        stout::borrowed_ref<
            std::optional<typename StoppedOrVariantFrom<Errors...>::type>>&& stopped_or_error) {
```

eventuals/concurrent.h
```diff
@@ -270,7 +270,8 @@ struct _Concurrent final {
...
       return Synchronized(
           Eventual<void>()
               .start([this, fiber](auto& k) {
```
benh, line 277:
```suggestion
          Eventual<void>()
              .context(std::move(stopped_or_error))
              .start([this, fiber](auto& /* stopped_or_error */, auto& k) {
```

eventuals/concurrent.h
```diff
@@ -285,12 +286,15 @@ struct _Concurrent final {
 
                 k.Start(); // Exits the synchronized block!
               })
-              .fail([this, fiber](auto& k, auto&& error) {
+              .fail([this, fiber, &exception](auto& k, auto&& error) {
```
benh, line 289:
```suggestion
              .fail([this, fiber](auto& stopped_or_error, auto& k, auto&& error) {
```

eventuals/concurrent.h
```diff
@@ -285,12 +286,15 @@ struct _Concurrent final {
...
                 fiber->done = true;
 
-                if (!exception_) {
-                  exception_ = make_exception_ptr_or_forward(
-                      std::forward<decltype(error)>(error));
+                if (!exception) {
```
benh, line 292:
```suggestion
                if (!stopped_or_error->has_value()) {
```

eventuals/concurrent.h
```diff
@@ -285,12 +286,15 @@ struct _Concurrent final {
...
+                  // Can't provide compile-time guarantees that it won't be executed at runtime, so
+                  // have to check explicitly.
+                  if constexpr (!std::is_same_v<std::decay_t<decltype(exception.value())>, Stopped>) {
+                    exception.emplace(std::forward<decltype(error)>(error));
+                  }
```
benh, line 297: I'm not following why you need to do this `if constexpr` here or what the comment is trying to say?

eventuals/concurrent.h
```diff
@@ -302,12 +306,11 @@ struct _Concurrent final {
 
                 k.Start(); // Exits the synchronized block!
               })
-              .stop([this, fiber](auto& k) {
+              .stop([this, fiber, &exception](auto& k) {
```
benh, line 309:
```suggestion
              .stop([this, fiber](auto& stopped_or_error, auto& k) {
```

eventuals/concurrent.h
```diff
@@ -302,12 +306,11 @@ struct _Concurrent final {
...
                 fiber->done = true;
 
-                if (!exception_) {
-                  exception_ = std::make_exception_ptr(
-                      eventuals::StoppedException());
+                if (!exception) {
+                  exception.emplace(eventuals::Stopped());
```
benh, line 313:
```suggestion
                if (!stopped_or_error->has_value()) {
                  stopped_or_error->emplace(eventuals::Stopped());
```

eventuals/concurrent.h
```diff
@@ -478,7 +477,7 @@ struct _Concurrent final {
                notify_egress_();
              }))
           >> Loop()
-          >> FiberEpilogue(fiber)
+          >> FiberEpilogue(fiber, exception_)
```
benh, line 480:
```suggestion
          >> FiberEpilogue(fiber, stopped_or_error_.Borrow())
```

eventuals/concurrent.h
```diff
@@ -513,7 +512,7 @@ struct _Concurrent final {
     // upstream value.
     [[nodiscard]] auto Ingress() {
       return Map(Let([this](Arg_& arg) {
-               return CreateOrReuseFiber()
+               return CreateOrReuseFiber(exception_)
```
benh, line 515:
```suggestion
               return CreateOrReuseFiber(stopped_or_error_.Borrow())
```

eventuals/concurrent.h
```diff
@@ -529,7 +528,7 @@ struct _Concurrent final {
              }))
           >> Until([](bool done) { return done; })
           >> Loop() // Eagerly try to get next value to run concurrently!
-          >> IngressEpilogue()
+          >> IngressEpilogue(exception_)
```
benh, line 531:
```suggestion
          >> IngressEpilogue(stopped_or_error_.Borrow())
```

eventuals/concurrent.h
```diff
@@ -555,12 +554,13 @@ struct _Concurrent final {
                          .start([this](auto& k) {
                            if (exception_ && upstream_done_ && fibers_done_) {
```
benh, line 555:
```suggestion
                           if (stopped_or_value_->has_value() && upstream_done_ && fibers_done_) {
```

eventuals/concurrent.h
```diff
@@ -555,12 +554,13 @@ struct _Concurrent final {
...
-                             try {
-                               std::rethrow_exception(*exception_);
-                             } catch (const StoppedException&) {
+                             if constexpr (!std::is_same_v<std::decay_t<decltype(exception_.value())>, Stopped>) {
+                               std::visit([this, &k](auto&& error) {
+                                 k.Fail(std::forward<decltype(error)>(error));
+                               },
+                                          exception_.value());
+                             } else {
                                k.Stop();
```
benh, line 563: I'm not sure the code you have will do what you want ... basically you're asking for the type of the `stopped_or_error_` at run-time but then trying to decide what you do with it at compile-time? I think we need to get the run-time value and add in checks at compile-time that if that run-time value is `Stopped` then do something different:
```suggestion
                            std::visit(
                                 [&k](auto&& error) {
                                   if constexpr (!std::is_same_v<std::decay_t<decltype(error)>, Stopped>) {
                                     k.Fail(std::forward<decltype(error)>(error));
                                   } else {
                                     k.Stop();
                                   }
                                 },
                                 stopped_or_value_->value());
```

eventuals/concurrent.h
```diff
@@ -555,12 +554,13 @@ struct _Concurrent final {
                          .start([this](auto& k) {
                            if (exception_ && upstream_done_ && fibers_done_) {
                              // TODO(benh): flush remaining values first?
-                             try {
-                               std::rethrow_exception(*exception_);
-                             } catch (const StoppedException&) {
+                             if constexpr (!std::is_same_v<std::decay_t<decltype(exception_.value())>, Stopped>) {
+                               std::visit([this, &k](auto&& error) {
```
benh, line 558: What do you need to capture `this` for?

benh (changes requested):

eventuals/eventual.h
```diff
@@ -243,8 +241,10 @@ struct _Eventual {
           Stop_,
           Interruptible_,
           Value_,
-          Errors_>(
-          Reschedulable<K, Value_>{std::move(k)},
+          tuple_types_union_t<Errors_, Errors>>(
```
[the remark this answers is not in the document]
onelxj: Task.FailOnCallback test - the second eventual just propagates an error, but we can't define it and seems that we don't want to raise anything there

eventuals/generator.h
```diff
@@ -387,21 +394,33 @@ struct _Generator final {
...
           "eventual result type can not be converted into type of 'Generator'");
 
       dispatch_ = [f = std::move(f)](
                       Action action,
-                      std::optional<std::exception_ptr>&& exception,
+                      std::optional<
+                          typename StoppedOrVariantFrom<
+                              tuple_types_union_t<
+                                  Errors_,
+                                  Catches_>>::type>&& exception,
```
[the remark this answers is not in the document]
onelxj: Imagine if we specify `Raises`, we raise an error from inside of `Generator`, so HeapGenerator needs to capture it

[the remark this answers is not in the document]
onelxj: Let's chat about all union types for `Generator` and `Task`.

eventuals/loop.h
```diff
@@ -305,7 +305,7 @@ struct _Loop final {
           Stop_,
           Interruptible_,
           Value_,
-          Errors_>(
+          tuple_types_union_t<Errors, Errors_>>(
```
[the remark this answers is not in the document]
onelxj: Generator.FailStream

eventuals/generator.h
```diff
@@ -38,14 +40,11 @@ struct HeapGenerator final {
...
-          make_exception_ptr_or_forward(
-              std::forward<Error>(error)));
-    }
-
-    // NOTE: overload so we don't create nested std::exception_ptr.
-    void Fail(std::exception_ptr exception) {
-      (*fail_)(std::move(exception));
+      // Using to avoid compile error when we do both 'Raise' and 'Catch'
+      // inside a task.
+      if constexpr (std::tuple_size_v<Catches_> != 0) {
```
benh, line 45: As discussed, this isn't the right spot for this!

eventuals/type-traits.h
```diff
@@ -298,6 +303,104 @@ using apply_tuple_types_t = typename apply_tuple_types<T, Tuple>::type;
...
+
+template <typename T>
+using has_operator = decltype(&T::operator());
+
+template <typename T>
+using is_default_lambda =
+    std::experimental::is_detected<has_operator, T>;
+
+////////////////////////////////////////////////////////////////////////
+
```
benh, line 403: Just flagging that this is likely dead code now given our discussion in `catch.h`.

test/catch.cc
```diff
@@ -163,14 +157,8 @@ TEST(CatchTest, UnexpectedAll) {
...
                })
-               .all([](std::exception_ptr&& error) {
-                 try {
-                   std::rethrow_exception(error);
-                 } catch (const Error& e) {
-                   EXPECT_STREQ(e.what(), "child exception");
-                 } catch (...) {
-                   ADD_FAILURE() << "Failure on rethrowing";
-                 }
+               .all([](auto&& error) {
```
benh, line 160: Let's please spell out this type!

test/catch.cc
```diff
@@ -163,14 +157,8 @@ TEST(CatchTest, UnexpectedAll) {
...
+                 EXPECT_STREQ(std::get<0>(error).what(), "child exception");
```
benh, line 161: Let's use `std::get` with the type please!

test/eventual.cc
```diff
@@ -261,9 +261,9 @@ TEST(EventualTest, CatchVoid) {
   auto e = []() {
     return Just()
         >> Raise("error")
-        >> Catch(Let([](auto& error) {
+        >> Catch(Let([](std::variant<std::runtime_error>& error) {
              return Then([&]() {
-               EXPECT_EQ("error", What(error));
+               EXPECT_STREQ(std::get<0>(error).what(), "error");
```
benh, line 266: Please use `std::get` with type.

benh (changes requested):

eventuals/task.h
```diff
@@ -19,12 +20,13 @@ namespace eventuals {
...
 
-template <typename E_, typename From_, typename To_>
+template <typename E_, typename From_, typename To_, typename Catches_, typename Raises_>
 struct HeapTask final {
   struct Adaptor final {
     Adaptor(
         Callback<function_type_t<void, To_>>* start,
-        Callback<void(std::exception_ptr)>* fail,
+        Callback<void(
+            typename VariantOfStoppedAndErrors<Raises_>::type)>* fail,
```
benh, line 29: Ok, getting back to a previous comment I made, this should not be a `std::variant<Stopped, Raises_...>`? We have the `stop` callback, should this just be a `std::variant` of `Raises_`.

And to handle an empty `std::variant` we should use `function_type_t`. So we'll want something like:
```suggestion
        Callback<void(variant_of_type_and_tuple_t<std::monostate, Raises_>&&)>* fail,
```
Then in `type-traits.h`:
```cpp
template <typename Type, typename Tuple>
using variant_of_type_and_tuple_t = apply_tuple_types_t<
    std::variant, 
    tuple_types_concatenate_t<std::tuple<Type>, Tuple>>;
```

eventuals/task.h
```diff
@@ -89,9 +85,10 @@ struct HeapTask final {
 
   void Fail(
       Interrupt& interrupt,
-      std::exception_ptr&& exception,
+      typename VariantOfStoppedAndErrors<Catches_>::type&& error,
```
benh, line 88:
```suggestion
      variant_of_type_and_tuple_t<std::monostate, Catches_>      
```

eventuals/task.h
```diff
@@ -19,12 +20,13 @@ namespace eventuals {
 
 ////////////////////////////////////////////////////////////////////////
 
-template <typename E_, typename From_, typename To_>
+template <typename E_, typename From_, typename To_, typename Catches_, typename Raises_>
 struct HeapTask final {
   struct Adaptor final {
     Adaptor(
         Callback<function_type_t<void, To_>>* start,
```
benh, line 27:
```suggestion
        Callback<function_type_t<void, To_&&>>* start,
```

eventuals/task.h
```diff
@@ -241,12 +244,11 @@ struct _TaskFromToWith final {
 
     template <typename Error>
     void Fail(Error&& error) {
-      std::exception_ptr exception;
-
-      exception = make_exception_ptr_or_forward(
-          std::forward<Error>(error));
-
-      Dispatch(Action::Fail, std::nullopt, std::move(exception));
+      if constexpr (tuple_contains_exact_type_v<Error, Catches_>) {
```
benh, line 247:
```suggestion
      // NOTE: we only propagate an upstream error into our type-erased eventual
      // if the eventual catches the error, otherwise we skip it all together.
      if constexpr (tuple_contains_exact_type_v<Error, Catches_>) {
```

eventuals/task.h
```diff
@@ -385,23 +400,46 @@ struct _TaskFromToWith final {
...
-                               std::optional<std::exception_ptr>&& exception,
+                               std::optional<
+                                   typename VariantOfStoppedAndErrors<
+                                       Catches_>::type>&& error,
                                Args_&... args,
                                std::optional<MonostateIfVoidOr<From_>>&& arg,
-                               std::unique_ptr<void, Callback<void(void*)>>& e_,
+                               std::unique_ptr<
+                                   void,
+                                   Callback<void(void*)>>& e_,
```
benh, line 410: This formatting change looks unnecessary, can we please revert?

eventuals/task.h
```diff
@@ -369,14 +383,15 @@ struct _TaskFromToWith final {
...
+      using ErrorsFromE = typename E::template ErrorsFrom<From_, Catches_>;
 
       static_assert(
-          tuple_types_subset_subtype_v<ErrorsFromE, Errors_>,
+          tuple_types_subset_subtype_v<ErrorsFromE, Raises_>,
           "Specified errors can't be raised within 'Task'");
 
+      using Value = typename E::template ValueFrom<From_, Catches_>;
+
+
```
benh, line 394:
```suggestion

```

eventuals/task.h
```diff
@@ -339,7 +352,8 @@ struct _TaskFromToWith final {
               HAS_ARGS,
               std::true_type,
               std::is_invocable<F>>::value,
-          "'Task' expects a callable (e.g., a lambda) that takes no arguments");
+          "'Task' expects a callable (e.g., a lambda) "
+          "that takes no arguments");
```
benh, line 356: This formatting change looks unnecessary, can we please revert?

eventuals/task.h
```diff
@@ -261,23 +263,31 @@ struct _TaskFromToWith final {
...
-                [this](std::exception_ptr error) {
-                  k_.Fail(std::move(error));
+                [this](
+                    typename VariantOfStoppedAndErrors<
+                        Raises_>::type errors) {
+                  std::visit(
+                      [this](auto&& error) {
+                        k_.Fail(std::forward<decltype(error)>(error));
+                      },
+                      errors);
```
benh, line 290:
```suggestion
                      std::move(error));
```

eventuals/task.h
```diff
@@ -261,23 +263,31 @@ struct _TaskFromToWith final {
...
                 e_,
                 *interrupt_,
                 [this](auto&&... args) {
                   k_.Start(std::forward<decltype(args)>(args)...);
                 },
-                [this](std::exception_ptr error) {
-                  k_.Fail(std::move(error));
+                [this](
+                    typename VariantOfStoppedAndErrors<
+                        Raises_>::type errors) {
```
benh, line 285:
```suggestion
                    variant_of_type_and_tuple_t<std::monostate, Raises_>&& error) {
```

eventuals/task.h
```diff
@@ -261,23 +263,31 @@ struct _TaskFromToWith final {
...
+                  std::visit(
+                      [this](auto&& error) {
+                        k_.Fail(std::forward<decltype(error)>(error));
```
benh, line 288:
```suggestion
                        if constexpr (!std::is_same_v<std::monostate, std::decay_t<decltype(error)>>) {
                          k_.Fail(std::forward<decltype(error)>(error));
                        }
```

eventuals/task.h
```diff
@@ -412,14 +450,21 @@ struct _TaskFromToWith final {
...
-                interrupt,
-                std::move(exception.value()),
-                std::move(start),
-                std::move(fail),
-                std::move(stop));
+          case Action::Fail: {
+            // If 'Catches_' is not specified, that means that we
+            // will never invoke 'e->Fail', but we are using not constexpr
+            // Action as well, so compiler might try to go to 'e->Fail' path
+            // with wring error types, which might cause some errors.
```
benh, line 457:
```suggestion
            // If 'Catches_' is empty then we will never dispatch
            // with an action of 'Action::Fail' but since 'action' is a runtime value
            // the compiler will assume that it's possible that we can call
            // 'e->Fail()', with what ever type we have for 'error' so to keep
            // the compiler from trying to compile that code path we need to
            // add the following 'if constexpr'.
```

eventuals/task.h
```diff
@@ -600,17 +654,21 @@ class _Task final {
...
-          promise_->set_exception(
-              make_exception_ptr_or_forward(
-                  std::forward<decltype(error)>(error)));
+          std::visit(
+              [this](auto&& error) {
+                promise_->set_exception(
+                    std::make_exception_ptr(
+                        std::forward<decltype(error)>(error)));
+              },
+              error);
```
benh, line 663:
```suggestion
              std::forward<decltype(error)>(error));
```

eventuals/task.h
```diff
@@ -600,17 +654,21 @@ class _Task final {
         // from within 'this' and borrowing could create a cycle.
         [this](auto&& error) {
           CHECK(promise_.has_value());
-          promise_->set_exception(
-              make_exception_ptr_or_forward(
-                  std::forward<decltype(error)>(error)));
+          std::visit(
+              [this](auto&& error) {
+                promise_->set_exception(
```
benh, line 659:
```suggestion
                static_assert(
                    std::is_same_v<std::exception_ptr, std::decay_t<decltype(error)>>,
                    "PLEASE USE THE SAME ERROR MESSAGE WE ARE USING OTHER PLACES THAT WE ARE DOING THIS WITH PROMISES");
                promise_->set_exception(
```

benh (changes requested):

eventuals/concurrent.h
```diff
@@ -552,16 +570,27 @@ struct _Concurrent final {
...
+                                           Stopped>) {
+                                     k.Stop();
+                                   } else {
+                                     k.Fail(
+                                         std::forward<
+                                             decltype(stopped_or_error)>(
+                                             stopped_or_error));
+                                   }
+                                 },
+                                 stopped_or_error_->value());
```
benh, line 593:
```suggestion
                                 std::move(stopped_or_error_->value()));
```

eventuals/concurrent.h
```diff
@@ -710,13 +744,28 @@ struct _Concurrent final {
...
+
+    using UpstreamErrorsAndErrorsFromE_ = tuple_types_union_t<
+        typename E_::template ErrorsFrom<Arg_, std::tuple<>>,
+        Errors_>;
+
+    stout::Borrowable<
+        std::optional<
+            typename VariantOfStoppedAndErrors<
+                UpstreamErrorsAndErrorsFromE_>::type>>
+        stopped_or_error_;
```
benh, line 760:
```suggestion
    using E_ = typename std::invoke_result_t<F_>;

    using UpstreamErrorsAndErrorsFromE_ = tuple_types_union_t<
        typename E_::template ErrorsFrom<Arg_, std::tuple<>>,
        Errors_>;

    // Indicates whether the continuation was stopped or we encountered
    // an error, either from upstream, or from the eventual returned from 'F_'.
    stout::Borrowable<
        std::optional<
            typename VariantOfStoppedAndErrors<
                UpstreamErrorsAndErrorsFromE_>::type>>
        stopped_or_error_;
```

benh (changes requested):

test/finally.cc
```diff
@@ -37,48 +43,56 @@ TEST(Finally, Fail) {
...
+      std::variant<Stopped>>
+      result = *e();
 
   ASSERT_FALSE(result.has_value());
 
-  EXPECT_THROW(
-      std::rethrow_exception(result.error()),
-      eventuals::StoppedException);
+  EXPECT_STREQ(
+      std::get<Stopped>(result.error()).what(),
```
benh, line 83: Please don't forget to make sure that `Stopped` does not inherit from `std::exception`, thanks!

test/finally.cc
```diff
@@ -129,20 +147,25 @@ TEST(Finally, FinallyInsideThen) {
...
+                          })
                           .no([e = std::move(e)]() {
-                            return Raise(std::move(e.error()))
+                            CHECK(std::holds_alternative<std::runtime_error>(e.error()));
+                            return Raise(std::get<std::runtime_error>(std::move(e.error())))
                                 >> Catch()
-                                       .raised<std::exception>(
-                                           [](std::exception&& e) {
+                                       .raised<std::runtime_error>(
+                                           [](std::runtime_error e) {
```
benh, line 164:
```suggestion
                                           [](std::runtime_error&& error) {
```

test/finally.cc
```diff
@@ -87,37 +101,41 @@ TEST(Finally, VoidFail) {
...
+  EXPECT_STREQ(std::get<std::runtime_error>(result.error()).what(), "error");
 }
 
 TEST(Finally, VoidStop) {
   auto e = []() {
     return Eventual<void>([](auto& k) {
              k.Stop();
            })
-        >> Finally([](expected<void, std::exception_ptr>&& exception) {
+        >> Finally([](expected<void, std::variant<Stopped>>&& exception) {
```
benh, line 127:
```suggestion
        >> Finally([](expected<void, std::variant<Stopped>>&& error) {
```

test/finally.cc
```diff
@@ -87,37 +101,41 @@ TEST(Finally, VoidFail) {
   auto e = []() {
     return Just()
         >> Raise("error")
-        >> Finally([](expected<void, std::exception_ptr>&& exception) {
+        >> Finally([](expected<
+                       void,
+                       std::variant<
+                           Stopped,
+                           std::runtime_error>>&& exception) {
```
benh, line 108:
```suggestion
                           std::runtime_error>>&& expected) {
```

test/grpc/helloworld.eventuals.cc
```diff
@@ -64,9 +66,13 @@ Task::Of<void> Greeter::TypeErasedService::Serve() {
                    }));
                  })
                >> Loop())
-        >> Finally([&](auto&& expected) {
+        >> Finally([&](expected<std::tuple<std::monostate>, std::variant<Stopped, std::runtime_error>> expected) {
```
benh, line 69:
```suggestion
        >> Finally([&](expected<std::tuple<std::monostate>, std::variant<Stopped, std::runtime_error>>&& expected) {
```

test/grpc/helloworld.eventuals.cc
```diff
@@ -64,9 +66,13 @@ Task::Of<void> Greeter::TypeErasedService::Serve() {
...
              if (!expected.has_value()) {
-               LOG(WARNING) << "Failed to serve: " << What(expected.error());
+               std::visit(
+                   [](auto& error) {
+                     LOG(WARNING) << "Failed to serve: " << error.what();
+                   },
+                   expected.error());
```
benh, line 75:
```suggestion
                   std::move(expected.error()));
```

test/task.cc
```diff
@@ -299,31 +390,27 @@ TEST(Task, StartFuture) {
...
 
   task->Fail(
       GenerateTestTaskName(),
       std::runtime_error("error"),
       [](int) {
         FAIL() << "test should not have succeeded";
       },
-      [&](std::exception_ptr exception) {
-        result = std::move(exception);
+      [&](std::variant<Stopped, std::runtime_error>&& exception) {
```
benh, line 411: This looks wrong, we should only be passing the variant of errors here, not a possible stopped! That's what we have the next callback for.

test/task.cc
```diff
@@ -336,9 +423,7 @@ TEST(Task, FailContinuation) {
               std::tuple<>>,
           std::tuple<std::runtime_error>>);
 
-  EXPECT_THAT(
-      [&]() { std::rethrow_exception(result); },
-      ThrowsMessage<std::runtime_error>(StrEq("error")));
+  EXPECT_STREQ(result.value().what(), "error");
```
benh, line 426: Let's just do this in the lambda above rather than copying it outside via a `std::optional`.

benh (changes requested):

eventuals/then.h
```diff
@@ -14,30 +14,25 @@ template <typename T, typename Arg>
 using ValueFromMaybeComposable = typename std::conditional_t<
     !HasValueFrom<T>::value,
     decltype(Eventual<T>()),
-    T>::template ValueFrom<Arg>;
+    T>::template ValueFrom<Arg, std::tuple<>>;
```
benh, line 17: I'd actually like to keep `ValueFromMaybeComposable` a generic and reusable template, so let's instead update it to take in `Errors` just like `ErrorsFromComposable` and `ErrorsFromComposed` does and pass those errors here. Then when we use it down in the `Then` implementation we can pass `std::tuple<>` there.

That way these can possibly be used other places and it's only the call to them by the `Then` code that is saying that the errors we'd propagate in here is `std::tuple<>`. Make sense?

eventuals/then.h
```diff
@@ -14,30 +14,25 @@ template <typename T, typename Arg>
...
+template <typename T, typename Arg>
 using ErrorsFromMaybeComposable = typename ErrorsFromComposed<
     Arg,
     std::conditional_t<
         !HasValueFrom<T>::value,
         decltype(Just()),
         T>,
-    decltype(Just()),
-    Errors>::Errors;
+    decltype(Just())>::Errors;
```
benh, line 35: Let's revert all of this, see comment above!

eventuals/then.h
```diff
@@ -95,16 +90,14 @@ struct _Then final {
                     std::is_void_v<Arg>,
                     std::invoke_result<F_>,
                     std::invoke_result<F_, Arg>>::type,
-                void,
-                Errors>>;
-
+                void>>;
```
benh, line 93:
```suggestion
                void,
                std::tuple<>>>;
```
And we also need to pass `std::tuple<>` in `ValueFromMaybeComposable` above too!

benh (changes requested):

eventuals/task.h
onelxj: exceed 80 characters

eventuals/task.h
onelxj: exceed 80 characters

test/task.cc
```diff
@@ -299,31 +390,27 @@ TEST(Task, StartFuture) {
 }
 
 TEST(Task, FailContinuation) {
-  auto e = []() -> Task::Of<int> {
+  auto e = []() -> Task::Of<int>::Raises<std::runtime_error> {
```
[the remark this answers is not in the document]
onelxj: let's chat about it

eventuals/catch.h
```diff
@@ -225,26 +268,52 @@ struct _Catch final {
...
       // and then return 'Continuation'.
       return std::apply(
           [&](auto&&... catch_handler) {
             return Continuation<
                 K,
-                decltype(std::move(catch_handler).template Convert<K>())...>(
+                std::conditional_t<
+                    has_all_,
+                    std::tuple<>,
+                    tuple_types_subtract_t<Errors, Catches_>>,
```
benh, line 309:
```suggestion
```

eventuals/catch.h
```diff
@@ -275,33 +341,19 @@ struct _Catch final {
...
           std::tuple_cat(
               std::move(catch_handlers_),
-              std::tuple{Handler<Undefined, Error, F>{std::move(f)}}));
+              std::tuple{
+                  Handler<Undefined, Error, F, std::tuple<>>{std::move(f)}}));
     }
 
     template <typename F>
     auto all(F f) {
       static_assert(!has_all_, "Duplicate 'all'");
```
benh, line 350:
```suggestion
      static_assert(!has_all_, "Duplicate 'all'");
      
      static_assert(
          !tuple_types_contains_v<std::exception, Catches_>,
          "You already have a handler that catches 'std::exception'
          so your '.all()' handler is redundant");
```

eventuals/catch.h
```diff
@@ -275,33 +341,19 @@ struct _Catch final {
       return create<Unify_<Value_, Value>, has_all_>(
           std::tuple_cat(
               std::move(catch_handlers_),
-              std::tuple{Handler<Undefined, Error, F>{std::move(f)}}));
+              std::tuple{
+                  Handler<Undefined, Error, F, std::tuple<>>{std::move(f)}}));
```
benh, line 345:
```suggestion
                  Handler<Undefined, Error, F>{std::move(f)}}));
```

eventuals/catch.h
```diff
@@ -275,33 +341,19 @@ struct _Catch final {
...
-              std::is_void<Value_>>,
-          "Catch handlers do not return an eventual value of the same type");
-
-      return create<Unify_<Value_, Value>, true>(
+      return create<Value_, true>(
           std::tuple_cat(
               std::move(catch_handlers_),
               std::tuple{
-                  Handler<Undefined, std::exception_ptr, F>{std::move(f)}}));
+                  Handler<Undefined, std::monostate, F, std::tuple<>>{std::move(f)}}));
```
benh, line 356:
```suggestion
                  Handler<Undefined, std::monostate, F>{std::move(f)}}));
```

eventuals/type-traits.h
```diff
@@ -298,6 +303,85 @@ using apply_tuple_types_t = typename apply_tuple_types<T, Tuple>::type;
...
+template <typename>
+struct check_variant_errors;
+
+template <typename... Errors>
+struct check_variant_errors<std::variant<Errors...>> {
+  static constexpr bool value = check_errors_v<Errors...>;
+};
+
+template <typename Variant>
+inline constexpr bool check_variant_errors_v = check_variant_errors<Variant>::value;
```
benh, line 374: I do not see a use case for any of these.
```suggestion
```

test/catch.cc
```diff
@@ -86,12 +86,39 @@ TEST(CatchTest, All) {
...
                  return 10;
                })
-               .all([](std::exception_ptr&& error) {
-                 try {
-                   std::rethrow_exception(error);
-                 } catch (const std::runtime_error& error) {
-                   EXPECT_STREQ(error.what(), "10");
-                 }
+               .all([](std::variant<std::runtime_error>&& error) {
+                 EXPECT_STREQ(std::get<0>(error).what(), "10");
```
benh, line 90: No more please unless we use explicit indexing like in `do-all.h`!!!
```suggestion
                 EXPECT_STREQ(std::get<std::runtime_error>(error).what(), "10");
```

test/catch.cc
```diff
@@ -86,12 +86,39 @@ TEST(CatchTest, All) {
...
+TEST(CatchTest, AllRaisedOneException) {
+  auto e = []() {
+    return Just(500)
+        >> Raise(std::runtime_error("runtime_error"))
+        >> Raise(std::underflow_error("underflow_error"))
+        >> Catch()
+               .raised<std::overflow_error>([](std::overflow_error&& error) {
+                 ADD_FAILURE() << "Encountered unexpected matched raised";
+                 return 10;
+               })
```
benh, line 115: We need to make this be a bug, you shouldn't be able to try and catch an error that will never propagate to you!

eventuals/catch.h
```diff
@@ -148,12 +151,30 @@ struct _Catch final {
...
+      // 'Error' type so to keep the compiler from trying to compile that
+      // code path we need to add the following 'if constexpr'.
+      if constexpr (!std::disjunction_v<
+                        tuple_types_contains_subtype<
+                            std::decay_t<Error>,
+                            Catches>,
+                        std::conditional_t<
+                            has_all_,
+                            std::true_type,
+                            std::false_type>>) {
```
benh, line 170:
```suggestion
                        std::integral_constant<bool, has_all_>) {
```

eventuals/catch.h
```diff
@@ -148,12 +151,30 @@ struct _Catch final {
           },
           catch_handlers_);
 
-      if (!handled) {
-        if (interrupt_ != nullptr) {
-          k_.Register(*interrupt_);
+      using Catches =
+          std::tuple<typename CatchHandlers_::Error...>;
```
benh, line 155:
```suggestion
      using Catches_ = tuple_types_subtract_t<
          std::tuple<typename CatchHandlers_::Error...>,
          std::tuple<std::monostate>>;
```

eventuals/catch.h
```diff
@@ -188,22 +209,24 @@ struct _Catch final {
...
             // If 'std::exception' is caught
             // then all exceptions will be caught.
-            tuple_types_contains<std::exception, CatchErrors_>,
-            // A 'std::exception_ptr' implies we have a '.all(...)'
-            // and all exceptions are caught.
-            tuple_types_contains<std::exception_ptr, CatchErrors_>>,
+            tuple_types_contains<std::exception, Catches_>,
+            // If we have a '.all(...)' handler then
+            // all exceptions are caught.
+            std::conditional_t<has_all_, std::true_type, std::false_type>>,
```
benh, line 227:
```suggestion
            std::integral_constant<bool, has_all_>>,
```

eventuals/concurrent.h
```diff
@@ -270,10 +277,15 @@ struct _Concurrent final {
     // NOTE: this could also be moved to a .cc file but see the
     // comments above 'CreateOrReuseFiber()' above for what's
     // preventing us from doing so.
-    [[nodiscard]] auto FiberEpilogue(TypeErasedFiber* fiber) {
+    template <typename... Errors>
+    [[nodiscard]] auto FiberEpilogue(
+        TypeErasedFiber* fiber,
+        stout::borrowed_ref<
+            std::optional<std::variant<Errors...>>>&& stopped_or_error) {
```
benh, line 284:
```suggestion
            std::optional<std::variant<Stopped, Errors...>>>&& stopped_or_error) {
```

eventuals/concurrent.h
```diff
@@ -213,10 +218,14 @@ struct _Concurrent final {
     // NOTE: this could also be moved to a .cc file but see the
     // comments above 'CreateOrReuseFiber()' above for what's
     // preventing us from doing so.
-    [[nodiscard]] auto IngressEpilogue() {
+    template <typename... Errors>
+    [[nodiscard]] auto IngressEpilogue(
+        stout::borrowed_ref<
+            std::optional<std::variant<Errors...>>>&& stopped_or_error) {
```
benh, line 224:
```suggestion
            std::optional<std::variant<Stopped, Errors...>>>&& stopped_or_error) {
```

eventuals/concurrent.h
```diff
@@ -158,51 +158,56 @@ struct _Concurrent final {
...
-                  fiber->Reuse();
-                  break;
-                } else if (!fiber->next) {
-                  fiber->next.reset(CreateFiber());
-                  fiber = fiber->next.get();
-                  break;
+    template <typename... Errors>
+    [[nodiscard]] auto CreateOrReuseFiber(
+        stout::borrowed_ref<
+            std::optional<std::variant<Errors...>>>&& stopped_or_error) {
```
benh, line 164:
```suggestion
            std::optional<std::variant<Stopped, Errors...>>>&& stopped_or_error) {
```

eventuals/control-loop.h
```diff
@@ -94,7 +86,7 @@ class ControlLoop final
     task_.Start(
         std::string(name_), // Copy.
         [borrow = std::move(borrow)]() mutable { borrow.relinquish(); },
-        [](std::exception_ptr) { LOG(FATAL) << "Unreachable"; },
+        []() { LOG(FATAL) << "Unreachable"; },
```
benh, line 89: HOW COOL IS THAT! IF NO ERRORS ARE PROPAGATED THAN WE HAVE AN EMPTY ARGUMENT LIST! SOOOOOO COOOOOL! 💪

eventuals/do-all.h
```diff
@@ -93,64 +106,65 @@ struct _DoAll final {
...
-                         k_.Stop();
-                       } catch (...) {
-                         k_.Fail(std::current_exception());
-                       }
+                     std::optional<StoppedOrError> stopped_or_error =
+                         GetStoppedOrErrorIfExists();
+
+                     if (stopped_or_error) {
+                       std::visit(
+                           [this](auto&& error) {
```
benh, line 114:
```suggestion
                           [this](auto&& stopped_or_error) {
```
And this is consistent with what you had in a previous file I took a look at too!

eventuals/do-all.h
```diff
@@ -93,64 +106,65 @@ struct _DoAll final {
...
-                       k_.Stop();
-                     } catch (...) {
-                       k_.Fail(std::current_exception());
-                     }
+                     std::optional<StoppedOrError> stopped_or_error =
+                         GetStoppedOrErrorIfExists();
+
+                     CHECK(stopped_or_error);
+                     std::visit(
+                         [this](auto&& error) {
```
benh, line 137:
```suggestion
                         [this](auto&& stopped_or_error) {
```

eventuals/do-all.h
```diff
@@ -93,64 +106,65 @@ struct _DoAll final {
...
```
benh, line 160:
```suggestion
                         [this](auto&& stopped_or_error) {
```

eventuals/finally.h
```diff
@@ -10,31 +10,50 @@ namespace eventuals {
...
 ////////////////////////////////////////////////////////////////////////
 
+template <typename...>
+struct VariantOfStoppedAndErrors;
+
+template <typename... Errors>
+struct VariantOfStoppedAndErrors<std::tuple<Errors...>> {
+  using type = std::variant<Stopped, Errors...>;
+};
+
```
benh, line 20: Dead code now?
```suggestion
```

eventuals/generator.h
```diff
@@ -16,15 +18,32 @@ namespace eventuals {
...
-        Callback<void(TypeErasedStream&)>* begin,
-        Callback<void(std::exception_ptr)>* fail,
-        Callback<void()>* stop,
-        Callback<function_type_t<void, To_>>* body,
-        Callback<void()>* ended)
+        BeginCallback_* begin,
+        FailCallback_* fail,
+        StopCallback_* stop,
+        BodyCallback_* body,
+        EndedCallback_* ended)
```
benh, line 46: So clean, nice!

eventuals/generator.h
```diff
@@ -187,7 +219,14 @@ struct _Generator final {
           std::unique_ptr<void, Callback<void(void*)>>&,
           Interrupt&,
           Callback<void(TypeErasedStream&)>&&,
```
benh, line 221:
```suggestion
          BeginCallback_&&,
```

eventuals/generator.h
```diff
@@ -187,7 +219,14 @@ struct _Generator final {
...
-          Callback<void(std::exception_ptr)>&&,
+          Callback<
+              function_type_t<
+                  void,
+                  get_rvalue_type_or_void_t<
+                      typename VariantErrorsHelper<
+                          variant_of_type_and_tuple_t<
+                              std::monostate,
+                              Raises>>::type>>>&&,
```
benh, line 229:
```suggestion
          FailCallback_&&,
```

eventuals/generator.h
```diff
@@ -187,7 +219,14 @@ struct _Generator final {
...
           Callback<void()>&&,
```
benh, line 230:
```suggestion
          StopCallback_&&,
```

eventuals/generator.h
```diff
@@ -187,7 +219,14 @@ struct _Generator final {
...
           Callback<function_type_t<void, To>>&&,
```
benh, line 231:
```suggestion
          BodyCallback_&&,
```

eventuals/generator.h
```diff
@@ -187,7 +219,14 @@ struct _Generator final {
...
           Callback<void()>&&)>;
```
benh, line 232:
```suggestion
          EndedCallback_&&)>;
```

eventuals/generator.h
```diff
@@ -16,15 +18,32 @@ namespace eventuals {
...
+  using FailCallback_ = Callback<function_type_t<
+      void,
+      get_rvalue_type_or_void_t<
+          typename VariantErrorsHelper<
+              variant_of_type_and_tuple_t<
+                  std::monostate,
+                  Raises_>>::type>>>;
+  using StopCallback_ = Callback<void()>;
+  using BodyCallback_ = Callback<function_type_t<void, To_>>;
+  using EndedCallback_ = Callback<void()>;
```
benh, line 38: Let's move these to the top level and call them `GeneratorBeginCallback` (without the trailing underscore) and make them templated if necessary (e.g.,  `GeneratorFailCallback` will take a `Raises` template parameter) and then we can use them throughout this file without having to create them 3 places!

benh (changes requested):

eventuals/concurrent.h
```diff
@@ -435,18 +445,26 @@ struct _Concurrent final {
...
-
-    // Indicates whether or not we've got a failure and we should stop
-    // requesting the next upstream value.
-    std::optional<std::exception_ptr> exception_;
   };
 
   // 'Adaptor' is our typeful adaptor that the concurrent continuation
   // uses in order to implement the semantics of 'Concurrent()'.
-  template <typename F_, typename Arg_>
+  template <typename F_, typename Arg_, typename Errors_>
```
benh, line 452:
```suggestion
  template <typename F_, typename Arg_, typename UpstreamErrorsAndErrorsFromE_>
```

eventuals/concurrent.h
```diff
@@ -435,18 +445,26 @@ struct _Concurrent final {
...
   struct Adaptor final : TypeErasedAdaptor {
-    Adaptor(F_ f)
-      : f_(std::move(f)) {}
+    using E_ = typename std::invoke_result_t<F_>;
+
+    using UpstreamErrorsAndErrorsFromE_ = tuple_types_union_t<
+        typename E_::template ErrorsFrom<Arg_, std::tuple<>>,
+        Errors_>;
+
```
benh, line 459:
```suggestion
```

eventuals/concurrent.h
```diff
@@ -710,13 +747,26 @@ struct _Concurrent final {
...
+    // Indicates whether the continuation was stopped or we received
+    // a failure and we should stop requesting the next upstream value.
+    stout::Borrowable<
+        std::optional<
+            variant_of_type_and_tuple_t<Stopped, UpstreamErrorsAndErrorsFromE_>>>
+        stopped_or_error_;
+
+    Adaptor<F_, Arg_, Errors_> adaptor_;
```
benh, line 763:
```suggestion
    Adaptor<F_, Arg_, UpstreamErrorsAndErrorsFromE_> adaptor_;
```

eventuals/loop.h
```diff
@@ -398,10 +398,10 @@ struct _Loop final {
           std::move(stop_));
     }
 
-    template <typename Error = std::exception, typename... Errors>
+    template <typename... Errors>
     auto raises() && {
```
benh, line 402: For consistency let's add the version of `raises<std::tuple<Error1, Error2, ...>>()` here just like we have with `Eventual`. And let's also add in `stream.h` too.

eventuals/scheduler.h
```diff
@@ -396,7 +398,7 @@ struct Reschedulable final {
 
   using Continuation_ =
       decltype(std::declval<_Reschedule::Composable>()
-                   .template k<Arg_>(std::declval<K_>()));
+                   .template k<Arg_, std::tuple<>>(std::declval<K_>()));
```
benh, line 401: This does not match the `continuation.emplace(...)` above.
```suggestion
                   .template k<Arg_, Errors_>(std::declval<K_>()));
```

eventuals/task.h
```diff
@@ -19,13 +20,42 @@ namespace eventuals {
...
+////////////////////////////////////////////////////////////////////////
+
+template <
+    typename E_,
+    typename From_,
+    typename To_,
+    typename Catches_,
+    typename Raises_>
 struct HeapTask final {
+  using StartCallback_ = Callback<function_type_t<void, To_>>;
```
benh, line 43: Just like in `generator.h`, let's move these to the top-level so that we an reuse them multiple places!

eventuals/generator.h
```diff
@@ -436,16 +520,26 @@ struct _Generator final {
...
-                std::move(ended));
-            break;
+          case Action::Fail: {
+            // If 'Catches_' is empty then we will never dispatch
+            // with an action of 'Action::Fail' but since 'action' is a runtime
+            // value the compiler will assume that it's possible that we can
+            // call 'e->Fail()', with what ever type we have for 'error' so to
+            // keep the compiler from trying to compile that code path we need
+            // to add the following 'if constexpr'.
+            if constexpr (std::tuple_size_v<Catches_> != 0) {
```
benh, line 530: I don't understand why we need both this `if constexpr` and the one in `HeapGenerator::Fail`. Seems like we should only need one of them!

eventuals/task.h
```diff
@@ -101,14 +127,22 @@ struct HeapTask final {
     // 'Register()' more than once is well-defined.
     adapted_.Register(interrupt);
 
-    adapted_.Fail(std::move(exception));
+    std::visit(
+        [this](auto&& error) {
+          if constexpr (!std::is_same_v<
+                            std::decay_t<decltype(error)>,
+                            std::monostate>) {
```
benh, line 134: As called out in `generator.h`, why do we need both this `if constexpr` and the one in the dispatch callback? Seems like we should only need one!

eventuals/task.h
```diff
@@ -241,12 +292,14 @@ struct _TaskFromToWith final {
...
     void Fail(Error&& error) {
-      std::exception_ptr exception;
-
-      exception = make_exception_ptr_or_forward(
-          std::forward<Error>(error));
-
-      Dispatch(Action::Fail, std::nullopt, std::move(exception));
+      // NOTE: we only propagate an upstream error into our type-erased
+      // eventual if the eventual catches the error, otherwise we skip it
+      // all together.
```
benh, line 297: Let's also copy this comment into `generator.h` and `transformer.h`!

eventuals/task.h
```diff
@@ -261,29 +314,40 @@ struct _TaskFromToWith final {
...
 
       std::apply(
-          [&](auto&... args) {
+          [&](auto&&... args) {
             std::get<1>(value_or_dispatch_)(
                 action,
-                std::move(exception),
-                args...,
+                std::move(error),
+                std::forward<decltype(args)>(args)...,
```
benh, line 329: Please update `generator.h` and `transformer.h` with this change too! Consistency!

eventuals/task.h
```diff
@@ -261,29 +314,40 @@ struct _TaskFromToWith final {
...
                 std::forward<decltype(from)>(from),
                 e_,
                 *interrupt_,
                 [this](auto&&... args) {
                   k_.Start(std::forward<decltype(args)>(args)...);
                 },
-                [this](std::exception_ptr error) {
-                  k_.Fail(std::move(error));
+                [this](auto&&... void_or_errors) {
+                  if constexpr (sizeof...(void_or_errors)) {
```
benh, line 337: The previous code always got a `std::exception_ptr`, which means WE SHOULD ALWAYS HAVE HAD AN ERROR!
```suggestion
                [this](auto&& errors) {
```
If there is another place in the code that we need to add a `constexpr` to handle the `void` case then we should, but we shouldn't expect end users that are passing callbacks to us to ever need to do this. Either there should be no errors and they should be able to pass a `void` lambda and know that the code is UNREACHABLE or they should take a `std::variant<Error1, Error2, ...>` and know that there is the possibility of errors.

eventuals/task.h
```diff
@@ -261,29 +314,40 @@ struct _TaskFromToWith final {
...
+                        std::forward<
+                            decltype(void_or_errors)>(void_or_errors)...);
+                  }
                 },
                 [this]() {
                   k_.Stop();
                 });
           },
-          args_);
+          std::move(args_));
```
benh, line 350: Do we need to do this in `generator.h` and `transformer.h` too?

eventuals/task.h
benh: How did we get it committed before!? It doesn't look like it to me!

eventuals/task.h
benh: It doesn't look like it to me!

eventuals/task.h
```diff
@@ -494,27 +610,54 @@ class _Task final {
...
+  struct tuple_first_error_type;
+
+  template <typename... Errors>
+  struct tuple_first_error_type<std::tuple<Errors...>> {
+    using type = std::tuple<>;
+  };
+  template <typename Error>
+  struct tuple_first_error_type<std::tuple<Error>> {
+    using type = Error;
+  };
```
benh, line 660: I believe you want to pass _all_ catch errors, not just the first one!

benh (changes requested):

test/finally.cc
```diff
@@ -21,14 +21,20 @@ using testing::ThrowsMessage;
 TEST(Finally, Succeed) {
   auto e = []() {
     return Just(42)
-        >> Finally([](expected<int, std::exception_ptr>&& expected) {
+        >> Finally([](expected<int, std::variant<Stopped>>&& expected) {
```
benh, line 24:
```suggestion
        >> Finally([](expected<int, Stopped>&& expected) {
```

test/generator.cc
```diff
@@ -499,7 +495,7 @@ TEST(Generator, FromTo) {
                    });
                  })
               >> Closure([&]() {
-                   return Iterate(std::move(data));
+                   return Iterate(data);
```
benh, line 498: Why this change?

test/task.cc
```diff
@@ -657,8 +777,8 @@ TEST(Task, RaisesWith) {
           EXPECT_EQ(s, "hello world");
           return Eventual<int>()
               .raises<std::runtime_error>()
-              .start([i = i](auto& k) {
-                return k.Start(i);
+              .start([i = std::move(i)](auto& k) {
+                return k.Start(std::move(i));
```
benh, line 781: Why was this necessary?

benh (changes requested):

eventuals/generator.h
onelxj: We need two of them: one is in Fail to be sure we invoke dispatch with Catch error, second is inside Dispatch (prevents compiler to go in fail path with a wrong error)

test/task.cc
```diff
@@ -168,11 +221,11 @@ TEST(Task, FailOnCallback) {
       ThrowsMessage<std::runtime_error>(StrEq("error from start")));
 }
 
-TEST(Task, FailTerminated) {
+TEST(Task, FailTerminatedPropagate) {
   MockFunction<void()> fail;
 
   EXPECT_CALL(fail, Call())
-      .Times(1);
+      .Times(0);
```
benh, line 228: Let's add `FAIL()` inside the eventual instead so that we are more explicit. I know we aren't doing that everywhere but I want to start doing that in our tests rather than counting using `.Times()`.

test/task.cc
```diff
@@ -204,6 +257,45 @@ TEST(Task, FailTerminated) {
...
+      .Times(1);
+
+  auto e = [&]() -> Task::Of<int>::Raises<
+                     std::runtime_error>::Catches<std::runtime_error> {
+    return [&]() {
+      return Raise("error")
+          >> Eventual<int>()
+                 .raises<std::runtime_error>()
+                 .start([](auto& k) {
+                   k.Fail(std::runtime_error("error from start"));
```
benh, line 276: Let's make this a `FAIL()`.

test/task.cc
```diff
@@ -299,31 +391,35 @@ TEST(Task, StartFuture) {
 }
 
 TEST(Task, FailContinuation) {
-  auto e = []() -> Task::Of<int> {
-    return [x = 42]() {
-      return Just(x);
+  auto e = []() -> Task::Of<int>::Catches<std::runtime_error> {
+    return []() {
+      return Finally([](
+                         expected<
```
benh, line 397:
```suggestion
      return Finally([](expected<
```

test/task.cc
```diff
@@ -204,6 +257,45 @@ TEST(Task, FailTerminated) {
...
+}
+
+TEST(Task, FailTerminatedCatch) {
+  MockFunction<void()> fail;
+
+  EXPECT_CALL(fail, Call())
+      .Times(1);
+
+  auto e = [&]() -> Task::Of<int>::Raises<
+                     std::runtime_error>::Catches<std::runtime_error> {
```
benh, line 270: Doesn't seem like we are exercising `Catches` here? Is this test redundant with any others once we drop `Catches<std::runtime_error>`?

benh (approved):
Woo hoo!!! Congrats!!! So exciting! Well done!

eventuals/transformer.h
```diff
@@ -185,41 +231,61 @@ struct _Transformer final {
...
+                     "compile this lambda as a 'Callback' with some "
+                     "arguments even if 'Raises_' is empty";
+            }
           },
           [this]() {
             k_.Stop();
           },
+          [this](To_ arg) {
+            k_.Body(std::move(arg));
+          },
```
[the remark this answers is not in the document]
onelxj: It is not logic change, but to be consistent with Generator, it could be better to have the same params order in the same containers, isn't it?

PR #451, opened 2022-07-11
Remove unused `include`s

CodingCanuck:
Prefer to avoid including `eventuals/eventual.h` when not directly using
the symbols it provides (`Eventual()` and `_Eventual`).

Add missing headers that were previously transitively included to fix
builds.


ArthurBandaryk (approved):
Thanks Alex!

CodingCanuck:
Mergequeue looks broken since it's spuriously warning about rebasing this PR onto main:
Screen Shot 2022-07-11 at 5 40 11 PM
 
I filed feedback on Mergequeue, and I'll manually merge this for now.

PR #461, opened 2022-07-14
Extracting a separate library for 'backward' stuff

ArthurBandaryk:
Inspired on [`this`](https://github.com/3rdparty/eventuals/pull/460) PR!

### DIFF
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -38,6 +38,25 @@ cc_library(
     ],
 )
 
+cc_library(
+    name = "backward",
+    srcs = ["backward-stacktrace.cc"],
+    hdrs = ["backward-stacktrace.h"],
+    copts = copts(),
+    # By default this target is visible from current package
+    # (test/BUILD.bazel). We need to make it visible for
+    # 'test/concurrent' and 'test/grpc' packages too.
+    visibility = [
+        "//test/concurrent:__pkg__",
+        "//test/grpc:__pkg__",
+    ],
+    deps = [
+        "@com_github_backward_cpp//:backward",
+        "@com_github_google_googletest//:gtest",
+    ],
+    alwayslink = True,
+)
+
 cc_test(
     name = "eventuals",
     srcs = [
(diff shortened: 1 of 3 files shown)


test/BUILD.bazel
```diff
@@ -38,6 +38,28 @@ cc_library(
     ],
 )
 
+cc_library(
+    name = "backward",
+    alwayslink = True,
+    srcs = ["backward-stacktrace.cc"],
+    hdrs = ["backward-stacktrace.h"],
+    copts = copts(),
+    visibility = ["//visibility:public"],
```
CodingCanuck, line 47: Does this need to be public? If so, document why.

test/BUILD.bazel
```diff
@@ -38,6 +38,28 @@ cc_library(
...
+    deps = [
+        "@com_github_backward_cpp//:backward",
+        "@com_github_google_googletest//:gtest",
+    ],
+)
+
+eventuals_deps = [
```
CodingCanuck, line 54: I'd inline this rather than have a variable for it.

I think of a shared deps list like this as a bit of an anti-pattern, since it can make it easy to have unnecessary dependencies in the shared list that are hard to spot or remove. From https://bazel.build/rules/tutorial-sharing-variables:

> Code duplication in BUILD files is usually fine. This can make the file more readable: each declaration can be read and understood without any context. This is important, not only for humans, but also for external tools. For example, a tool might be able to read and update BUILD files to add missing dependencies. Code refactoring and code reuse might prevent this kind of automated modification.

test/concurrent/BUILD.bazel
```diff
@@ -4,6 +4,13 @@
 load("@rules_cc//cc:defs.bzl", "cc_library")
 load("//bazel:copts.bzl", "copts")
 
+concurrent_deps = [
```
CodingCanuck, line 7: ditto, I'd inline this

test/concurrent/fail.cc
```diff
@@ -20,6 +20,7 @@ using testing::ThrowsMessage;
 
 // Tests when at least one of the eventuals fails.
 TYPED_TEST(ConcurrentTypedTest, Fail) {
+  CHECK(false);
```
CodingCanuck, line 23: Revert please :)

test/grpc/BUILD.bazel
```diff
@@ -6,17 +6,27 @@ cc_library(
     hdrs = ["death-constants.h"],
 )
 
+death_deps = [
```
CodingCanuck, line 9: ditto, inline this (and allow the duplication)

CodingCanuck (approved):

PR #462, opened 2022-07-15
Remove 'backward' cc_library target from test/eventuals

ArthurBandaryk:
Since 'backward-stacktrace.cc|.h' are completely unrelated to eventuals,
so it makes sense to keep them separate. This PR removes this files. Moreover, [#2](https://github.com/3rdparty/bazel-rules-backward-cpp/pull/2) moves this files to `3rdparty/bazel-rules-backward-cpp` repo. Thus it would be easy to use
that in '3rdparty/eventuals' and 'reboot-dev/respect' repos.

### DIFF
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -38,25 +38,6 @@ cc_library(
     ],
 )
 
-cc_library(
-    name = "backward",
-    srcs = ["backward-stacktrace.cc"],
-    hdrs = ["backward-stacktrace.h"],
-    copts = copts(),
-    # By default this target is visible from current package
-    # (test/BUILD.bazel). We need to make it visible for
-    # 'test/concurrent' and 'test/grpc' packages too.
-    visibility = [
-        "//test/concurrent:__pkg__",
-        "//test/grpc:__pkg__",
-    ],
-    deps = [
-        "@com_github_backward_cpp//:backward",
-        "@com_github_google_googletest//:gtest",
-    ],
-    alwayslink = True,
-)
-
 cc_test(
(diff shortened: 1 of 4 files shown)


FolMing:
If we are using it in multiple repos, how about moving it over to bazel-rules?

CodingCanuck:
> If we are using it in multiple repos, how about moving it over to bazel-rules?

I like this suggestion: looking through https://github.com/3rdparty/eventuals/blob/main/test/backward-stacktrace.h and https://github.com/3rdparty/eventuals/blob/main/test/backward-stacktrace.cc they appear to have no dependencies on the rest of eventuals, so this doesn't seem to need to live in the eventuals repo.

ArthurBandaryk:
> If we are using it in multiple repos, how about moving it over to bazel-rules?

I think this would be too much redundant. By the way 'promisify-for-test' has the same strategy as mine

CodingCanuck (changes requested):
Can you please explore Nik's suggestion of providing this in the `backward`-specific repo, rather than in `eventuals`? (That feels attractive because the code you're exposing isn't specific to eventuals)

FolMing:
BTW, the compilation time is super nice now thanks to cache 😋

test/BUILD.bazel
```diff
@@ -38,25 +38,6 @@ cc_library(
     ],
 )
 
-cc_library(
```
CodingCanuck, line 41: Please update the PR description: this PR no longer changes `backward`'s visibility

CodingCanuck (approved):

PR #465, opened 2022-07-17
Use several clang-tidy suggestions

FolMing:
So, I was able to run clang-tidy on our codebase.
Some of the changes can be a bit controversial and maybe will require a bit of a talk between us.


ArthurBandaryk (approved):

eventuals/http.h
```diff
@@ -958,7 +958,9 @@ struct _HTTP final {
 
               // Transform 'Request' headers to curl's linked list.
               for (const auto& [key, value] : request_.headers()) {
-                std::string header = key + ": " + value;
+                std::string header = key;
```
CodingCanuck, line 961: I'm not in love with this switch to repeated calls to `append`: It seems nicer to instead switch to https://github.com/fmtlib/fmt

CodingCanuck (approved):
Overall, I like this a lot! I love losing the unneeded `move`s on primitive types.

eventuals/http.h
FolMing: I'm adding a comment for now, but I agree, we should probably use fmt library

benh (approved):
LGTM!

Issue #470, opened 2022-07-18
`//eventuals/grpc` breaks `cc_binary` targets built with `linkshared = True`


CodingCanuck:
Given a minimal `sample-binary.cc` that defines an empty main function:
```c++
int main() {
  return 0;
}
```

These two binaries build:
```bzl

cc_binary(
    name = "sample-eventuals-grpc-binary-default-link",
    srcs = ["sample-binary.cc"],
    copts = copts(),
    # Use default linking.
    deps = ["//eventuals/grpc"],
)

cc_binary(
    name = "sample-eventuals-grpc-binary-static-link",
    srcs = ["sample-binary.cc"],
    copts = copts(),
    # Use static linking (build libraries as `.a` files).
    linkstatic = True,
    deps = ["//eventuals/grpc"],
)
```

But this binary does not:
```bzl
cc_binary(
    name = "sample-eventuals-grpc-binary-shared-link",
    srcs = ["sample-binary.cc"],
    copts = copts(),
    # Use dynamic linking (build libraries as `.so` files).
    linkshared = True,
    deps = ["//eventuals/grpc"],
)
```

The error is reproducible without `//eventuals/grpc` by depending on both `@com_github_grpc_grpc//:grpc++` `@com_github_jemalloc_jemalloc//:jemalloc` together:

```bzl
cc_binary(
    name = "sample-grpc-and-jemalloc-binary-shared-link",
    srcs = ["sample-binary.cc"],
    copts = copts(),
    linkshared = True,
    deps = [
        "@com_github_grpc_grpc//:grpc++",
        "@com_github_jemalloc_jemalloc//:jemalloc",
    ],
)
```

The error is about invalid relocations given the compiler's flags:
```
 /workspaces/respect/submodules/eventuals/test/BUILD.bazel:118:10: Linking test/libsample-grpc-and-eventuals-binary-shared-link.so failed: (Exit 1): clang failed: error executing command 
  (cd /home/vscode/.cache/dazel/eventuals/sandbox/processwrapper-sandbox/63/execroot/com_github_3rdparty_eventuals && \
  exec env - \
    PATH=/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
    PWD=/proc/self/cwd \
  /usr/lib/llvm-14/bin/clang @bazel-out/k8-fastbuild/bin/test/libsample-grpc-and-eventuals-binary-shared-link.so-2.params)
[...]

PR #480, opened 2022-07-19
Prefer to use `malloc` to override malloc

CodingCanuck:
This is recommended here: https://github.com/bazelbuild/bazel/issues/818#issuecomment-180362316

But more importantly, this fixes issues when trying to include eventuals
in targets that are built as shared libraries.

Fixes https://github.com/3rdparty/eventuals/issues/470.

### DIFF
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -1,5 +1,6 @@
 load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
 load("//bazel:copts.bzl", "copts")
+load("//bazel:malloc.bzl", "malloc")
 
 cc_library(
     name = "generate-test-task-name",
@@ -82,6 +83,8 @@ cc_test(
         "unpack.cc",
     ],
     copts = copts(),
+    # Use a faster implementation of malloc (and show that tests pass with it).
+    malloc = malloc(),
     deps = [
         ":generate-test-task-name",
         ":http-mock-server",
--- a/test/grpc/BUILD.bazel
+++ b/test/grpc/BUILD.bazel
@@ -1,5 +1,6 @@
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
 load("//bazel:copts.bzl", "copts")
+load("//bazel:malloc.bzl", "malloc")
 
 cc_library(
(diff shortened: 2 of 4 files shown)


test/BUILD.bazel
```diff
@@ -82,6 +82,8 @@ cc_test(
         "unpack.cc",
     ],
     copts = copts(),
+    # Use a faster implementation of malloc (and show that tests pass with it).
+    malloc = "@com_github_jemalloc_jemalloc//:jemalloc",
```
FolMing, line 86:
```suggestion
    select({
        # NOTE!
        # For now we build jemalloc on macOS and Linux.
        # The way how to build jemalloc on Windows is more
        # complicated. In future we are going to make this
        # lib build on Windows too.
        # https://github.com/3rdparty/bazel-rules-jemalloc/tree/ArthurBandaryk.jemalloc-windows
        "@bazel_tools//src/conditions:windows": [],
        "//conditions:default": [
            "@com_github_jemalloc_jemalloc//:jemalloc",
        ],
```

test/grpc/BUILD.bazel
```diff
@@ -12,6 +12,8 @@ cc_binary(
         "death-client.cc",
     ],
     copts = copts(),
+    # Use a faster implementation of malloc (and show that tests pass with it).
+    malloc = "@com_github_jemalloc_jemalloc//:jemalloc",
```
FolMing, line 16: I'm ok with not using `select()` in `test/grpc/BUILD.bazel` since we don't build these rules on Windows anyway.

FolMing (approved):

test/BUILD.bazel
CodingCanuck: This won't work as-is since `malloc` needs to be a single entry, not a list.

My first attempts at getting `malloc` working with a `select` statement failed, but I'll give it another go.

FolMing:
Let's add the following code to our targets in `test/grpc/`, so that Bazel will not try to compile `jemalloc` while building `...` target:
```
target_compatible_with = select({
        "@platforms//os:linux": [],
        "@platforms//os:macos": [],
        "//conditions:default": ["@platforms//:incompatible"],
    }),
```

Issue #477, opened 2022-07-19
`Collect` into `google::protobuf::RepeatedPtrField`


rjhuijsman:
Imagine I have a gRPC method like this:

```proto
rpc Foo(FooRequest) returns (FooResponse);

message FooResponse {
  repeated Bar bars = 1;
}
```

When writing the body of the Foo method I want to do something like:
```cpp
return Iterate(some_collection)
| Map([](const Item& item) {
    Bar bar;
    bar.set_some_field(item.some_value());
    return bar;
  })
| Collect<google::protobuf::RepeatedPtrField<Bar>>()  // Not currently possible!
| Then([](const google::protobuf::RepeatedPtrField<Bar>& bars) {
    FooResponse response;
    response.mutable_bars()->Swap(&bars);  // Efficient! No copies!
    return response;
  }
```

I can't currently `Collect` into a `RepeatedPtrField` (nor a `RepeatedField`), which means I can't do the efficient `repeated_field->Swap(other_repeated_field)`.

Feature request: allow me to `Collect` into `google::protobuf::RepeatedPtrField` (and less importantly, also `google::protobuf::RepeatedField`).

FolMing:
I think this should be implemented as a template overload in `eventuals/grpc/collect.h`. This way we will not break our Windows build (since for now it doesn't support gRPC).

rjhuijsman:
FWIW, `RepeatedField` and `RepeatedPtrField` are part of the protobuf library, they don't require gRPC.

FolMing:
Added in https://github.com/3rdparty/eventuals/pull/484.

PR #484, opened 2022-07-20
Add support for google::protobuf::Repeated(Ptr)Field in Collect()

FolMing:
Addresses #477.

### DIFF
--- a/eventuals/collect.h
+++ b/eventuals/collect.h
@@ -5,6 +5,9 @@
 #include "eventuals/loop.h"
 #include "eventuals/type-traits.h"
 
+// NOTE: The following include is a rather large one.
+#include "google/protobuf/repeated_field.h"
+
 ////////////////////////////////////////////////////////////////////////
 
 namespace eventuals {
@@ -16,7 +19,23 @@ template <typename Container>
   return Loop<Container>()
       .context(Container())
       .body([](auto& data, auto& stream, auto&& value) {
-        if constexpr (HasEmplaceBack<Container>::value) {
+        if constexpr (
+            std::is_same_v<
+                Container,
+                google::protobuf::RepeatedField<
+                    typename Container::value_type>>) {
+          data.Add(value);
+        } else if constexpr (
+            std::is_same_v<
+                Container,
+                google::protobuf::RepeatedPtrField<
+                    typename Container::value_type>>) {
+          // Should probably use std::forward instead of std::move,
+          // but using std::forward results in a compiler error,
+          // because google::protobuf::RepeatedPtrField::Add
+          // expects an rvalue reference.
+          data.Add(std::move<decltype(value)>(value));
+        } else if constexpr (HasEmplaceBack<Container>::value) {
           data.emplace_back(std::forward<decltype(value)>(value));
         } else {
           data.insert(data.cend(), std::forward<decltype(value)>(value));
--- a/test/collect.cc
+++ b/test/collect.cc
@@ -19,9 +19,9 @@ TEST(Collect, CommonVectorPass) {
 
   std::vector<int> result = *s();
 
+  ASSERT_EQ(2, result.size());
   EXPECT_EQ(5, result.at(0));
   EXPECT_EQ(12, result.at(1));
-  EXPECT_EQ(2, result.size());
 }
 
 TEST(Collect, CommonSetPass) {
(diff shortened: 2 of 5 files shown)


eventuals/collect.h
```diff
@@ -4,6 +4,7 @@
 
 #include "eventuals/loop.h"
 #include "eventuals/type-traits.h"
+#include "google/protobuf/repeated_field.h"
```
CodingCanuck, line 7: I'm confused: how does this `#include` succeed? I don't see any dependencies on something like `com_google_protobuf`.

Ah, I'm guessing we're pulling this in from an indirect dependency on `stout`: https://github.com/3rdparty/stout-flags/blob/fdff8dd353991afc193f77506f04da505237a796/bazel/repos.bzl

Can you please add a direct dependency on a relevant proto target to this file's build rule so the source of the dependency is more clear?

eventuals/collect.h
```diff
@@ -16,7 +17,23 @@ template <typename Container>
   return Loop<Container>()
       .context(Container())
       .body([](auto& data, auto& stream, auto&& value) {
-        if constexpr (HasEmplaceBack<Container>::value) {
+        if constexpr (
+            std::is_same_v<
+                Container,
+                google::protobuf::RepeatedField<
+                    typename Container::value_type>>) {
+          data.Add(value);
```
CodingCanuck, line 25: std::move(value)

test/collect.cc
```diff
@@ -39,5 +39,35 @@ TEST(Collect, CommonSetPass) {
...
+  std::vector<std::string> v = {"Hello", "World"};
+
+  auto s = [&]() {
+    return Iterate(v)
+        | Collect<google::protobuf::RepeatedPtrField<std::string>>();
+  };
+
+  google::protobuf::RepeatedPtrField<std::string> result = *s();
+
+  EXPECT_EQ(result.size(), 2);
```
CodingCanuck, line 52: Prefer `ASSERT` over `EXPECT` when checking a container's size before dereferencing elements, or doing other things that could crash if previous `EXPECT` statements fail.

> Use ASSERT when it doesn't make sense to continue

https://testing.googleblog.com/2008/07/tott-expect-vs-assert.html

test/collect.cc
```diff
@@ -39,5 +39,35 @@ TEST(Collect, CommonSetPass) {
...
+  EXPECT_STREQ("Hello", result.begin()->c_str());
```
CodingCanuck, line 53: Whenever you're passing `foo->c_str()` to `EXPECT_STREQ`, use `EXPECT_EQ` instead:

`EXPECT_EQ("Hello", *result.begin());`

Here and below.

See discussion: https://github.com/3rdparty/stout/pull/53#discussion_r903157994

eventuals/collect.h
CodingCanuck, line 7: This feels like a pretty heavy dependency for users that aren't using protos.

Is there any way we can avoid it, e.g. by splitting out the proto parts into a separate file?

I have no idea how to do this, so if I was writing this PR I'd have this same `#include`, even though it feels messy :/

Maybe add a TODO to figure out a cleaner way to separate out proto overrides from proto-agnostic stuff?

CodingCanuck (approved):

eventuals/collect.h
FolMing: Add's argument is `const &`, so no need to move

CodingCanuck: TIL!

eventuals/collect.h
FolMing: Done, please review the files again and resolve the conversation if everything is ok!

FolMing: I'm not sure if we want a TODO here. It might be better to make an issue for that.

CodingCanuck: Looks excellent, thanks!

CodingCanuck: why not both

I think of TODOs kind of like hyperlinks to issues, plus they give the reader more context (like normal contexts). Personally, I add them liberally :)

Up to you, though.

CodingCanuck (approved):

rjhuijsman (approved):
Awesome! Thanks! 🎉

eventuals/collect.h
FolMing: Added a simple note. If we encounter any problems with compilation times while including `eventuals/collect.h` we'll at least see "HERE'S YOUR PROBLEM" thingy :D

PR #485, opened 2022-07-21
Add an option for users to overload an eventuals::Collector for Collect()

FolMing:
Makes `Collect()` more generic and allows others to extend its functionality to other collections by implementing their own `eventuals::Collector`.

### DIFF
--- a/eventuals/collect.h
+++ b/eventuals/collect.h
@@ -5,45 +5,67 @@
 #include "eventuals/loop.h"
 #include "eventuals/type-traits.h"
 
-// NOTE: The following include is a rather large one.
-#include "google/protobuf/repeated_field.h"
-
 ////////////////////////////////////////////////////////////////////////
 
 namespace eventuals {
 
 ////////////////////////////////////////////////////////////////////////
 
-template <typename Container>
+template <typename Collection, typename = void>
+struct Collector {
+  template <typename T>
+  static void Collect(Collection& collection, T&& value) {
+    // This static_assert will always fail
+    // if the compiler needs to compile this implementation.
+    static_assert(
+        always_false_v<T>,
+        "Provide your own EventualsCollector");
+  }
+};
+
+////////////////////////////////////////////////////////////////////////
+
+// Collectors for STL collections.
+
+template <typename Collection>
+struct Collector<
+    Collection,
+    std::enable_if_t<
+        eventuals::HasEmplaceBack<Collection>::value>> {
+  template <typename T>
+  static void Collect(Collection& collection, T&& value) {
+    static_assert(std::is_convertible_v<T, typename Collection::value_type>);
+    collection.emplace_back(std::forward<T>(value));
+  }
+};
+
+template <typename Collection>
+struct Collector<
+    Collection,
+    std::enable_if_t<
+        eventuals::HasInsert<Collection>::value>> {
+  template <typename T>
+  static void Collect(Collection& collection, T&& value) {
+    static_assert(std::is_convertible_v<T, typename Collection::value_type>);
+    collection.insert(std::forward<T>(value));
+  }
+};
+
+////////////////////////////////////////////////////////////////////////
+
+// TODO(folming): Issue #486.
+template <typename Collection>
 [[nodiscard]] auto Collect() {
-  return Loop<Container>()
-      .context(Container())
-      .body([](auto& data, auto& stream, auto&& value) {
-        if constexpr (
-            std::is_same_v<
-                Container,
-                google::protobuf::RepeatedField<
-                    typename Container::value_type>>) {
-          data.Add(value);
-        } else if constexpr (
-            std::is_same_v<
-                Container,
-                google::protobuf::RepeatedPtrField<
-                    typename Container::value_type>>) {
-          // Should probably use std::forward instead of std::move,
-          // but using std::forward results in a compiler error,
-          // because google::protobuf::RepeatedPtrField::Add
-          // expects an rvalue reference.
-          data.Add(std::move<decltype(value)>(value));
-        } else if constexpr (HasEmplaceBack<Container>::value) {
-          data.emplace_back(std::forward<decltype(value)>(value));
-        } else {
-          data.insert(data.cend(), std::forward<decltype(value)>(value));
-        }
+  return Loop<Collection>()
+      .context(Collection())
+      .body([](auto& collection, auto& stream, auto&& value) {
+        Collector<Collection>::Collect(
+            collection,
+            std::forward<decltype(value)>(value));
         stream.Next();
       })
-      .ended([](auto& data, auto& k) {
-        k.Start(std::move(data));
+      .ended([](auto& collection, auto& k) {
+        k.Start(std::move(collection));
       });
 }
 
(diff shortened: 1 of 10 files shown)


eventuals/collect.h
```diff
@@ -5,50 +5,62 @@
...
 #include "eventuals/type-traits.h"
 
-// NOTE: The following include is a rather large one.
-#include "google/protobuf/repeated_field.h"
-
 ////////////////////////////////////////////////////////////////////////
 
 namespace eventuals {
 
 ////////////////////////////////////////////////////////////////////////
```
benh, line 13: I like this approach!

What if we provided an extension point called `EventualsCollect` instead that let us not have to provide the same "insert function" for all non-STL types.

For example, we could have (very much pseudo code, I have not compiled this):

```cpp
template <typename T, typename U, std::enable_if_t<std::is_convertible_v<U, T>>>
void EventualsCollect(google::protobuf::RepeatedField<T>& repeated, U&& u) {
  repeated.Add(std::forward<U>(u));
}
```

Then when we get a value we want to insert we can just do:

`EventualsCollect(data, std::forward<decltype(value)>(value));`

Then we can implement an `EventualsCollect` for STL things like so:

```cpp
template <typename Collection, typename T, std::enable_if_t<HasEmplaceBack<Collection>::value>>
void EventualsCollect(Collection& collection, T&& t) {
  collection.emplace_back(std::forward<T>(t));
}

template <typename Collection, typename T, std::enable_if_t<HasInsert<Collection>::value>>
void EventualsCollect(Collection& collection, T&& t) {
  collection.insert(std::forward<T>(t));
}
```

benh (changes requested):

CodingCanuck:
I'll defer to Ben for this review.

rjhuijsman:
I too will defer to Ben. 👍🏻

eventuals/collect.h
```diff
@@ -5,45 +5,70 @@
...
 
-template <typename Container>
+template <typename Collection, typename Enable = void>
+struct EventualsCollector {
+  template <typename T>
+  void Collect(Collection& collection, T&& value) {
+    static_assert(
+        !std::is_same_v<void, Enable>,
```
benh, line 19: Hmm, I'm not sure if this is doing what you want? I love the idea of a static time check here though, and I've been meaning to add an `always_false_v<T>` for a while. Let's add it now!

In `eventuals/type-traits.h` lets add:

```cpp

template <typename>
inline constexpr bool always_false_v = false;

``` 

Then here we can do:
```suggestion
    static_assert(
        always_false_v<T>,
```

Which should only get triggered if the compiler needs to compile this implementation!

eventuals/collect.h
```diff
@@ -5,45 +5,70 @@
...
+template <typename Collection>
+struct EventualsCollector<
+    Collection,
+    std::enable_if_t<
+        eventuals::HasEmplaceBack<Collection>::value>> {
+  template <typename T>
+  std::enable_if_t<
+      std::is_convertible_v<
+          T,
+          typename Collection::value_type>>
```
benh, line 37: I love this checking this constraint, but are we trying to "select" vis SFINAE here or just providing better error messages? It seems like the latter, as in, we've already selected via `EventualsCollector<C>`, and now we just want better error messages, so let's use a `static_assert()` in the body instead! Here and below please!

eventuals/collect.h
```diff
@@ -5,45 +5,70 @@
...
-          data.Add(std::move<decltype(value)>(value));
-        } else if constexpr (HasEmplaceBack<Container>::value) {
-          data.emplace_back(std::forward<decltype(value)>(value));
-        } else {
-          data.insert(data.cend(), std::forward<decltype(value)>(value));
-        }
+  return Loop<Collection>()
+      .context(Collection())
+      .body([](auto& collection, auto& stream, auto&& value) {
+        EventualsCollector<Collection>().Collect(
```
benh, line 65: Let's make `Collect()` be a static function and then we can do:
```suggestion
        EventualsCollector<Collection>::Collect(
```

test/collect.cc
```diff
@@ -37,8 +82,14 @@ TEST(Collect, CommonSetPass) {
   ASSERT_EQ(2, result.size());
   EXPECT_EQ(5, *result.begin());
   EXPECT_EQ(12, *++result.begin());
+
+  // The initial set should remain unchanged.
+  ASSERT_EQ(2, v.size());
+  EXPECT_EQ(5, *v.begin());
+  EXPECT_EQ(12, *++v.begin());
```
benh, line 89: Let's use an `EXPECT_THAT(v,  ElementsAre(5, 12))` or it's variants here and below please!

test/collect.cc
```diff
@@ -1,15 +1,54 @@
...
 
 #include <set>
+#include <utility>
 #include <vector>
 
 #include "eventuals/iterate.h"
+#include "google/protobuf/repeated_field.h"
 #include "gtest/gtest.h"
 
+template <typename Collection>
```
benh, line 11: Great to have these! Let's put them in a `protobuf` header and library that we don't add to the default `eventuals` library so others can rely on them via `@eventuals//protobuf` if they want this functionality rather than having to reimplement this again.

test/collect.cc
```diff
@@ -1,15 +1,54 @@
...
+        std::is_same_v<
+            Collection,
+            google::protobuf::RepeatedPtrField<
+                typename Collection::value_type>>>> {
+  // Since 'Add(T&& value)' takes ownership of value, we pass a copy.
+  template <typename T>
+  std::enable_if_t<
+      std::is_convertible_v<
+          T,
+          typename Collection::value_type>>
```
benh, line 24: `static_assert` for these ones too!

benh (approved):
Thanks for this, Nik!

Please let me know if you have any questions about my requests!

eventuals/collect.h
FolMing: Yeah, you're right x)

Let me fix that..

eventuals/collect.h
FolMing: Hmm, I've played around with my implementation and it seems like it's working, but I do like your solution instead! I'll try it

FolMing:
@benh, I know you've already approved, but I would like you to have a quick glance one last time, thank you :)

eventuals/protobuf/collectors.h
```diff
@@ -0,0 +1,47 @@
...
+    Collection,
+    std::enable_if_t<
+        std::is_same_v<
+            Collection,
+            google::protobuf::RepeatedField<
+                typename Collection::value_type>>>> {
+  template <typename T>
+  static void Collect(Collection& collection, T&& value) {
+    static_assert(std::is_convertible_v<T, typename Collection::value_type>);
+    collection.Add(value);
```
benh, line 39:
```suggestion
    collection.Add(std::forward<T>(value));
```
Or leave a comment, better yet a static assert, that we don't need to forward because it's a primitive type. Seems like it doesn't hurt to `std::forward`!

benh (approved):
Looks good!

eventuals/protobuf/collectors.h
FolMing: Should we use `std::forward` here? `Add` expects an `lvalue reference`:
```
template <typename Element>
inline void RepeatedField<Element>::Add(const Element& value)
```

PR #488, opened 2022-07-21
Update example condition in README

stephaniepalocz:
The intended condition seems to need a `!`; not important to the content, but I won't be able to unsee it 🙃


sp33drac3r (approved):
👀 🙈

PR #493, opened 2022-07-22
Refactoring operator '*'

ArthurBandaryk:
Move 'Promisify()' and 'operator*()' in a separate
'eventuals/promisify.h' file plus minor refactoring.


test/dns-resolver.cc
```diff
@@ -29,15 +29,9 @@ TEST_F(DomainNameResolveTest, Succeed) {
...
-  auto [future, k] = PromisifyForTest(std::move(e));
-
-  k.Start();
-
-  EventLoop::Default().RunUntil(future);
-
   EXPECT_TRUE(
       std::regex_match(
-          future.get(),
+          *e,
```
benh, line 34:
```suggestion
          *std::move(e),
```

Here and everywhere else where we were doing a `PromisifyForTest(std::move(e))` please!

test/filesystem.cc
```diff
@@ -27,8 +27,9 @@ TEST_F(FilesystemTest, OpenAndCloseFileSucceed) {
 
   EXPECT_TRUE(std::filesystem::exists(path));
 
-  auto e = OpenFile(path, UV_FS_O_RDONLY, 0)
-      | Then([&path](auto&& file) {
+  auto e = [&path]() {
```
benh, line 30: Awesome! Love these cleanups!

test/http.cc
```diff
@@ -50,12 +50,7 @@ TEST_P(HttpTest, Get) {
           typename decltype(e)::template ErrorsFrom<void, std::tuple<>>,
           std::tuple<std::runtime_error>>);
 
-  auto [future, k] = PromisifyForTest(std::move(e));
-  k.Start();
-
-  EventLoop::Default().RunUntil(future);
-
-  auto response = future.get();
+  auto response = *e;
```
benh, line 53: Another good example of where we need to `std::move(e)` here and below or better yet wrap in a lambda like you did for the filesystem tests!

benh (approved):

Issue #486, opened 2022-07-21
Remove the need for writing an underlying `Container` type in `Collect()`


FolMing:
Right now we are writing code using `Collect()` this way:
```
std::vector<int> v = /* whatever */;

auto e = Iterate(v)
    | Collect<std::vector<int>>();
```

The fact is, we don't need to write the underlying type for `std::vector` twice, since we can get it from the previous function.
With this change the code would look like this:
```
std::vector<int> v = /* whatever */;

auto e = Iterate(v)
    | Collect<std::vector>();
```

FolMing:
Implemented in #495

PR #495, opened 2022-07-26
Erase Collect's collection type

FolMing:
Addresses https://github.com/3rdparty/eventuals/issues/486.

There's one specific test in `test/flat-map.cc` that worked before because we put the collection into the context of the `Loop`, that's why we could access the collection inside the `.stop` function. With the current implementation this doesn't work (though it's possible to make it work, but it would go against what we currently have in other `Eventuals` (No `Stop` method passes anything into the `k.Stop()`)). What are your thoughts on this test, @benh?

For now, I commented out this test.

### DIFF
--- a/eventuals/collect.h
+++ b/eventuals/collect.h
@@ -53,7 +53,9 @@ struct Collector<
 
 ////////////////////////////////////////////////////////////////////////
 
-// TODO(folming): Issue #486.
+// Used when Collection is a completely defined type, e.g.:
+// Collect<std::vector<int>>
+
 template <typename Collection>
 [[nodiscard]] auto Collect() {
   return Loop<Collection>()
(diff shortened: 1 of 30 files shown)


eventuals/collect.h
```diff
@@ -53,20 +53,99 @@ struct EventualsCollector<
...
+    }
+  };
+};
+
+////////////////////////////////////////////////////////////////////////
+
+template <template <typename...> class Collection>
+[[nodiscard]] auto Collect() {
+  return _Collect::Composable<Collection>();
+}
```
benh, line 142: What if we made the "typed" version just return what we always had:

```cpp
    return Loop<Collection>()
      .context(Collection())
      .body([](auto& collection, auto& stream, auto&& value) {
        EventualsCollector<Collection>::Collect(
            collection,
            std::forward<decltype(value)>(value));
        stream.Next();
      })
      .ended([](auto& collection, auto& k) {
        k.Start(std::move(collection));
      });
```

Then we made this version be defined _after_ and just do:

```suggestion
struct _Collect {
  template <template <typename...> class Collection_>
  struct Composable {
    template <typename Arg>
    using ValueFrom = Collection_<std::decay_t<Arg>>;

    template <typename Arg, typename Errors>
    using ErrorsFrom = Errors;

    template <typename Arg, typename K>
    auto k(K k) && {
      return Build(Collect<Collection_<std::decay_t<Arg>>>(), std::move(k));
    }  
  };
};

template <template <typename...> class Collection>
[[nodiscard]] auto Collect() {
  return _Collect::Composable<Collection>{};
}
```

As we "grow" up as a library we strive to only use our existing abstractions! When we need to create a "new' abstraction it always begs to ask if our existing ones are not powerful enough. If you see a reason why this won't work let me know and I'll review `_Collect::Continuation`!

benh: And if you can think of an abstraction that lets us do this all without ever creating the `_Collect` struct that would be even better!

benh (approved):

test/flat-map.cc
```diff
@@ -197,9 +197,9 @@ TEST_F(FlatMapTest, Interrupt) {
                  | Just(x);
            })
         | FlatMap([](int x) { return Iterate({1, 2}); })
-        | Collect<std::vector<int>>()
-              .stop([](auto& container, auto& k) {
-                k.Start(std::move(container));
+        | Collect<std::vector>()
+              .stop([](auto& collection, auto& k) {
+                k.Start(std::move(collection));
```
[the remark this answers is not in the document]
FolMing: For now, I'm using the old Collect version with a fully-specialized Collection type, but we should definitely resolve this inconsistency in the future, I'm putting this into the backlog.

PR #502, opened 2022-07-28
Use Interrupt::Handler in all callback for Loop and Stream

onelxj:


test/concurrent/emit-fail-interrupt.cc
```diff
@@ -30,13 +30,13 @@ TYPED_TEST(ConcurrentTypedTest, EmitFailInterrupt) {
   auto e = [&]() {
     return Stream<int>()
                .interruptible()
-               .begin([](auto& k, Interrupt::Handler& handler) {
-                 handler.Install([&k]() {
+               .begin([](auto& k, auto& handler) {
+                 handler->Install([&k]() {
```
benh, line 34:
```suggestion
                 CHECK(handler) << "Test expects interrupt to be registered";
                 handler->Install([&k]() {
```

Here and everywhere else!

benh (approved):

PR #506, opened 2022-08-02
Introduce 'RequestResponseChannel'

ArthurBandaryk:
This PR adds 'RequestResponseChannel' which is generally
speaking a synchronized "pipe in a pipe" for handling
requests and responses. So it guarantees that for every
request that is written we read a response in order.

### DIFF
--- a/test/request-response-channel.cc
+++ b/test/request-response-channel.cc
@@ -1,0 +1,108 @@
+#include "eventuals/request-response-channel.h"
+
+#include <deque>
+#include <optional>
+#include <tuple>
+
+#include "eventuals/do-all.h"
+#include "eventuals/loop.h"
+#include "eventuals/map.h"
+#include "eventuals/take.h"
+#include "eventuals/then.h"
+#include "gtest/gtest.h"
+#include "test/promisify-for-test.h"
+
+using eventuals::DoAll;
+using eventuals::Loop;
+using eventuals::Map;
+using eventuals::RequestResponseChannel;
+using eventuals::TakeFirst;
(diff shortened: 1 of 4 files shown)


test/request-response-channel.cc
```diff
@@ -0,0 +1,153 @@
...
+};
+
+TEST(RequestResponseChannel, SingleRequestSingleResponse) {
+  RequestResponseChannel<Request, Response> channel;
+
+  auto write = [&]() {
+    return channel.Request(Request{"request1"})
+        >> Then([](std::optional<Response>&& response) {
+             EXPECT_EQ(response->data, "response1");
+           });
```
benh, line 36:
```suggestion
    return channel.Request(Request{"request1"});
```

benh (changes requested):

test/request-response-channel.cc
```diff
@@ -0,0 +1,121 @@
...
+  *read();
+
+  std::optional<Response> response1 = future1.get();
+  std::optional<Response> response2 = future2.get();
+
+  ASSERT_TRUE(response1);
+  ASSERT_TRUE(response2);
+
+  EXPECT_EQ(response1.value(), Response{"response for request1"});
+  EXPECT_EQ(response2.value(), Response{"response for request2"});
```
benh, line 120: This is fine, but we can remove the extra `operator==` above and just do:
```suggestion
  EXPECT_EQ(response1->data, "response for request1");
  EXPECT_EQ(response2->data, "response for request2");
```

benh (approved):

PR #511, opened 2022-08-03
Use `= default` where applicable. Add missing include in `eventuals/grpc/logging.h` 

FolMing:
According to [clang-tidy](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-equals-default.html):
```
The explicitly defaulted function declarations enable more opportunities in optimization,
because the compiler might treat explicitly defaulted functions as trivial.
```


CodingCanuck (approved):
Love it!

ArthurBandaryk (approved):

PR #515, opened 2022-08-03
Explicit definition of all ctrs, operator= and dstrs

NikitaNikolaenko:
To comply with the rule C.21 of C++ Core Guidelines:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all

This *should* have the same behavior as our implicitly-generated version.

### DIFF
--- a/eventuals/builder.h
+++ b/eventuals/builder.h
@@ -18,6 +18,14 @@ class Field;
 template <typename Value_>
 class Field<Value_, false> {
  public:
+  Field() = default;
+
+  Field(const Field&) = default;
+  Field(Field&&) noexcept = default;
+
+  Field& operator=(const Field&) = default;
+  Field& operator=(Field&&) noexcept = default;
+
   virtual ~Field() = default;
 
   template <typename... Args>
@@ -32,6 +40,12 @@ class Field<Value_, true> {
  public:
   Field(Value_ value)
     : value_(std::move(value)) {}
+
+  Field(const Field&) = default;
+  Field(Field&&) noexcept = default;
+
+  Field& operator=(const Field&) = default;
+  Field& operator=(Field&&) noexcept = default;
 
   virtual ~Field() = default;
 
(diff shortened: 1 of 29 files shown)


eventuals/builder.h
```diff
@@ -18,6 +18,14 @@ class Field;
 template <typename Value_>
 class Field<Value_, false> {
  public:
+  Field() = default;
+
+  Field(const Field&) = default;
```
CodingCanuck, line 23: Why are some copy / move operations `noexcept` and others are not? I'd expect `noexcept` use to be consistent here: please document.

Glancing at code snippets below, I'd expect us to use `noexcept` everywhere we can.

CodingCanuck (commented):
I'm going to remove myself from the reviewer list for this PR: the mechanics look good while skimming through the diffs, but IMHO this could use @benh 's eyes on what we intend to be copy/moveable vs. which constructors and assignment operators we want to `delete`.

CodingCanuck:
> I'm going to remove myself from the reviewer list for this PR: the mechanics look good while skimming through the diffs, but IMHO this could use @benh 's eyes on what we intend to be copy/moveable vs. which constructors and assignment operators we want to `delete`.

Ooh, I don't know how to remove myself now that I've "left a review" :/ https://github.com/orgs/community/discussions/23054

If this is useful, here's a doc on how to dismiss PR reviews: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review (this might be a note for myself if a re-review request comes up)

FolMing:
>IMHO this could use @benh 's eyes on what we intend to be copy/moveable vs. which constructors and assignment operators we want to delete

This is one of the reasons why this PR was created :D
We have quite a lot of implicitness in our code right now (which I personally don't like) and this PR makes everyone think about whether we need this or that.

eventuals/builder.h
FolMing: I don't think we have `noexcept` on copy operations (unless I made a typo?).

https://clang.llvm.org/extra/clang-tidy/checks/performance/noexcept-move-constructor.html

CodingCanuck: > I don't think we have `noexcept` on copy operations (unless I made a typo?).

Why don't we have `noexcept` on our copy operations? Can our copy operations throw exceptions in a way that our move operation can't?

If not, it seems like our copy operations should also be `noexcept`.

NikitaNikolaenko:
Closing an old PR

PR #517, opened 2022-08-03
Introduce 'ControlLoop'

ArthurBandaryk:
This PR adds a new abstraction called 'ControlLoop'. It is
basically just a piece of code that is for ever running and
doing some kind of computation for us.

### DIFF
--- a/eventuals/control-loop.h
+++ b/eventuals/control-loop.h
@@ -1,0 +1,131 @@
+#pragma once
+
+#include "eventuals/callback.h"
+#include "eventuals/catch.h"
+#include "eventuals/finally.h"
+#include "eventuals/lock.h"
+#include "eventuals/promisify.h"
+#include "eventuals/task.h"
+#include "eventuals/type-check.h"
+#include "stout/borrowable.h"
+
+////////////////////////////////////////////////////////////////////////
+
+namespace eventuals {
+
+////////////////////////////////////////////////////////////////////////
+
+// Helper for running, interrupting, and waiting on a "control loop".
+//
+// Your control loop MUST not raise any errors or return a value to
+// force you to properly handle your errors and store any values
+// that you may generate.
+//
+// A control loop is always automatically started on construction
+// and you can interrupt it by calling 'Interrupt()' and wait for it
+// to complete by calling 'Wait()'.
+class ControlLoop final
+  : public Synchronizable,
+    public stout::enable_borrowable_from_this<ControlLoop> {
+ public:
+  template <typename F>
+  ControlLoop(std::string&& name, F f)
+    : Synchronizable(),
+      name_(name),
+      wait_until_finished_(&lock()),
+      task_(this, [f = std::move(f)](ControlLoop* loop) mutable {
+        static_assert(
+            std::is_invocable_v<F>,
+            "'ControlLoop' expects a callable (e.g., a lambda) "
+            "that takes no arguments");
+
+        static_assert(
+            !HasValueFrom<F>::value,
+            "'ControlLoop' expects a callable not an eventual");
+
+        using E = decltype(f());
+
+        static_assert(
+            std::is_void_v<E> || HasValueFrom<E>::value,
+            "'ControlLoop' expects a callable (e.g., a lambda) that "
+            "returns an eventual but you're returning a value");
+
+        static_assert(
+            !std::is_void_v<E>,
+            "'ControlLoop' expects a callable (e.g., a lambda) that "
+            "returns an eventual but you're not returning anything");
+
+        using Value = typename E::template ValueFrom<void>;
+
+        static_assert(
+            std::is_void_v<Value>,
+            "'ControlLoop' eventual should return 'void'");
+
+        using Errors = typename E::template ErrorsFrom<void, std::tuple<>>;
+
+        static_assert(
+            std::is_same_v<Errors, std::tuple<>>,
+            "'ControlLoop' eventual should not raise any errors");
+
+        return std::move(f)()
+            >> loop->Synchronized(
+                Finally([loop](expected<void, std::exception_ptr>&& expected) {
+                  if (!expected) {
+                    try {
+                      std::rethrow_exception(expected.error());
+                    } catch (StoppedException& e) {
+                    } catch (std::exception& e) {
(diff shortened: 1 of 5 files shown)


eventuals/control-loop.h
```diff
@@ -0,0 +1,118 @@
...
+        Synchronized(
+            wait_until_finished_.Wait(Borrow([this]() {
+              return /* wait = */ !finished_.load();
+            }))));
+  }
+
+ private:
+  std::string name_;
+  ConditionVariable wait_until_finished_;
+  std::atomic<bool> finished_ = false;
```
benh, line 110: We can make this a normal `bool` now.

eventuals/control-loop.h
```diff
@@ -0,0 +1,118 @@
...
+        std::string(name_), // Copy.
+        []() {},
+        [](std::exception_ptr) { LOG(FATAL) << "Unreachable"; },
+        []() { LOG(FATAL) << "Unreachable"; });
+  }
+
+  ~ControlLoop() override {
+    CHECK(finished_.load())
+        << "'ControlLoop' must have finished before destruction!";
+
```
benh, line 88:
```suggestion
```

eventuals/control-loop.h
```diff
@@ -0,0 +1,118 @@
...
+            >> loop->Synchronized(
+                Then([loop]() {
+                  loop->finished_.store(true);
+                  loop->wait_until_finished_.Notify();
+                }));
+      }) {
+    // Start the task!
+    task_.Start(
+        std::string(name_), // Copy.
+        []() {},
```
benh, line 80:
```suggestion
    // Start the task!
    //
    // NOTE: we borrow ourselves so that we can ensure 
    // that the task has really finished before destructing.
    stout::borrowed_ptr<ControlLoop> borrow = Borrow();
    task_.Start(
        std::string(name_), // Copy.
        [borrow = std::move(borrow)]() { borrow.relinquish(); },
```

test/control-loop.cc
```diff
@@ -0,0 +1,50 @@
...
+TEST(ControlLoop, SimplePipeHandling) {
+  Pipe<std::string> pipe;
+
+  ControlLoop control_loop("Simple pipe writing", [&]() {
+    return Repeat([&, i = 0]() mutable {
+             return i++;
+           })
+        >> Until([&](auto& i) {
+             return i == 5;
+           })
```
benh, line 31: Let's simplify and use `Iterate({1, 2, 3, 4, 5})` for all of this, thanks!

eventuals/control-loop.h
```diff
@@ -0,0 +1,118 @@
...
+#include "eventuals/type-check.h"
+#include "stout/borrowable.h"
+
+////////////////////////////////////////////////////////////////////////
+
+namespace eventuals {
+
+////////////////////////////////////////////////////////////////////////
+
+// Helper for running, interrupting, and waiting on a "control loop".
```
benh, line 19:
```suggestion
// Helper for running, interrupting, and waiting on a "control loop".
//
// Your control loop MUST not raise any errors or return a value to
// force you to properly handle your errors and store any values
// that you may generate. 
//
// A control loop is always automatically started on construction
// and you can interrupt it by calling 'Interrupt()' and wait for it
// to complete by calling 'Wait()'.
```

test/control-loop.cc
```diff
@@ -0,0 +1,50 @@
...
+using eventuals::ControlLoop;
+using eventuals::Loop;
+using eventuals::Map;
+using eventuals::Pipe;
+using eventuals::Repeat;
+using eventuals::Until;
+
+using testing::ElementsAre;
+
+TEST(ControlLoop, SimplePipeHandling) {
```
benh, line 22: Please add a test that interrupts the control loop and shows that you can still wait for it and it will complete! You can use an `Eventual`, see tests/eventual.cc for some examples.

benh (changes requested):

eventuals/control-loop.h
```diff
@@ -0,0 +1,127 @@
...
+            // just in case.
+            >> Catch()
+                   .raised<std::exception>([](std::exception&& e) {
+                     LOG(FATAL) << e.what();
+                   })
+            >> loop->Synchronized(
+                Then([loop]() {
+                  loop->finished_ = true;
+                  loop->wait_until_finished_.Notify();
+                }));
```
benh, line 81:
```suggestion
            >> loop->Synchronized(
                Finally([loop](expected<void, std::exception_ptr>&& expected) {
                  if (!expected) {
                    try {
                      std::rethrow_exception(expected.error());
                    } catch (StoppedException& e) {
                    } catch (std::exception& e) {
                      LOG(WARNING) << "Unreachable: " << e.what();
                    } catch (...) {
                      LOG(WARNING) << "Unreachable (error does not derive from 'std::exception')";
                    }
                  }                  
                  loop->finished_ = true;
                  loop->wait_until_finished_.Notify();
                }));
```

benh (changes requested):

test/control-loop.cc
```diff
@@ -0,0 +1,72 @@
...
+  ControlLoop control_loop("Interrupt eventual", [&]() {
+    return Eventual<void>()
+        .interruptible()
+        .start([&](
+                   auto& k,
+                   std::optional<Interrupt::Handler>& handler) {
+          handler->Install([&k]() {
+            k.Stop();
+          });
+          start.Call();
```
benh, line 66: Let's move `start.Call()` to the top of the function so that we don't race if in the future we have a control loop execute in it's own dedicated thread.
```suggestion
          start.Call();
          handler->Install([&k]() {
            k.Stop();
          });
```

test/control-loop.cc
```diff
@@ -0,0 +1,72 @@
...
+
+  EXPECT_THAT(
+      *collect_from_pipe(),
+      ElementsAre("0", "1", "2", "3", "4"));
+}
+
+TEST(ControlLoop, Interrupt) {
+  MockFunction<void()> start;
+
+  EXPECT_CALL(start, Call()).Times(1);
```
benh, line 55: It's convention to put the `.Times()` or other functions below the `EXPECT_CALL`.
```suggestion
  EXPECT_CALL(start, Call())
    .Times(1);
```

benh (approved):

PR #519, opened 2022-08-04
Avoid unecessary `auto`

CodingCanuck:
When using `auto` makes a type invisible to the reader, it makes code
harder for unfamiliar readers to understand. This is especially
impactful for readers looking at snippets of code to understand how to
use eventuals types, which happens a lot while most eventuals lack any
documentation.

Per https://google.github.io/styleguide/cppguide.html#Type_deduction :

> Use type deduction only if it makes the code clearer to readers who
> aren't familiar with the project, or if it makes the code safer. Do
> not use it merely to avoid the inconvenience of writing an explicit
> type.

### DIFF
--- a/eventuals/builder.h
+++ b/eventuals/builder.h
@@ -35,27 +35,27 @@ class Field<Value_, true> {
 
   virtual ~Field() = default;
 
-  auto& value() & {
+  Value_& value() & {
     return value_;
   }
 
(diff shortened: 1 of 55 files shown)


ArthurBandaryk (approved):
Thanks Alex! Now the code is much more readable!

PR #528, opened 2022-08-07
Use single handler for `Interrupt`

onelxj:


eventuals/interrupt.h
```diff
@@ -50,56 +49,32 @@ class Interrupt final {
...
-        handler->next_,
-        handler,
-        std::memory_order_release,
-        std::memory_order_relaxed));
-
-    return true;
+  [[nodiscard]] bool Install(Handler* handler) {
+    auto* stored = handler_.load();
+    CHECK(stored) << "Interrupt has already being triggered";
+    return handler_.compare_exchange_weak(stored, handler);
```
benh, line 60: We want this to return `false` if the interrupt has been triggered and we are not able to install the handler otherwise `true` after we've installed the handler.

benh (changes requested):

eventuals/interrupt.h
```diff
@@ -50,56 +49,36 @@ class Interrupt final {
...
-      if (handler->next_ == nullptr) {
-        return false;
-      }
-    } while (!head_.compare_exchange_weak(
-        handler->next_,
-        handler,
-        std::memory_order_release,
-        std::memory_order_relaxed));
+  [[nodiscard]] bool Install(Handler* handler) {
+    auto* stored = handler_.load();
```
benh, line 58:
```suggestion
    Handler* stored = handler_.load();
```

eventuals/interrupt.h
```diff
@@ -50,56 +49,36 @@ class Interrupt final {
...
+    if (!stored) {
+      // Means that interrupt has already been triggerred.
+      return false;
+    }
 
-    return true;
+    return handler_.compare_exchange_weak(stored, handler);
```
benh, line 64: I'd like to make this stronger:
```suggestion
    // NOTE: we should be the *only* ones trying to install an interrupt at a time,
    // so the only way we fail to install ourselves is if the interrupt got triggered.
    bool installed = handler_.compare_exchange_weak(
        stored,
        handler,
        std::memory_order_release,
        std::memory_order_relaxed);

  CHECK(installed || stored == nullptr);;

  return installed;
```

eventuals/interrupt.h
```diff
@@ -50,56 +49,36 @@ class Interrupt final {
...
   }
 
   void Trigger() {
     // NOTE: nullptr signifies that the interrupt has been triggered.
-    auto* handler = head_.exchange(nullptr);
+    auto* handler = handler_.exchange(nullptr);
```
benh, line 69:
```suggestion
    Handler* handler = handler_.exchange(nullptr);
```

benh (approved):

PR #532, opened 2022-08-08
Add combinators 'OnBegin' and 'OnEnded'

### LINKED ISSUE #527: Provide a way to run cleanup at the end of stream generation
I don't know how to write "cleanup" code that runs at the end of stream generation. For example (pseudocode):

```
return TypeCheck<Generator<int>>(Iterate({1, 2})
    >> Map([](int value) { return value; })
    >> Then([]() { LOG(INFO) << "All done!; })
);
```

The final `Then` statement changes the type of the resulting eventual chain such that it's no longer able to create a generator from the result.

This came up when writing a gRPC server handler for a bidirectional streaming RPC, where the server handler wanted to run some extra logic after sending a final request to a client.

benh:
Adds combinators for doing computation (possibly asynchronously) when a stream/generator begins or ends.

Fixes #527

### DIFF
--- a/eventuals/on-begin.h
+++ b/eventuals/on-begin.h
@@ -1,0 +1,164 @@
+#pragma once
+
+#include "eventuals/eventual.h"
+#include "eventuals/stream.h"
+#include "eventuals/then.h"
+
+////////////////////////////////////////////////////////////////////////
+
+namespace eventuals {
+
+////////////////////////////////////////////////////////////////////////
+
+// Returns an eventual which will invoke the specified function when a
+// stream/generator begins. Note that the function can return an
+// eventual itself but that eventual must eventually return 'void' and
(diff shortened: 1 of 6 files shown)


ArthurBandaryk (approved):

eventuals/on-begin.h
```diff
@@ -0,0 +1,164 @@
...
+#include "eventuals/stream.h"
+#include "eventuals/then.h"
+
+////////////////////////////////////////////////////////////////////////
+
+namespace eventuals {
+
+////////////////////////////////////////////////////////////////////////
+
+// Returns an eventual which will invoke the specified function when a
```
CodingCanuck, line 13: Drive-by: thanks for including such great documentation for these!

PR #534, opened 2022-08-08
Add `State` to `Interrupt`

onelxj:


eventuals/interrupt.h
```diff
@@ -28,6 +48,21 @@ class Interrupt final {
         callback_(std::move(that.callback_)) {
     }
 
+    ~Handler() {
+      State state = state_.state();
+
+      if (state == INSTALLED && state_.Update(state, UNINSTALLING)) {
+        interrupt().Uninstall(this);
```
benh, line 55:
```suggestion
        if (!interrupt().Uninstall(this)) {
          CHECK(state_.Update(state, INSTALLED));
        }
```

benh (changes requested):

eventuals/interrupt.h
```diff
@@ -49,11 +85,41 @@ class Interrupt final {
...
+            nullptr,
+            std::memory_order_release,
+            std::memory_order_relaxed)) {
+      handler->state_.Wait([](Handler::State state, size_t count) {
+        return /* until = */ state == Handler::State::INSTALLED;
+      });
+
+      Handler::State state = handler->state_.state();
+      CHECK_EQ(state, Handler::State::INSTALLED);
+      CHECK(handler->state_.Update(state, Handler::State::UNINSTALLED));
```
benh, line 116:
```suggestion
      // We've "won" the race to uninstall the handler but we need to 
      // make sure that `~Handler()` isn't simultaneously trying which
      // may end up reading deleted memory if they happen to be
      // context switched out just before they try and use `handler_`
      // themselves. To deal with this case we _wait_ until we can
      // officially do the uninstalling.
      Handler::State state = Handler::State::INSTALLED;
      while (!handler->state_.Update(state, Handler::State::UNINSTALLED) {
        // TODO(benh): pause CPU.
      }
```

eventuals/interrupt.h
```diff
@@ -49,11 +85,41 @@ class Interrupt final {
...
+    } else {
+      handler = handler_.load();
+      CHECK(handler == nullptr || handler == &placeholder_handler_);
```
benh, line 119:
```suggestion
      CHECK(handler == nullptr || handler == &placeholder_handler_)
          << "'Interrupt' is being used while destructing!";
```

eventuals/interrupt.h
```diff
@@ -63,22 +129,59 @@ class Interrupt final {
...
+      while (!stored->state_.Update(state, Handler::State::UNINSTALLED)) {}
+    }
+
     return installed;
   }
 
+  [[nodiscard]] bool Uninstall(Handler* handler) {
+    // If 'handler' is current 'handler_' then the interrupt *must not*
+    // have been triggered so we should set it back to 'placeholder_handler_'
+    // instead of nullptr. If we successfully did the compare-and-swap then
```
benh, line 159:
```suggestion
    // If 'handler' is the current 'handler_' then the interrupt *must not*
    // have been triggered so we want to set 'handler_' back to 'placeholder_handler_'
    // instead of nullptr. If we successfully did the compare-and-swap then that must mean
    // that 'handler' is the current 'handler_' so we "won" the race of uninstalling and
```

eventuals/interrupt.h
```diff
@@ -63,22 +129,59 @@ class Interrupt final {
...
+    // we can update our state to UNINSTALLED.
+    //
+    // Otherwise someone else must have done the compare-and-swap
+    // and they are responsible for setting us to UNINSTALLED.
```
benh, line 163:
```suggestion
    // Otherwise someone else "won" the compare-and-swap race
    // and they are responsible for setting us to UNINSTALLED.
```

eventuals/interrupt.h
```diff
@@ -63,22 +129,59 @@ class Interrupt final {
...
+
   void Trigger() {
     // NOTE: nullptr signifies that the interrupt has been triggered.
     Handler* handler = handler_.exchange(nullptr);
-    if (handler != nullptr) {
+    if (handler != nullptr && handler != &placeholder_handler_) {
+      Handler::State state = handler->state_.state();
+      // To be sure that callback won't be invoked while destructing.
+      CHECK_EQ(state, Handler::State::INSTALLED);
+      CHECK(handler->state_.Update(state, Handler::State::UNINSTALLED));
```
benh, line 184:
```suggestion
      // Need to uninstall the handler so that in the event invoking the callback causes the handler to get destructed it won't try and uninstall itself!
      CHECK_EQ(state, Handler::State::INSTALLED);
      CHECK(handler->state_.Update(state, Handler::State::UNINSTALLED));
```

benh (changes requested):
There are multiple `bool Install(...)` functions in `Handler` that need `[[nodiscard]]`!

eventuals/interrupt.h
```diff
@@ -28,20 +45,66 @@ class Interrupt final {
...
       return interrupt().Install(this);
     }
 
-    bool Install() {
+    [[nodiscard]] bool Install() {
       CHECK(callback_);
       return interrupt().Install(this);
     }
 
+    [[nodiscard]] State InstallOrExecuteIfTriggered(
```
benh, line 83:
```suggestion
    State InstallOrExecuteIfTriggered(
```

test/concurrent/emit-fail-pipe.cc
```diff
@@ -80,3 +66,6 @@ TYPED_TEST(ConcurrentTypedTest, EmitFailInterrupt) {
 
 } // namespace
 } // namespace eventuals::test
+
+
+// Stream -> Pipe
```
benh, line 71:
```suggestion
```

benh (approved):

PR #539, opened 2022-08-10
Introduce 'Notification' abstraction

benh:
This is similar to 'absl::Notification', except with less functions (for now, we can add more later) and built to work with eventuals.

### DIFF
--- a/eventuals/notification.h
+++ b/eventuals/notification.h
@@ -1,0 +1,75 @@
+#pragma once
+
+#include "eventuals/lock.h"
+#include "eventuals/then.h"
+
+////////////////////////////////////////////////////////////////////////
+
+namespace eventuals {
+
+////////////////////////////////////////////////////////////////////////
+
+// This header file defines a 'Notification' abstraction, which allows
+// an eventual to be able to wait for a single occurrence of a single
+// event. Said another way, the 'Notification' object maintains a
+// private boolean "notified" that transitions to 'true' at most once.
+//
+// This is similar to 'absl::Notification'
+// (https://abseil.io/docs/cpp/guides/synchronization#notification).
+//
+// You can call 'Wait()' on a 'Notification' to wait until the
+// "notified" state is 'true' and 'Notify()' to set the notification's
+// "notified" state to 'true' and notify all waiting eventuals that
+// the event has occurred. This method may only be called once and
+// will abort if called more than once.
+//
+// Note that while 'Notify()' should only be called once, it is
+// perfectly valid to call 'Wait()' multiple times and/or from
+// multiple eventuals -- even after the notification's "notified"
+// state has been set -- in which case those calls will not wait.
+//
+// Note that the lifetime of a 'Notification' requires careful
+// consideration; it might not be safe to destroy a 'Notification'
+// after calling 'Notify()' since other eventuals may have called, or
+// will call, 'Wait()'.
+class Notification final : public Synchronizable {
+ public:
+  Notification()
+    : notification_(&lock()) {}
+
+  // Sets "notified" to true and notifies waiting eventuals.
+  //
+  // Should only be called once. Repeated calls result in an abort.
+  //
+  // Returns an eventual 'void'.
+  [[nodiscard]] auto Notify() {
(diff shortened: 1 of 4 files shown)


eventuals/notification.h
```diff
@@ -0,0 +1,68 @@
...
+// private boolean "notified" that transitions to 'true' at most once.
+//
+// This is similar to 'absl::Notification'
+// (https://abseil.io/docs/cpp/guides/synchronization#notification).
+//
+// You can call 'Wait()' on a 'Notification' to wait until the
+// "notified" state is 'true' and 'Notify()' to set the notification's
+// "notified" state to 'true' and notify all waiting eventuals that
+// the event has occurred. This method should only be called once and
+// will do nothing on repeated calls.
```
CodingCanuck, line 24: WDYT about mimicking `absl`'s behavior of disallowing repeated `Notify()` calls, since they're likely due to buggy code?

`absl` does this by crashing in non-opt builds (to make it super quick and easy to notice and find your buggy usage), but not crashing in opt builds (where you might not want bugs to crash your server): https://github.com/abseil/abseil-cpp/blob/149ab40581d33febf3f8bddb52505c9ee9edd838/absl/synchronization/notification.cc#L30-L37

IMHO ~`CHECK` failures on errors (at least optionally / by default) make it much easier to get your code to a bug-free state.

eventuals/notification.h
```diff
@@ -0,0 +1,68 @@
...
+// Note that the lifetime of a 'Notification' requires careful
+// consideration; it might not be safe to destroy a 'Notification'
+// after calling 'Notify()' since other eventuals may have called, or
+// will call, 'Wait()'.
+class Notification final : public Synchronizable {
+ public:
+  Notification()
+    : notification_(&lock()) {}
+
+  ~Notification() override = default;
```
CodingCanuck, line 40: Can we omit this line? (I didn't think you needed to explicitly override a default destructor [or any virtual method] when the one you're overriding is non-pure and has the same visibility as your override)

eventuals/notification.h
```diff
@@ -0,0 +1,68 @@
...
+
+  [[nodiscard]] auto Notify() {
```
CodingCanuck, line 42: Please add comments on these public methods.

We can get good inspiration from https://github.com/abseil/abseil-cpp/blob/149ab40581d33febf3f8bddb52505c9ee9edd838/absl/synchronization/notification.h#L103-L108 :

Here's my attempt:
```
// Sets the notified state of this notification to true and wakes waiting threads.
// Should only be called once. Repeated calls are undefined behavior.
```

eventuals/notification.h
```diff
@@ -0,0 +1,68 @@
...
+    return Synchronized(Then([this]() {
+      if (!notified_) {
+        notified_ = true;
+        notification_.NotifyAll();
+      }
+    }));
+  }
+
+  [[nodiscard]] auto Wait() {
```
CodingCanuck, line 51: In the spirit of more explicit method names, how about absl's `WaitForNotification()`?

CodingCanuck, line 51: Please add comments for non-trivial public API methods.

Here's my take, inspired by `absl`'s: https://github.com/abseil/abseil-cpp/blob/149ab40581d33febf3f8bddb52505c9ee9edd838/absl/synchronization/notification.h#L87

```
// Continues once the notification's "notification" state is true, which might be immediately
// if Notify() has already been called.
```

test/notification.cc
```diff
@@ -0,0 +1,56 @@
...
+
+namespace eventuals::test {
+namespace {
+
+TEST(NotificationTest, NotifyThenWait) {
+  Notification notification;
+
+  *notification.Notify();
+
+  *notification.Wait();
```
CodingCanuck, line 18: Should we chain these two eventuals? (Why do we need two separate operator stars? My inference of eventuals style was to generally prefer to first build long chains, then execute the whole chain in one go)

Maybe add comments here if their separation is important.

CodingCanuck (approved):
Excellent! Thanks :)

PR #543, opened 2022-08-12
Check for insufficient stack space when creating continuations

benh:


eventuals/compose.h
```diff
@@ -95,8 +96,49 @@ struct Composed final {
...
     using Value = typename Left_::template ValueFrom<Arg>;
-    return std::move(left_)
-        .template k<Arg>(std::move(right_).template k<Value>(std::move(k)));
+
+    auto composed = [&]() {
+      return std::move(left_).template k<Arg>(
+          std::move(right_).template k<Value>(std::move(k)));
+    };
+
+#ifndef __APPLE__
```
FolMing, line 105:
```suggestion
#if defined(_WIN32) || defined(__linux__)
```

ArthurBandaryk: Agreed!

ArthurBandaryk: For Windows there is a need to find other api for this stuff.

ArthurBandaryk: As I got `pthread_get_stackaddr_np` and `pthread_get_stacksize_np` functions are only for macOS platform (see this [`doc`](https://opensource.apple.com/source/libpthread/libpthread-105.1.4/src/pthread.c.auto.html)).

FolMing: Ok, now I'm confused, if it's only available on MacOS, why do we have `#ifndef __APPLE__`?

FolMing: Also, we are using `__MACH__` in other files

eventuals/compose.h
```diff
@@ -95,8 +96,49 @@ struct Composed final {
...
+    pthread_t self = pthread_self();
+
+    void* stack_start = pthread_get_stackaddr_np(self);
+    size_t stack_size = pthread_get_stacksize_np(self);
+    void* stack_end = (char*) stack_start - stack_size;
+
+    size_t size = sizeof(decltype(composed()));
+
+    size_t available = ((char*) &size) - ((char*) stack_end);
```
FolMing, line 114: I would prefer this code to be in a separate function in `os.h`

FolMing: It would also be super cool if we had some tests for this function as well

eventuals/os.h
```diff
@@ -58,8 +61,83 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
           &cpuset),
       0);
 }
+
+
 #endif // __MACH__
 
+static inline void CheckSufficientStackSpace(const size_t composed_size) {
```
benh, line 68: Let's add `inline Bytes GetStackSize()` and `inline void* GetStackStart()` and `inline void* GetStackEnd()` and `inline Bytes GetStackAvailable()`.

eventuals/os.h
```diff
@@ -58,8 +61,83 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+
+  CHECK(has_sufficient_stack_space)
+      << "\n"
+      << "\n"
+      << "You've got a large continuation that may exceed the available\n"
+      << "space on the stack!\n"
+      << "\n"
+      << "It looks like your stack size is: "
+      << std::to_string(stack_size * 0.000001) << "Mb\n"
+      << (stack_size < 8000000
```
benh, line 131:
```suggestion
      << (stack_size < Megabytes(8)
```

eventuals/os.h
benh, line 68:
```suggestion
inline void CheckSufficientStackSpace(const size_t composed_size) {
```

PR #545, opened 2022-08-13
Introduce 'TestingCompletionThreadPool'

benh:
NOTE: this PR has multiple smaller commits to make it easier to see the progression of changes.

Often when writing tests with gRPC you want to deterministically ensure some ordering of events to test a particular code path. You can do so by using a 'TestingCompletionThreadPool' which only has a single thread that can be paused so as to ensure that no computation takes place until a particular time or you can wait on it until there is no more work to be done.

### DIFF
--- a/eventuals/grpc/completion-thread-pool.h
+++ b/eventuals/grpc/completion-thread-pool.h
@@ -233,20 +233,7 @@ class TestingCompletionThreadPool {
   stout::borrowed_ref<CompletionThreadPool<::grpc::CompletionQueue>>
   ClientCompletionThreadPool();
 
-  // NOTE: taking a 'std::function' here instead of a 'Callback'
-  // because this is for testing where we don't care about dynamic
-  // memory allocation and it simplifies the tests.
-  void RunUntilCondition(const std::function<bool()>& condition);
-
-  template <typename T>
-  void RunUntil(const std::future<T>& future) {
-    return RunUntilCondition([&future]() {
-      auto status = future.wait_for(std::chrono::nanoseconds::zero());
-      return status == std::future_status::ready;
-    });
-  }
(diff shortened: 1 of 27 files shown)


ArthurBandaryk (approved):

eventuals/grpc/completion-thread-pool.h
```diff
@@ -14,35 +14,66 @@ namespace grpc {
...
+            });
+      }
     }
   }
 
-  ~CompletionThreadPool() {
+  StaticCompletionThreadPool(
+      unsigned int number_of_completion_queues =
+          std::thread::hardware_concurrency(),
+      unsigned int number_of_polling_threads_per_completion_queue = 1);
```
onelxj: Where is an implementation?

eventuals/grpc/server.cc
```diff
@@ -377,20 +408,53 @@ ServerStatusOrServer ServerBuilder::BuildAndStart() {
...
 
-  for (size_t i = 0; i < numberOfCompletionQueues_.value(); ++i) {
-    cqs.push_back(builder_.AddCompletionQueue());
+  cqs.reserve(number_of_completion_queues_.value());
+
+  for (size_t i = 0; i < number_of_completion_queues_.value(); ++i) {
+    cqs.emplace_back(builder_.AddCompletionQueue());
   }
 
+  auto pool = [&]() -> std::variant<
```
onelxj, line 439: Let's add some name for this type since it is used at least twice

onelxj (approved):

eventuals/grpc/server.cc
benh: Great idea! Thank you!

PR #549, opened 2022-08-16
Make Client interruptable

while-false:
With this commit the grpc client can now be interrupted.

### DIFF
--- a/eventuals/grpc/client.h
+++ b/eventuals/grpc/client.h
@@ -367,6 +367,7 @@ class Client {
     };
 
     return Eventual<ClientCall<Request, Response>>()
+        .interruptible()
         .template raises<std::runtime_error>()
         .start(
             [data = Data{
@@ -378,7 +379,24 @@ class Client {
                  ::grpc::TemplatedGenericStub<
                      RequestType,
                      ResponseType>(channel_)},
-             callback = Callback<void(bool)>()](auto& k) mutable {
+             callback = Callback<void(bool)>()]( //
+                auto& k,
+                std::optional<Interrupt::Handler>& handler) mutable {
+              // Install an interrupt handler that will cancel and close the
+              // connection context if/when the client is interrupted.
+              // If we can not install the handler then the interrupt has
+              // already been triggered so we try to cancel.
+              if (handler) {
+                bool installed = handler->Install([&data]() {
+                  data.context->TryCancel();
+                });
+
+                if (!installed) {
+                  // TODO: Should we just k.Stop() here?
+                  data.context->TryCancel();
+                }
+              }
+
               const auto* method =
                   google::protobuf::DescriptorPool::generated_pool()
                       ->FindMethodByName(data.name);
--- a/test/grpc/client-interruptible.cc
+++ b/test/grpc/client-interruptible.cc
@@ -1,0 +1,96 @@
+#include <sys/socket.h>
+#include <sys/un.h>
+
+#include "eventuals/grpc/client.h"
+#include "eventuals/grpc/completion-thread-pool.h"
+#include "eventuals/then.h"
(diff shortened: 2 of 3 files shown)


eventuals/grpc/client.h
```diff
@@ -378,7 +379,22 @@ class Client {
                  ::grpc::TemplatedGenericStub<
                      RequestType,
                      ResponseType>(channel_)},
-             callback = Callback<void(bool)>()](auto& k) mutable {
+             callback = Callback<void(bool)>()](auto& k, std::optional<Interrupt::Handler>& handler) mutable {
```
benh, line 382: You'll need to wrap this line, I'm ok for now with just doing the following if it's under 80 characters:
```suggestion
             callback = Callback<void(bool)>()](auto& k, auto& handler) mutable {
```

eventuals/grpc/client.h
```diff
@@ -378,7 +379,22 @@ class Client {
...
+              // Install an interrupt handler that will cancel and close the
+              // connection context if/when the client is interrupted.
+              // If for some reason we can not install the handler, we'll go
+              // ahead and close the connection.
```
benh, line 386:
```suggestion
              // If we can not install the handler then the interrupt has
              // already been triggered so we try to cancel.
```

test/grpc/client-interruptible.cc
```diff
@@ -0,0 +1,94 @@
...
+using helloworld::HelloReply;
+using helloworld::HelloRequest;
+
+using testing::MockFunction;
+
+using stout::Borrowable;
+
+class ClientInterruptibleTest : public TemporaryDirectoryTest {
+  // Create test fixture that allows us to create a (temporary) domain socket
+  // for use in our test.
```
benh, line 26:
```suggestion

// Create test fixture that allows us to create a (temporary) domain socket
// for use in our test.
class ClientInterruptibleTest : public TemporaryDirectoryTest {
```

test/grpc/client-interruptible.cc
```diff
@@ -0,0 +1,94 @@
...
+  EXPECT_CALL(client_call_success, Call())
+      .Times(0);
+
+  // Low key, create a domain socket and start listening on it.
+  sockaddr_un addr;
+  addr.sun_family = AF_UNIX;
+  strcpy(addr.sun_path, domain_socket_path().c_str());
+
+  int server_fd = socket(AF_UNIX, SOCK_STREAM, 0);
+  ASSERT_EQ(bind(server_fd, (sockaddr*) (&addr), sizeof(addr)), 0);
```
benh, line 50:
```suggestion
  int server_fd = socket(AF_UNIX, SOCK_STREAM, 0);
  ASSERT_GE(server_fd, 0);
  ASSERT_EQ(bind(server_fd, (sockaddr*) (&addr), sizeof(addr)), 0);
```

test/grpc/client-interruptible.cc
```diff
@@ -0,0 +1,94 @@
...
+  ASSERT_EQ(listen(server_fd, 100), 0);
+
+
```
benh, line 53:
```suggestion
```

test/grpc/client-interruptible.cc
```diff
@@ -0,0 +1,94 @@
...
+  // Register the interrupt that we will later trigger.
+  Interrupt interrupt;
+  k.Register(interrupt);
+
+  // Start the job in the background.
+  k.Start();
+
+  // Wait for the grpc client to attempt to connect on the socket and accept
+  // the connection. The client should now be taling grpc to the very quite
```
benh, line 83:
```suggestion
  // the connection. The client should now be talking grpc to the very quiet
```

benh (approved):
So low key.

PR #551, opened 2022-08-17
Introduce 'EventLoop::RunOnce()' and refactor 'operator*' to use

benh:
This commit also introduces a 'Run()' function which 'operator*' now
calls instead (since 'operator*' is just syntactic sugar).


onelxj (approved):

benh:
Yes I'll look into it!

PR #552, opened 2022-08-17
Thread implementation

ArthurBandaryk:
This PR implements 'Thread' class which has the possibility
to run the task in another thread as well as setting stack
size with specific value. For now this stuff works only on
`Linux` and `macOS`. For `Windows` the implementation might
be added in the future.

### DIFF
--- a/eventuals/os.h
+++ b/eventuals/os.h
@@ -192,6 +192,152 @@ inline void CheckSufficientStackSpace(const size_t size) {
       << "doesn't get allocated on the stack!\n"
       << "\n";
 }
+
+////////////////////////////////////////////////////////////////////////
+
+class Thread {
+ public:
+  Thread() = default;
+
+  ~Thread() noexcept {
+    CHECK(!joinable_) << "A thread was left not joined/not detached";
+  }
+
+  Thread(const Thread& other) = delete;
+
+  // IMPORTANT NOTE: on macos stacksize should be a multiple of the system
+  // page size!!!
+  // (check_line_length skip)
+  // https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/pthread_attr_setstacksize.3.html
+  template <typename Callable>
+  Thread(
+      Callable&& callable,
+      const std::string& name,
+      const Bytes& stack_size = Megabytes(8))
+    : joinable_{true} {
+    CHECK_GE(stack_size.bytes(), PTHREAD_STACK_MIN)
+        << "Stack size should not be less than"
+           " the system-defined minimum size";
+
+    pthread_attr_t attr;
+
+    PCHECK(pthread_attr_init(&attr) == 0)
+        << "Failed to initialize thread attributes via "
+           "'pthread_attr_init(...)'";
+
+    // IMPORTANT NOTE: on macos stacksize should be a multiple of the system
+    // page size, otherwise `pthread_attr_setstacksize` will fail.
+    // (check_line_length skip)
+    // https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/pthread_attr_setstacksize.3.html
+    PCHECK(pthread_attr_setstacksize(&attr, stack_size.bytes()) == 0)
+        << "Failed to set the stack size via 'pthread_attr_setstacksize'"
+           " (if you are on macOS - probably you are trying"
+           " to set the stack size which is not a multiple"
+           " of the system page size)";
+
+    struct Data {
+      std::string thread_name{};
+      Callable callable;
+      Data(const std::string& name, Callable&& f)
+        : thread_name{name},
+          callable{std::forward<Callable>(f)} {}
+    };
+
+    PCHECK(pthread_create(
+               &thread_handle_,
+               &attr,
+               +[](void* arg) -> void* {
+                 Data* data = reinterpret_cast<Data*>(arg);
+
+                 PCHECK(
+                     pthread_setname_np(
+#ifdef __linux__
+                         pthread_self(),
+#endif
+                         data->thread_name.c_str())
+                     == 0)
+                     << "Failed to set thread name via "
+                        "'pthread_setname_np(...)'";
+
+                 try {
+                   data->callable();
+                 } catch (const std::exception& e) {
+                   LOG(WARNING) << "Caught exception while running thread '"
+                                << data->thread_name
+                                << "': " << e.what();
+                 } catch (...) {
+                   LOG(WARNING) << "Caught unknown exception while "
+                                   "running thread '"
+                                << data->thread_name
+                                << "'";
+                 }
+                 delete data;
+                 return nullptr;
+               },
+               new Data(name, std::forward<Callable>(callable)))
+           == 0)
+        << "Failed to create a new thread via 'pthread_create'";
+
+    PCHECK(pthread_attr_destroy(&attr) == 0)
+        << "Failed to destroy thread attributes via "
+           "'pthread_attr_destroy(...)'";
+  }
+
+  Thread(Thread&& that) noexcept {
+    *this = std::move(that);
+  }
+
+  Thread& operator=(const Thread& other) = delete;
+
+  Thread& operator=(Thread&& that) noexcept {
+    if (this == &that) {
+      return *this;
+    }
+
+    CHECK(!joinable_) << "Thread not joined nor detached";
+
+    thread_handle_ = std::exchange(that.thread_handle_, pthread_t{});
+    joinable_ = std::exchange(that.joinable_, false);
+
+    return *this;
+  }
+
+  pthread_t native_handle() const {
+    return thread_handle_;
+  }
+
+  bool is_joinable() const {
+    return joinable_;
+  }
+
+  void join() {
+    if (joinable_) {
+      PCHECK(pthread_join(thread_handle_, nullptr) == 0)
+          << "Failed to join thread via 'pthread_join(...)'";
+    }
+    joinable_ = false;
+  }
+
+  void detach() {
+    CHECK(joinable_)
+        << "Trying to detach already joined/detached thread";
+
+    joinable_ = false;
+
+    PCHECK(pthread_detach(thread_handle_) == 0)
+        << "Failed to detach thread via 'pthread_detach(...)'";
+  }
+
+ private:
+  // For Linux `pthread_t` is unsigned long, for
+  // macOS this is an `_opaque_pthread_t` struct.
+  // https://opensource.apple.com/source/xnu/xnu-517.7.7/bsd/sys/types.h
+  pthread_t thread_handle_{};
+  bool joinable_ = false;
+};
+
+////////////////////////////////////////////////////////////////////////
+
 #else // If Windows.
 inline void CheckSufficientStackSpace(const size_t size) {}
 #endif
--- a/test/os.cc
+++ b/test/os.cc
@@ -1,0 +1,85 @@
+#include "eventuals/os.h"
+
+#include <memory>
+
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+namespace eventuals::test {
+namespace {
+
+#ifndef _WIN32
+
+using eventuals::os::Thread;
+using testing::MockFunction;
+
+void foo() {
+  // Do nothing.
+}
+
+TEST(Thread, NotJoinable) {
+  Thread t;
+  EXPECT_FALSE(t.is_joinable());
+}
+
+TEST(Thread, Joinable) {
+  Thread t{&foo, "thread_name"};
+  EXPECT_TRUE(t.is_joinable());
+  t.join();
+  EXPECT_FALSE(t.is_joinable());
+}
+
+TEST(Thread, SetStackSize) {
+  Thread t{
+      []() {
+        EXPECT_EQ(os::GetStackInfo().size, Bytes(16'777'216));
+      },
+      "thread_name",
+      Bytes(16'777'216)};
+  EXPECT_TRUE(t.is_joinable());
+  t.join();
+  EXPECT_FALSE(t.is_joinable());
+}
+
+TEST(Thread, LambdaThatCapturesEverything) {
+  MockFunction<void()> start;
+
+  EXPECT_CALL(start, Call())
+      .Times(1);
+
+  Thread t{[&]() {
+             start.Call();
+           },
+           "thread_name"};
+
+  t.join();
+}
+
+TEST(Thread, LambdaThatCapturesNothing) {
+  Thread t{[]() {}, "thread_name"};
+  t.detach();
+}
+
+TEST(Thread, FunctionPointer) {
+  Thread t1{&foo, "thread_name1"};
+  Thread t2{foo, "thread_name2"};
+
+  t1.join();
+  t2.join();
+}
+
+TEST(Thread, Moveable) {
+  auto done = std::make_unique<bool>(true);
+
+  Thread t{[done = std::move(done)]() {
+             EXPECT_TRUE(*done);
+           },
+           "thread_name"};
+
+  t.join();
+}
+
+#endif
+
+} // namespace
+} // namespace eventuals::test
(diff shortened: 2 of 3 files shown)


eventuals/os.h
```diff
@@ -60,6 +67,153 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+        << "Error in pthread_attr_destroy(...)";
+
+    if (creation_result != 0) {
+      delete state;
+      throw std::runtime_error{
+          "Failed to create a new thread with "
+          "'pthread_create(), error code = '"
+          + std::to_string(creation_result)};
+    }
+  }
```
benh, line 149:
```suggestion
  template <typename Callable>
  Thread(const std::string& name, Callable&& callable, const size_t stack_size_in_bytes = 8'388'608)
    : joinable_{true} {
    CHECK_GE(stack_size_in_bytes, PTHREAD_STACK_MIN)
        << "Stack size should not be less than"
           " the system-defined minimum size";

    pthread_attr_t attr;

    PCHECK(pthread_attr_init(&attr) == 0);
    
     // TODO(artur): use pthread_attr_setname_np

    // IMPORTANT NOTE: on macos stacksize should be a multiple of the system
    // page size, otherwise `pthread_attr_setstacksize` will fail.
    // (check_line_length skip)
    // https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/pthread_attr_setstacksize.3.html
    PCHECK(pthread_attr_setstacksize(&attr, stack_size_in_bytes) == 0)
        << "Failed to set the stack size via 'pthread_attr_setstacksize' (if you are on macOS - probably you are trying"
         << " to set the stack size which is not a multiple"
         << " of the system page size)";

    PCHECK(pthread_create(
        &thread_handle_,
        &attr,
        [](void* arg) {
          Callable* callable = reinterpret_cast<Callable*>(arg);
          try {
            (*callable)();
          } catch (const std::exception& e) {
            const std::string& name = GetThreadName();
            LOG(WARNING) << "caught exception while running thread '" << name << "': " << e.what();
          } catch (...) {
            const std::string& name = GetThreadName();
            LOG(WARNING) << "caught unknown exception while running thread '" << name << "'";
          }
          delete callable; 
        }
        new Callable(std::forward<Callable>(callable)) == 0)
        << "Failed to create a new thread via 'pthread_create'";

    CHECK_EQ(0, pthread_attr_destroy(&attr));
  }
```

eventuals/os.h
```diff
@@ -60,6 +67,153 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+void* call_target(void* target) {
+  Lambda* lambda = reinterpret_cast<Lambda*>(target);
+  (*lambda)();
+  delete lambda;
+  return nullptr;
+}
+
+////////////////////////////////////////////////////////////////////////
+
+} // namespace details
```
benh, line 86:
```suggestion
```

eventuals/os.h
```diff
@@ -60,6 +67,153 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+ public:
+  Thread()
+    : thread_handle_{pthread_t{}},
+      joinable_(false) {}
+
+  ~Thread() noexcept(false) {
+    if (joinable_) {
+      throw std::logic_error{"A thread was left not joined/not detached"};
+    }
+  }
```
benh, line 98:
```suggestion
  ~Thread() noexcept() {
    CHECK(joinable_) << "A thread was left not joined/not detached";
  }
```

eventuals/os.h
```diff
@@ -60,6 +67,153 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+  Thread(Thread&& other) noexcept
+    : thread_handle_{std::exchange(other.thread_handle_, pthread_t{})},
+      joinable_{std::exchange(other.joinable_, false)} {}
+
+  Thread& operator=(const Thread& other) = delete;
+
+  Thread& operator=(Thread&& other) {
+    if (joinable_) {
+      throw std::logic_error{"Trying to drop not joined/not detached thread"};
+    }
```
benh, line 160:
```suggestion
    CHECK(!joinable_) << "thread not joined nor detached thread";
```

eventuals/os.h
```diff
@@ -60,6 +67,153 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+      throw std::runtime_error{
+          "Failed to create a new thread with "
+          "'pthread_create(), error code = '"
+          + std::to_string(creation_result)};
+    }
+  }
+
+  Thread(Thread&& other) noexcept
+    : thread_handle_{std::exchange(other.thread_handle_, pthread_t{})},
+      joinable_{std::exchange(other.joinable_, false)} {}
```
benh, line 153:
```suggestion
  Thread(Thread&& that) noexcept {
    *this = std::move(that);
  }
```

eventuals/os.h
```diff
@@ -60,6 +67,153 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+
+  Thread& operator=(const Thread& other) = delete;
+
+  Thread& operator=(Thread&& other) {
```
benh, line 157:
```suggestion
  Thread& operator=(Thread&& that) {
```

eventuals/os.h
```diff
@@ -60,6 +67,153 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+  bool is_joinable() const {
+    return joinable_;
+  }
+
+  void join() {
+    if (joinable_) {
+      int join_result = pthread_join(thread_handle_, nullptr);
+      if (join_result != 0) {
+        throw std::runtime_error{"Join error: " + std::to_string(join_result)};
+      }
```
benh, line 181: PCHECK

eventuals/os.h
```diff
@@ -60,6 +67,153 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+    }
+    joinable_ = false;
+  }
+
+  void detach() {
```
benh, line 186: You need to actually call `pthread_detach`!

eventuals/os.h
```diff
@@ -60,6 +67,153 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+    Callable_ callable_;
+
+    State_(Callable_&& callable)
+      : callable_(std::forward<Callable_>(callable)) {}
+
+    void operator()() {
+      std::invoke(callable_);
+    }
+  };
+
```
benh, line 205:
```suggestion
```

eventuals/os.h
```diff
@@ -60,6 +67,153 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+  (*lambda)();
+  delete lambda;
+  return nullptr;
+}
+
+////////////////////////////////////////////////////////////////////////
+
+} // namespace details
+
+class Thread {
```
benh, line 88: Let's put this in the `os` namespace!

test/os.cc
```diff
@@ -0,0 +1,44 @@
+#include "eventuals/os.h"
+
+#include <limits.h>
+
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+#if defined(__linux__) || defined(__MACH__)
+
+using eventuals::Thread;
```
benh, line 10:
```suggestion
namespace eventuals::test {
namespace {

#if defined(__linux__) || defined(__MACH__)
```
See other tests so we're consistent!

benh (changes requested):

eventuals/os.h
```diff
@@ -60,6 +67,171 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+      const size_t stack_size_in_bytes = 8'388'608)
+    : joinable_{true} {
+    CHECK_GE(stack_size_in_bytes, PTHREAD_STACK_MIN)
+        << "Stack size should not be less than"
+           " the system-defined minimum size";
+
+    pthread_attr_t attr;
+
+    PCHECK(pthread_attr_init(&attr) == 0)
+        << "Error in pthread_attr_init(...)";
```
benh, line 114:
```suggestion
        << "Failed to initialize thread attributes via 'pthread_attr_init(...)'";
```

eventuals/os.h
```diff
@@ -60,6 +67,171 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+                     == 0)
+                     << "Error in pthread_setname_np(...)";
+#endif
+
+#ifdef __MACH__
+                 PCHECK(
+                     pthread_setname_np(
+                         data->thread_name.c_str())
+                     == 0)
+                     << "Error in pthread_setname_np(...)";
```
benh, line 153:
```suggestion
                     << "Failed to set thread name via 'pthread_setname_np(...)'";
```

eventuals/os.h
```diff
@@ -60,6 +67,171 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+                 }
+                 delete data;
+                 return nullptr;
+               },
+               new Data(name, std::forward<Callable>(callable)))
+           == 0)
+        << "Failed to create a new thread via 'pthread_create'";
+
+    PCHECK(pthread_attr_destroy(&attr) == 0)
+        << "Error in pthread_attr_destroy(...)";
```
benh, line 175:
```suggestion
        << "Failed to destroy thread attributes via 'pthread_attr_destroy(...)'";
```

eventuals/os.h
```diff
@@ -60,6 +67,171 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+  }
+
+  bool is_joinable() const {
+    return joinable_;
+  }
+
+  void join() {
+    if (joinable_) {
+      PCHECK(pthread_join(thread_handle_, nullptr) == 0)
+          << "Join error";
```
benh, line 204:
```suggestion
          << "Failed to join thread via 'thread_join(...)'";
```

eventuals/os.h
```diff
@@ -60,6 +67,171 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+  }
+
+  void detach() {
+    CHECK(joinable_)
+        << "Trying to detach already joined/detached thread";
+
+    joinable_ = false;
+
+    PCHECK(pthread_detach(thread_handle_) == 0)
+        << "Error in pthread_detach(...)";
```
benh, line 216:
```suggestion
        << "Failed to detach thread via 'pthread_detach(...)'";
```

test/os.cc
```diff
@@ -0,0 +1,87 @@
...
+  MockFunction<void()> start;
+
+  EXPECT_CALL(start, Call())
+      .Times(1);
+
+  Thread t{[&]() {
+             start.Call();
+           },
+           "thread_name",
+           4'194'304};
```
benh, line 54:
```suggestion
           Megabytes(4)};
```

test/os.cc
```diff
@@ -0,0 +1,87 @@
...
+}
+
+TEST(Thread, Joinable) {
+  Thread t{&foo, "thread_name"};
+  EXPECT_TRUE(t.is_joinable());
+  t.join();
+  EXPECT_FALSE(t.is_joinable());
+}
+
+TEST(Thread, LambdaThatCapturesEverything) {
```
benh, line 44: This test is doing two things, it's showing you can capture things in the lamdba and it's also showing that you can set the stack size. Separate this into two independent threads. And in the test that is setting the stack size, call `GetStackSize()` in the body to assert that the stack size has been successfully set. If the PR with `GetStackSize()` is not completed then let's complete that first.

test/os.cc
```diff
@@ -0,0 +1,87 @@
...
+
+TEST(Thread, FunctionPointer) {
+  Thread t1{&foo, "thread_name1"};
+  Thread t2{foo, "thread_name2"};
+
+  t1.join();
+  t2.join();
+}
+
+TEST(Thread, Moveable) {
```
benh, line 72: Let's simplify this test by just using a lambda and capturing a `std::unique_ptr` which can only be moved. Then we don't need the global `functor_was_called` nor the `Moveable` class.

benh (changes requested):

eventuals/os.h
```diff
@@ -60,6 +67,171 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+           " (if you are on macOS - probably you are trying"
+           " to set the stack size which is not a multiple"
+           " of the system page size)";
+
+    struct Data {
+      std::string thread_name{};
+      Callable callable;
+      Data(const std::string& name, Callable&& f)
+        : thread_name{name},
+          callable{std::forward<Callable>(f)} {}
```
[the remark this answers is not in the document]
ArthurBandaryk: Yep! This is necessary, otherwise compiler will message us out that there is no matching constructor!

eventuals/os.h
```diff
@@ -60,6 +67,171 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+  // length of name is greater than 15 characters, the excess characters
+  // are ignored. If name is null, the thread name attribute is reset to
+  // the default.
+  // (check_line_length skip)
+  // https://www.ibm.com/docs/en/i/7.4?topic=ssw_ibm_i_74/apis/pthread_attr_setname_np.htm
+  template <typename Callable>
+  Thread(
+      Callable&& callable,
+      const std::string& name,
+      const size_t stack_size_in_bytes = 8'388'608)
```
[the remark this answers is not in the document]
ArthurBandaryk: Will keep `8'388'608` value because on macOS 8 Mb is incorrect stack size!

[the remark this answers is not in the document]
ArthurBandaryk: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/pthread_attr_setstacksize.3.html

test/os.cc
ArthurBandaryk: Will use Bytes(4'194'304) instead. Again, on macOS stack size should be a multiple of the system page size!

eventuals/os.h
```diff
@@ -60,6 +64,161 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+
+  // IMPORTANT NOTE: on macos stacksize should be a multiple of the system
+  // page size!!!
+  // (check_line_length skip)
+  // https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/pthread_attr_setstacksize.3.html
+  template <typename Callable>
+  Thread(
+      Callable&& callable,
+      const std::string& name,
+      const Bytes& stack_size = Bytes(8'388'608))
```
benh, line 95:
```suggestion
      const Bytes& stack_size = Megabytes(8))
```

eventuals/os.h
```diff
@@ -60,6 +64,161 @@ inline void SetAffinity(std::thread& thread, const size_t cpu) {
...
+    PCHECK(pthread_detach(thread_handle_) == 0)
+        << "Failed to detach thread via 'pthread_detach(...)'";
+  }
+
+ private:
+  // For Linux `pthread_t` is unsigned long, for
+  // macOS this is an `_opaque_pthread_t` struct.
+  // https://opensource.apple.com/source/xnu/xnu-517.7.7/bsd/sys/types.h
+  pthread_t thread_handle_;
+  bool joinable_;
```
benh, line 211: Let's set these to their defaults and make the empty constructor `= default;` instead!
```suggestion
  pthread_t thread_handle_ = {};
  bool joinable_ = false;
```

eventuals/os.h
ArthurBandaryk: Nope! On macOS it seems that 8Mb is illegal size!

So, I will keep `Bytes(8'388'608)`!

FolMing: ????? But isn't Bytes(8'388'608) the same thing as Megabytes(8) ?????
I don't follow you, @ArthurBandaryk

eventuals/os.h
```diff
@@ -192,6 +192,148 @@ inline void CheckSufficientStackSpace(const size_t size) {
...
+           "'pthread_attr_destroy(...)'";
+  }
+
+  Thread(Thread&& that) noexcept {
+    *this = std::move(that);
+  }
+
+  Thread& operator=(const Thread& other) = delete;
+
+  Thread& operator=(Thread&& that) {
```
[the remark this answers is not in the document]
FolMing: Also, mark it as noexcept

benh (approved):

PR #564, opened 2022-08-24
Fix heap-after-use bug in 'Concurrent'

benh:
ASAN found [this](https://github.com/3rdparty/eventuals/runs/7944786807?check_suite_focus=true).

### DIFF
--- a/eventuals/concurrent.h
+++ b/eventuals/concurrent.h
@@ -358,10 +358,26 @@ struct _Concurrent final {
                          || !fibers_done_;
                    };
                  }))
-          >> Then([callback = std::move(callback)]() mutable {
-               callback();
-             })
-          >> Terminal();
+          >> Terminal()
+                 .start([callback = std::move(callback)]() mutable {
+                   // NOTE: we need to move 'callback' on the stack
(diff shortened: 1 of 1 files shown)


FolMing:
This solution seems weird! Is there a possible "better" way to avoid it?

benh:
> This solution seems weird! Is there a possible "better" way to avoid it?

`Concurrent` needs to get reworked now that we have a lot more abstractions to simplify it and not have this weird "dangling" continuation that gets type-erased into a `Callback`. Until then, any time we're executing code that could possibly be destructed we *must* ensure that we have everything that we might ever access on the stack.

ArthurBandaryk (approved):

CodingCanuck:
> ASAN found [this](https://github.com/3rdparty/eventuals/runs/7944786807?check_suite_focus=true).

I saw what looks like a very similar error: https://github.com/3rdparty/eventuals/runs/8028230226?check_suite_focus=true

In a run on https://github.com/3rdparty/eventuals/pull/507 which ran *after* this PR.

Is this issue still outstanding?

PR #568, opened 2022-08-26
Rename header files

onelxj:
To be consistent with srcs as `.cc` let's make headers be `.hh` (it also makes some IDE read files as C++ by default)


ArthurBandaryk (approved):

benh:
Can we not do something like what they do [here](https://github.com/clangd/coc-clangd/issues/40)?

onelxj:
probably might also been resolved with adding 
```
"files.associations": {
        "*.h": "cpp"
 }
```

PR #569, opened 2022-08-26
Introduce `HeapSize()`

ArthurBandaryk:
This PR adds the possibility to determine for a chain of
eventuals the maximum amount of heap memory that needs to
get allocated.


eventuals/task.h
```diff
@@ -304,6 +311,7 @@ struct _TaskFromToWith final {
     // issues during destruction of 'k_' if it holds any references or
     // pointers to any (or within any) of the above members.
     K_ k_;
+    Bytes heap_size_ = 0;
```
benh, line 314: Please see the comment above, `k_` must be the _last_ member.

eventuals/task.h
```diff
@@ -383,6 +391,8 @@ struct _TaskFromToWith final {
               std::is_convertible<Value, To_>>,
           "eventual result type can not be converted into type of 'Task'");
 
+      heap_size_ = Bytes(sizeof(HeapTask<E, From_, To_>(f(args...))));
```
benh, line 394:
```suggestion
      heap_size_ = Bytes(sizeof(HeapTask<E, From_, To_>));
```
Here and anywhere else you need to do this.

eventuals/task.h
```diff
@@ -204,10 +205,12 @@ struct _TaskFromToWith final {
         std::variant<
             MonostateIfVoidOrReferenceWrapperOr<To_>,
             DispatchCallback<From_, To_, Args_...>>&&
-            value_or_dispatch)
+            value_or_dispatch,
+        Bytes&& heap_size)
```
benh, line 209: Please call this `static_heap_size` everywhere applicable.

test/range.cc
```diff
@@ -89,5 +89,18 @@ TEST(Range, SpecifiedStepNegative) {
   EXPECT_THAT(*s, ElementsAre(0, -2, -4, -6, -8));
 }
 
+TEST(Range, StaticHeapSize) {
+  auto s = []() { return Range(0, 5)
+                      >> Collect<std::vector>(); };
+
```
benh, line 95:
```suggestion
  auto s = []() { 
      return Range(0, 5)
            >> Collect<std::vector>();
  };

```

benh (changes requested):
Instead of adding checks for heap size in every test, let's create some new tests to show that types that don't have any heap size equal zero but those that do are a value greater than zero.

PR #585, opened 2023-04-10
Split streaming.cc into separate files

onelxj:
`streaming.cc` before split took more than 3 minutes to build (and sometimes it failed compilation with a system freeze on my local machine with clang14).
After the split it goes much faster (less than 15 sec) and I haven't detected any issues with the change


test/grpc/streaming/streaming.h
```diff
@@ -0,0 +1,118 @@
...
+#include "eventuals/loop.h"
+#include "eventuals/map.h"
+#include "eventuals/then.h"
+#include "examples/protos/keyvaluestore.grpc.pb.h"
+#include "gtest/gtest.h"
+#include "test/grpc/test.h"
+#include "test/promisify-for-test.h"
+
+namespace eventuals::grpc::test {
+namespace {
```
benh, line 18: Not in headers.
```suggestion
```

test/grpc/streaming/write_last_after_reply_two_requests.cc
```diff
@@ -0,0 +1,45 @@
+#include "streaming.h"
```
benh, line 1: Let's make this explicit like we do (or should be doing) in other files. Here and all the other `*.cc` files in this PR please!
```suggestion
#include "test/grpc/streaming/streaming.h"
```

test/grpc/BUILD.bazel
```diff
@@ -53,6 +53,36 @@ cc_binary(
...
+        "streaming/write_last_before_reply_one_request.cc",
+        "streaming/write_last_before_reply_two_requests.cc",
+        "streaming/writes_done_after_reply_one_request.cc",
+        "streaming/writes_done_after_reply_two_requests.cc",
+        "streaming/writes_done_before_reply_one_request.cc",
+        "streaming/writes_done_before_reply_two_requests.cc",
+        "test.h",
+    ],
+    copts = copts(),
+    target_compatible_with = select({
```
benh, line 70: Please copy any associated comments when you copy code so that we know why we're doing things in particular ways!
```suggestion
    # TODO(benh): resolve build issues on Windows and then remove
    # these 'target_compatible_with' constraints.
    target_compatible_with = select({
```

test/grpc/BUILD.bazel
```diff
@@ -53,6 +53,36 @@ cc_binary(
     }),
 )
 
+cc_library(
+    name = "grpc_streaming",
```
benh, line 57: Unless there is something that is keeping this from working it seems redundant and unnecessary to include the `grpc` prefix so let's drop.
```suggestion
    name = "streaming",
```

benh (approved):

PR #588, opened 2023-04-27
Remove a bunch of annoying clang-tidy checks

NikitaNikolaenko:
Removes the following very annoying checks:

1. modernize-use-trailing-return-type - this check constantly notifies about an option to use smth like this:
```
auto add_two_values(int a, int b) -> int
```
Useless and is not used in eventuals codebase.

2.  readability-function-cognitive-complexity - eventuals by their nature have extremely complex codebase, some of the functions have a cognitive complexity of 300+, this check acts more like a noise at this point.

3. readability-identifier-length - eventuals use one-letter-variables extensively across the codebase, no need for this check today.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -15,3 +15,6 @@ Checks: >
     modernize-*,
     performance-*,
     readability-*,
+    -modernize-use-trailing-return-type,
+    -readability-function-cognitive-complexity,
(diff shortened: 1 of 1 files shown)


NikitaNikolaenko:
Closing an old PR

PR #589, opened 2023-05-05
Make `Error` extend `std::exception`

onelxj:


eventuals/errors.h
```diff
@@ -12,11 +12,13 @@ namespace eventuals {
...
 // since 'std::make_exception_ptr' works with a copy argument
 // and there is no way to move error in it.
-class Error {
+class Error : public std::exception {
  public:
   virtual ~Error() = default;
 
-  virtual std::string what() const noexcept = 0;
+  const char* what() const noexcept override {
+    return "Eventuals inner 'Error'";
```
rjhuijsman, line 20: Perhaps you mean `internal` instead of `inner`?
Internal = it happened inside Eventuals
Inner = it is wrapped inside some outer thing (and I don't think there's an outer 'Error'?)

rjhuijsman (approved):
As discussed, I reviewed (only) the final commit.

eventuals/errors.h
benh, line 20: No need to override this! The subclasses will, and if they forget to you want that to be a compile time error!

benh (approved):

PR #597, opened 2023-06-23
Build fix

onelxj:
- Memory usage by Docker (via Dazel) was causing crashes on our limited-memory GitHub Actions runners.
- Dazel is no longer useful now that we mostly use Codespaces, so we can remove it.
- Fix asan tool.
- Moved to larger GitHub runners.
#599


onelxj:
Since we are using codespaces for develop, all docker (dazel) things can be removed, as well as run build using github runner config, but with config=asan off

.github/workflows/build_and_test.yml
```diff
@@ -108,7 +91,47 @@ jobs:
...
+        run: |
+          ${{ matrix.bazel-command }} \
+            ${{ matrix.output-user_root }} \
+            build \
+            ${{ matrix.bazel-config }} \
+            ${{ env.BAZEL_REMOTE_CACHE }} \
+            --verbose_failures \
+            -c dbg \
+            --strip="never" \
+            test/grpc/route_guide:route-guide-test
```
rjhuijsman, line 131: Did you say this splitting into multiple steps didn't have any effect on the memory usage? If so, can we revert this part of the change?

onelxj: I’m doing tests now, recently found some additional hazel flags that might help, will revert if it would be not necessary, thank you!

.github/workflows/build_and_test.yml
```diff
@@ -115,8 +100,6 @@ jobs:
             build \
             ${{ matrix.bazel-config }} \
             ${{ env.BAZEL_REMOTE_CACHE }} \
-            --verbose_failures \
-            -c dbg \
```
rjhuijsman, line 119: Why did we remove these, here and below?

(I'd normally like to see such changes explained in the PR description)

test/grpc/streaming/streaming.cc
```diff
@@ -66,12 +66,14 @@ void test_client_behavior(
...
-        >> std::move(handler);
+  auto call = [&]() -> Task::Of<::grpc::Status>::Raises<RuntimeError> {
+    return [&]() {
+      return client.Call<
+                 Stream<Request>,
+                 Stream<Response>>(
+                 "keyvaluestore.KeyValueStore.GetValues")
+          >> std::move(handler);
+    };
   };
```
rjhuijsman, line 77: Why is this part of this PR?

rjhuijsman (commented):
I see some more unexpected-to-me changes have popped up, so I'd like some clarification on those before I approve.

For posterity, it would be good to record the "why" of this PR in the PR description:
* Memory usage by Docker (via Dazel) was causing crashes on our limited-memory GitHub Actions runners.
* Dazel is no longer useful now that we mostly use Codespaces, so we can remove it.

test/grpc/streaming/streaming.cc
onelxj: added to the description, thank you!

benh:
@onelxj what's the status of this? Does this fix the build?

benh (approved):
Let's try adding this to `.bazelrc`:

```
# By default run both (target) compilation mode and host compilation
# mode in 'fastbuild' so that any "tools" that get build (which are
# built for the host) don't require building things a second time.
build --compilation_mode=fastbuild
build --host_compilation_mode=fastbuild
```

PR #604, opened 2023-10-18
Add arm64 MACH support for os.h

onelxj:
Backtrace is not currently supported for arm mac, but we use it only in cc_test, so we can build the core library.

### DIFF
--- a/eventuals/os.h
+++ b/eventuals/os.h
@@ -78,7 +78,7 @@ struct StackInfo {
   Bytes size = 0;
   inline Bytes StackAvailable() {
     [[maybe_unused]] char local_var{};
-#if defined(__x86_64__)
+#if defined(__x86_64__) || defined(__MACH__)
     return Bytes(
         (&local_var)
         - ((char*) end) - sizeof(local_var));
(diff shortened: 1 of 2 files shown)


eventuals/os.h
```diff
@@ -133,17 +125,19 @@ inline StackInfo GetStackInfo() {
       /* size = */ Bytes(size)};
 #endif // If Linux.
 
-#ifdef __MACH__
+#if defined(__MACH__) && (defined(__arm64__) || defined(__x86_64__))
   // For macOS `pthread_get_stackaddr_np` can return either the base
-  // or the end of the stack.
-  // Check 1274 line on the link below:
-  // https://android.googlesource.com/platform/art/+/master/runtime/thread.cc
+  // or the end of the stack. On Mac OS 10.7, it's the end.
```
benh, line 130: Please match the comments around you for consistency.
```suggestion
  // or the end of the stack. On macOS 10.7, it's the end.
```

test/os.cc
```diff
@@ -32,7 +32,13 @@ TEST(Thread, Joinable) {
 TEST(Thread, SetStackSize) {
   Thread t{
       []() {
+#if defined(__MACH__) && defined(__arm64__)
+        // Mac OS ARM sets the stack size a bit more than
```
benh, line 36:
```suggestion
        // macOS ARM sets the stack size a bit more than
```

benh (approved):

PR #621, opened 2024-05-08
Fix Concurrent deadlock

onelxj:
Fixes https://github.com/reboot-dev/respect/issues/2951

### DIFF
--- a/eventuals/concurrent.h
+++ b/eventuals/concurrent.h
@@ -307,7 +307,8 @@ struct _Concurrent final {
                 CHECK_EQ(
                     &fiber->context.value(),
                     Scheduler::Context::Get().get());
-                CHECK(fiber->context->in_use());
+                CHECK(fiber->context->in_use())
+                    << "Context: " << fiber->context->name();
 
                 fiber->done = true;
 
@@ -329,7 +330,8 @@ struct _Concurrent final {
                 CHECK_EQ(
                     &fiber->context.value(),
                     Scheduler::Context::Get().get());
-                CHECK(fiber->context->in_use());
+                CHECK(fiber->context->in_use())
+                    << "Context: " << fiber->context->name();
 
                 fiber->done = true;
 
@@ -502,6 +504,8 @@ struct _Concurrent final {
         interrupt.~Interrupt();
         new (&interrupt) class Interrupt();
 
+        // We should reset 'k' before 'context', because 'k' may contain
+        // a borrowed reference to 'context', which may lead to a deadlock.
         k.reset();
 
         context.reset();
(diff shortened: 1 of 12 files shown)


eventuals/concurrent.h
```diff
@@ -474,6 +483,32 @@ struct _Concurrent final {
...
+
+        done = false;
+
+        // Need to reinitialize the interrupt so that the
+        // previous eventual that registered with this
+        // interrupt won't get invoked as a handler!
+        interrupt.~Interrupt();
+        new (&interrupt) class Interrupt();
+
+        k.reset();
```
benh, line 505: Let's add a comment as to why we reset `k` before `context`, i.e., because `context` may be borrowed in `k`.

eventuals/concurrent.h
```diff
@@ -290,6 +283,12 @@ struct _Concurrent final {
           Eventual<void>()
               .context(std::move(stopped_or_error))
               .start([this, fiber](auto& /* stopped_or_error */, auto& k) {
+                CHECK_EQ(
+                    &fiber->context.value(),
+                    Scheduler::Context::Get().get());
+                CHECK(fiber->context->in_use())
+                    << "Context: " << fiber->context->name();
```
benh, line 290: Let's either put this everywhere or nowhere.

eventuals/event-loop.cc
```diff
@@ -335,6 +335,8 @@ void EventLoop::Check() {
 
       context->unblock();
 
+      context->in_use_.fetch_add(1);
```
benh, line 338: Let's make this:
```suggestion
      context->use();
```

eventuals/event-loop.cc
```diff
@@ -350,6 +352,8 @@ void EventLoop::Check() {
       ////////////////////////////////////////////////////
 
       CHECK_EQ(context, Context::Switch(std::move(previous)).get());
+      CHECK(context->in_use_.load() > 0);
+      context->in_use_.fetch_sub(1);
```
benh, line 356:
```suggestion
      context->unuse();
```
You can put the `CHECK` in the `unuse()` method.

eventuals/scheduler.h
```diff
@@ -138,6 +148,8 @@ class Scheduler {
     // Every context includes a waiter that can be used by schedulers.
     Waiter waiter;
 
+    std::atomic<int> in_use_ = 0;
```
benh, line 151: Move this `private` and then add `use()` and `unuse()` that we call everywhere.

test/concurrent/concurrent-with-timer.cc
```diff
@@ -0,0 +1,58 @@
...
+          typename decltype(e())::template ErrorsFrom<void, std::tuple<>>,
+          std::tuple<RuntimeError>>);
+
+  auto [future, k] = PromisifyForTest(e());
+
+  k.Start();
+
+  this->RunUntil(future);
+
+  auto r = future.get();
```
benh, line 52: What is `r`?

benh (approved):

